PR #74, opened 2019-08-13
Updated gitignore using gitignore.io

mgrinshpon:
I kept `private/` and `package-lock.json` but generated a gitignore from gitignore.io using the following gitignore options (since I use WebStorm):

- node
- komodo edit
- WebStorm+all
- VisualStudio
- VisualStudioCode

I can add any others that people want here.

This also seems like a good time to ask why package-lock.json is being added to gitignore. Isn't it desirable to have deterministic builds? I'm mostly asking out of naivete- I mostly work with Java.


zelitzsi:
@mgrinshpon I agree with you on `package-lock.json`. It helps ensure a consistent set of dependencies for builds. It works great with `npm ci` to install packages before deploying (or even testing locally).

dekkerglen:
@mgrinshpon , you can go ahead and remove the package-lock.json from the .gitignore then. It will still be a couple days until I have internet on my desktop machine, so I can't do any work until then.

PR #101, opened 2019-08-19
implement sample pack with seeding

### LINKED ISSUE #53: Sample Pack
As a user, I would like to be able to easily view a sample pack from my cube, that I can share with others.

This should tie in with the planned 'pick poll' feature, and also use a js library that concatenates images into a solid image, and use that image for the metadata on the page. That way the image is displayed in generated previews on social media

omnihound:
Fixes #53 
AFAIK the metatag spec should work, but I'm not sure how to test until it actually goes out in to the wild.  Happy to hard link the image somewhere on the sample pack page as well, just not sure where.

Enjoy!

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "dependencies": {
     "bcryptjs": "^2.4.3",
     "body-parser": "^1.19.0",
+    "canvas": "^2.6.0",
     "chart.js": "^2.8.0",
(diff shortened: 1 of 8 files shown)


omnihound:
Also includes the navbar cause copying and pasting it all the time felt bad.

Issue #84, opened 2019-08-16
Better Default Sort Orders


dekkerglen:
As a user, I would like the default sort order for types make more sense. 

Rearrange the lists of labels from getLabels() to be something that makes a bit more sense.

jamesriley72281:
I do believe I have this one. I had noticed this when working with the app just now and went in an alphabetized them. I can create a PR.

PR #104, opened 2019-08-22
[Issue #84] Updated Sort Orders

jamesriley72281:
Updated sort orders to be alphabetical. This makes things much easier to navigate and ready. I did leave the color sorts in WUBURG. Everybody loves WUBURG.

Not sure why this is showing the removed space added space edits after the sort edits I made?

### DIFF
--- a/public/js/editcube.js
+++ b/public/js/editcube.js
@@ -605,7 +605,7 @@ function GetColorIdentity(colors) {
 }
 
 function getSorts() {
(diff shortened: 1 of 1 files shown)


jamesriley72281:
> I don't think alphabetic is the best strategy, each order should be chosen more consciously. When I choose my sort, there are a few common ones that should be at the top, but artist will be very rarely used, so I don't like it as the first field.
> For the types, think about the natural order you like to go through types, maybe Creature, Planeswalker, Instant, Sorcery, Artifact
> Viewing artifacts first is a bit weird

I have to disagree. I feel any order other than alphabetic is subjective and would vary by each user what they would want to see. To me having the "set new default" feature means this list should be quick and easy to parse to then use that feature. I also did not arbitrarily go with alphabetical. I looked into how The Gatherer page lists their sort order to see if there is indeed a default like WUBURG for color and they are doing alphabetical as well.  https://gatherer.wizards.com/Pages/Default.aspx

PR #141, opened 2019-08-27
make samplepack work again

omnihound:


omnihound:
this is in master already by mistake

Issue #156, opened 2019-08-28
add Name as a sort/filter field


emmett9001:
In the Sort option on the List view, Name should be an option in the Primary Sort and Secondary Sort dropdowns. This would allow alphabetical sorting of the cube list by card name. A similar option should be added to the Filter Type dropdown in the Filter option. This would allow exact matches and regex matches on card names.

emmett9001:
It sounds like the current coupling between filter and sort, under which both are based on the same categorizations of the cube, makes alphanumeric sorting somewhat difficult. I get the impression that whether or not we're using Scryfall syntax, the problem of needing large categories to support sorting will persist. @dekkerglen Am I missing something about how Scryfall syntax fixes the issue?

dekkerglen:
Scryfall syntax for filtering would allow us to make a filter like "name:the" which would match true for any card with a name that contains 'the'. Check out the details here: https://scryfall.com/docs/syntax

emmett9001:
whoops, this is a duplicate of https://github.com/dekkerglen/CubeCobra/issues/94

PR #157, opened 2019-08-28
add card.details.name as a sort field

emmett9001:
This pull request fixes https://github.com/dekkerglen/CubeCobra/issues/156 by adding `Name` as a sort field. This allows alphanumeric sorting of the cube list.


emmett9001:
Closing because it seems like the current category-based filter/sort logic is not a great fit for alphanumeric sort. See https://github.com/dekkerglen/CubeCobra/issues/156#issuecomment-525993447

PR #158, opened 2019-08-28
Global js beautify

joshfinnie:
This PR implements a standard way to run the preferred js-beautify settings for this project.

Instead of using the website https://beautifier.io/, you can run a simple npm script to format your javascript code.

This PR includes all the changes that running `npm run-script beautify` would implement right out of the gate.


mawills:
Nitpick: Could you squash some of these commits? I think this could all be a single commit "configure and run global-js-beautify"

joshfinnie:
@mawills  I certainly can do that, but for the longest time I have been allowing Github to do it for me with a lot of success. Maybe that is something you'd be interested in adopting here. I believe it's a settings toggle, but once turned on the PR merge button changes to `Squash and merge`. Below is an example from one of my open source projects where I have this turned on.

image

I also understand if this is something that should be placed on the responsibility of the contributor as well.

dekkerglen:
Oh I need to enable this squash thing very badly

Issue #165, opened 2019-08-29
deduplicate new card addition


emmett9001:
In investigating possible implementations of https://github.com/dekkerglen/CubeCobra/issues/18, I noticed six copies of the same logic in the addition of new cards to a cube:
```javascript
cube.cards.push(
  {
    tags:['New'],
    status:"Not Owned",
    colors:details.color_identity,
    cmc:details.cmc,
    cardID:currentId[0],
    type:carddb.carddict[currentId[0]].type
  }
);
```
Here are [a](https://github.com/dekkerglen/CubeCobra/blob/497d6709eaf5dd0b32707554c4ccce46681f4d7a/routes/cube_routes.js#L1312-L1313) [few](https://github.com/dekkerglen/CubeCobra/blob/497d6709eaf5dd0b32707554c4ccce46681f4d7a/routes/cube_routes.js#L1007-L1008) [examples](https://github.com/dekkerglen/CubeCobra/blob/497d6709eaf5dd0b32707554c4ccce46681f4d7a/routes/cube_routes.js#L1345-L1346).

This duplication makes changing the default tags or status of newly added cards unnecessarily toilsome. This code should be refactored into a single reusable function, probably in [`util.js`](https://github.com/dekkerglen/CubeCobra/blob/497d6709eaf5dd0b32707554c4ccce46681f4d7a/serverjs/util.js)

emmett9001:
Looking into this refactoring, five of the calls look quite similar, while [this one](https://github.com/dekkerglen/CubeCobra/blob/497d6709eaf5dd0b32707554c4ccce46681f4d7a/routes/cube_routes.js#L1316) looks different. Instead of passing the `_id` attribute of a card details object as the `cardID` parameter, it passes an entire card details object. This looks like a bug, but I'm probably missing something about the expected format of `cube.cards.push`.

PR #166, opened 2019-08-29
deduplicate card addition

emmett9001:
This pull request deduplicates the addition of new cards to a cube by implementing `util.addCardToCube`. This function simply pushes a card object onto a `cube`'s `cards` array. This pull request also adjusts duplicated instances of this logic to use the new function.

Note that [this instance](https://github.com/dekkerglen/CubeCobra/blob/497d6709eaf5dd0b32707554c4ccce46681f4d7a/routes/cube_routes.js#L1316) of the duplicated logic is subtly different than the others in the manner described in https://github.com/dekkerglen/CubeCobra/issues/165#issuecomment-525996895. This pull request maintains this difference through the use of an extra argument to `util.addCardToCube`. If it's later discovered that this difference is a bug, the third argument to `util.addCardToCube` could be removed.

Fixes https://github.com/dekkerglen/CubeCobra/issues/165

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -977,16 +977,7 @@ router.post('/importcubetutor/:id',ensureAuth, function(req,res) {
                     found = true;
                     added.push(carddb.carddict[possible]);
                     var details = carddb.carddict[possible];
-                    cube.cards.push(
-                      {
-                        tags:['New'],
-                        status:"Not Owned",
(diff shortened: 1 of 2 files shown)


emmett9001:
The various workflows this pull request touches still need to be tested for correctness.

emmett9001:
Ready for review

PR #170, opened 2019-08-29
beautify public/js automatically

emmett9001:
This pull request adds `public/js` to the list of beautified directories.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,7 @@ Cube Cobra currently uses 4 languages: Javascript, Jade/PUG, CSS, and HTML. Plea
 
 #### Javascript
 
-For Javascript please make sure your code is formatted using the JS Beautify standard. This can be done by running `npm run-script beautify public/js/*.js`.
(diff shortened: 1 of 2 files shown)


emmett9001:
@joshfinnie mind taking a look at this? I'm not certain whether you used the `beautify` or `js-beautify` npm module in https://github.com/dekkerglen/CubeCobra/pull/158.

joshfinnie:
@emmett9001 I am installing `js-beautify` in the dev dependencies of the package.json here: https://github.com/dekkerglen/CubeCobra/blob/master/package.json#L45

I also purposefully excluded the `public` folder but on second look `public/js/**/.js` should be added to the script. I would modify the npm script (line 7 on package.json) to read `js-beautify --config .js-beautify.json -r *.js {config,models,routes,serverjs,views,public/js}/**/*.js` to incorporate that folder.

emmett9001:
Fixed! Thanks for the pointer @joshfinnie

joshfinnie:
np, nice find @emmett9001 !

PR #183, opened 2019-08-29
Deckbuilderfix

jamesriley72281:
Syntax causing an issue with deck building after draft


jamesriley72281:
all these commits.. ugh

Issue #160, opened 2019-08-28
"edit selected" is confusing


emmett9001:
The "Edit Selected" button is visible when looking at the cube in views other than List View. This is confusing because it is not useful in these other views. On these views, consider changing "Edit Selected" to something like "Mass Edit", which would be a link to the List View.

PR #192, opened 2019-08-30
changed wording of edit selected

purxiz:
This is an attempted resolution to #160 . This update changes the "Edit Selected" button to say "Mass Edit" while the user is not in list view, and changes it to say "Edit Selected" while they are in list view. In addition, pressing the button while not in list view will switch the user to list view instead of showing a dialog.

I am not sure about my implementation for changing the text of the button. An alternative might be to put some javascript into the pug file, to be evaluated on the page as the user switches views. This might
be preferable as it keeps all display related code in the pug file (cube_list.pug) rather than in the javascript.

I have also changed the name for the element "editSelected" to "massEdit" to reflect it's new default value.

Sorry the changes to editcube.js are a little hard to read, because github thinks I edited the contents of the else statement, rather than adding a wrapping if/else. 

I also think that the snippet of code in `UpdateCubeList()` that handles changing the button text could be moved to another function for code clarity if necessary, especially if any future similar dynamic element changes will be added.

### DIFF
--- a/public/js/editcube.js
+++ b/public/js/editcube.js
@@ -336,21 +336,25 @@ if (canEdit) {
       $('#cubeSaveModal').modal('show');
     });
   });
-  $('#editSelected').click(function(e) {
+  $('#massEdit').click(function(e) {
     e.preventDefault();
-    groupSelect = [];
-    cube.forEach(function(card, index) {
-      if (card.checked) {
-        groupSelect.push(card);
+    if (view == 'list') {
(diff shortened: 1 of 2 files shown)


emmett9001:
Related to https://github.com/dekkerglen/CubeCobra/issues/160

dekkerglen:
Please request merge into development, not master

purxiz:
Oops, sorry, on it

PR #197, opened 2019-08-30
Fix spelling of cookie and lengthen expiration.

phulin:
app.js misspelled cookie, and so the secure attribute was not actually being set. Additionally, the maximum age was not being set, and so logouts occur frequently. On the current site, this often leads to putting in a bunch of changes and then having them disappear because the "save" button just leads to a re-login. maxAge is now set to 1 year.

### DIFF
--- a/app.js
+++ b/app.js
@@ -59,8 +59,9 @@ app.use(session({
   secret: 'vertical donkey gatorade helicopter',
   resave: false,
   saveUninitialized: true,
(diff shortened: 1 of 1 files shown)


dekkerglen:
The maxage issue makes the session never expire, what you mentioned occurs when the server reboots due to a crash, which occurred frequently due to the new update breaking something.

PR #211, opened 2019-08-30
secrets example

emmett9001:
This pull request improves developer onboarding by providing an example of the required secrets directory in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,37 +22,15 @@ Atom (optional): https://atom.io/
 
 ### Clone Project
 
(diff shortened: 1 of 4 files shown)


dekkerglen:
I don't understand changes to readme, shouldn't we still keep the other files so the user understands all are required?

emmett9001:
I've removed the listing of individual files in the secrets directory to avoid the need to update the readme every time a file is added or changed in that directory. The instructions this change puts in the readme should be enough to get folks up and running without needing to know the details. If it's important to have specifics of the individual files listed, I can add those somewhere a little closer to the files themselves.

PR #222, opened 2019-09-01
Fix spelling of "cookie."

phulin:
Took out the maxage change.


purxiz:
Unfortunately changing the spelling of cookie without additional checking for production environment will break everyone's local environment, unless they somehow have https. I have already addressed this in PR #221 . I would recommend you check there.

phulin:
OK, closing for now.

Issue #131, opened 2019-08-26
Date added field


dekkerglen:
As a user, I would like to use the date added to Cube as a sortable card field.

When a card is added to a cube, set the date added field, and create a sort for that field as well.

emmett9001:
I'm working on this

emmett9001:
Does sorting by a date field have the same problem as in  https://github.com/dekkerglen/CubeCobra/pull/157#issuecomment-525698699 in which every card is in its own category?

dekkerglen:
No, you'll see groups of cards added in the same day/update

emmett9001:
This is done

dekkerglen:
There is a bug with this feature: If a card's property is changed (like version) the date added field is removed.

emmett9001:
Fixed in https://github.com/dekkerglen/CubeCobra/pull/223

PR #223, opened 2019-09-01
allow card update objects not to contain all card fields

emmett9001:
This pull request fixes https://github.com/dekkerglen/CubeCobra/issues/131 by setting missing fields on the card update object to their database values. This ensures that update objects that don't include all card fields don't cause those missing fields to be unset in the resulting database update. At the moment, `addedTmsp` is the only field to which this applies.


dekkerglen:
Can you rebase and fix the conflict?

emmett9001:
Conflict resolved

Issue #95, opened 2019-08-18
Default Card Version


dekkerglen:
As a user, when I add cards to my cube via quick add and bulk upload, I would like to be able to select a default version from at least:  ['Newest Printing','Original Printing', 'Cheapest Version','Most Expensive Version']

dekkerglen:
Also add checkbox for "no promo printing" (credit to DrRuler for suggestion)

PR #224, opened 2019-09-01
Use non-promotional printings by default.

phulin:
This update also cleans up a bunch of code in cube_routes. This is a partial solution for #95.


phulin:
Note this adds several fields to the card database (digital, promo, border_color).

dekkerglen:
These fields need to be set for pre-flipped cards as well.

PR #226, opened 2019-09-01
Refactor cube views to pull out common code.

phulin:
Includes unifying the tabs to use two spaces everywhere, but could remove that if it's too distracting.


phulin:
Will update to no longer conflict.

phulin:
Should be good to go now.

Issue #150, opened 2019-08-27
Pass cube name to page title


omnihound:
Make the page title customizable, such that a cube title or other custom text will appear in the browser's header.

PR #228, opened 2019-09-01
First crack at page titles.

phulin:
Currently, only works with the overview and list pages, but easy to expand if this is the right approach. Happy to make any changes if people want this done differently. Addresses #150.


dekkerglen:
Reverted: Please do not merge directly into master. Create a pull request into development only.

PR #233, opened 2019-09-01
secrets

joshfinnie:
This PR will update the way secrets are done for this application. It uses a library called [dotenv](https://www.npmjs.com/package/dotenv) which will read from a `.env` file and inject those variables into `process.env` for ease of use.

This would require a change on the server as the secrets would have to move from the current location to the `.env` file.

I do believe this is the best practice in today's node environments.


dekkerglen:
BTW I'm holding off on approving this b/c the current workflow keeps it very easy for me. one git pull and secrets are up to date. This is how we'll do it in the future but right now I'm going to keep it as is.

purxiz:
It will be the same with this system, just instead of having cubecobrasecrets you'll have a .env file in .gitignore

joshfinnie:
well, that rebase seems to have messed EVERYTHING up...

dekkerglen:
So, I the .env file will be in the root of the project, and we'll be able to paste the secrets in there?

purxiz:
Correct @dekkerglen

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "app.js",
   "scripts": {
     "beautify": "js-beautify --config .js-beautify.json -r *.js {config,models,routes,serverjs,views,public/js}/**/*.js",
-    "start": "node app",
+    "start": "nodemon",
```
crookedneighbor: This will fail to run the app in production because nodemon is a dev dependency, and most production deploys do not include the dev dependencies. (don't know exactly how this app is deployed, so maybe it does!)

IMO, you should leave this as is and create a new script for development:

```json
"dev": "nodemon"
```

crookedneighbor: Also, this seems to me like a separate PR from the `dotenv` change you're making.

joshfinnie: that's fair... I'll remove this change @crookedneighbor :-)

Issue #265, opened 2019-09-05
Table View Columns Sizing Incorrectly


jamesriley72281:
**Describe the bug**
User reported column sizing issue on Discord. You can see that the columns appear to be independently sizing dynamically based on card name lengths in columns. See link to cube and image below. 

*Link:* https://cubecobra.com/cube/list/5d707e4c2d52e15c253755fa 

**To Reproduce**
1. Go to link above (or any cube that doesn't push each column length to max size with card names)

**Expected behavior**
Table columns will all be sized the same

**Screenshots**
<img width="870" alt="asdfasdf" src="https://user-images.githubusercontent.com/22227186/64310832-6683b700-cf67-11e9-90d0-16e7f2432c46.PNG">

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome

PR #266, opened 2019-09-05
Further updates to mobile + other CSS.

phulin:


phulin:
Last commit should fix #265

PR #272, opened 2019-09-05
Begin transition to React

phulin:
This is a work-in-progress PR to test the beginnings of a transition to React for client-side layout, and eventually possibly all layout.

To test this, just run npm start after installing packages, and then go to localhost:8080 (NOT localhost:5000).


phulin:
With that last commit, I think this PR should be ready to merge. I welcome comments!

Issue #124, opened 2019-08-23
Custom Cards


lane-eden:
(Adding for a discoed member) As a user, I want to be able to add custom cards to my cube. Images would be ideal, but until that happens being able save custom metadata would suffice.
Add a way to add a custom card and set its name, a casting cost, a card type, rules text, and P/T.

Awkamess:
This is my number one requested feature. Currently using a disgusting spreadsheet to organize my cube. This is the only thing I ever wanted CT to have, I think CC can make it happen :)

pozz94:
Now that we have custom images a card named "Custom Card" with no cost, type or effect could be added to the card database as a temporary solution. It wouldn't be searchable by the new filter but I am sure it would suffice for now.

lane-eden:
I don't think it is a good idea to go with a halfway measure in this case. If people's image links break there will be no way to know what any of the cards in their cube do. If we allow setting rules text, etc. we could at least show the card's information in place of a missing image.

pozz94:
It wasn't meant as a real solution but as a temporary fix that could be implemented right now as we wait for a proper one.
Once we have a proper solution it would also be nice to be able to edit regular cards since some people use errata to tweak their cubes. I imagine that it would be very very similar in implementation

lane-eden:
Yes I know it wasn't meant as the long term solution. I am arguing against doing this temporary solution and instead waiting for the correct solution in this case. If we allow people to do this half baked version then when we do get to the full solution, we will have old users with "Custom Card" cards that aren't filled out, resulting in the issue I brought up above.

Zekromaster:
What if CubeCobra, for now, just supported adding cards from Planesculptors? They could be then autoconverted into the internal custom card format or whatever other option is adopted as the definitive one when it's implemented, with maybe a link somewhere for the original Planesculptors card.

They have a discord bot that fetches cards by set, set version and name, and it basically just does some transformation on the card name and calls their autocard api: https://github.com/mzabsky/webdrafter/blob/master/bot/index.js

It's not as polished as scryfall's API, but until something like Scryfall comes out for custom cards, that's the best option to avoid managing a database of custom cards and still support the feature.

dekkerglen:
I'm not interested in integrating with Planesculptors. The current scryfall integration paradigm involves importing _every_ card daily, which is not realistic for this change.

[...]

Issue #57, opened 2019-08-09
Custom Art Upload


mvhsbball22:
As a user, I'd like to be able to upload custom art for individual cards. My cube has significant amounts of alters and digital renders, and it would be nice to share that with people, especially those who might draft in person so they're familiar with the cards.

It probably makes most sense to allow users to upload images near the selection dialog for sets, then populate the drop down with a name the user can provide at time of upload.

dekkerglen:
Applying wontfix for now, because this will require significant upgrades to the server to make it possible.
Currently the hosting budget is too small to support uploading and hosting. We can come back to this issue later on when we have a larger budget, and we are able to accommodate this.

mvhsbball22:
This makes sense as a later thing.  

In the meantime as a stopgap, what about allowing a link to an external hosting service such as imgur?

dekkerglen:
It would still require a large update, but that's definitely doable.

emmett9001:
I'm sketching out a version of this that allows users to provide image urls, sidestepping the issue of image uploading and storage, in https://github.com/dekkerglen/CubeCobra/pull/243

emmett9001:
After chatting with @lane-eden and @purxiz, it sounds like the external URL hosting stopgap mentioned above is worth working on. I've noted some gotchas from our conversation at https://github.com/emmett9001/CubeCobra/pull/4

emmett9001:
@dekkerglen mind assigning me to this issue?

emmett9001:
Was this closed because it was decided not to implement image hosting?

PR #274, opened 2019-09-06
custom image urls

emmett9001:
This pull request addresses part of https://github.com/dekkerglen/CubeCobra/issues/57 by adding support for custom card image URLs. Every card may have an optional image URL associated with it. If configured for a card, this image will display everywhere the card would display within the app. URLs that either point to broken images or non-image resources are handled with a `background-image` CSS property of a Magic card back. Custom images can be set one card at a time using the modal editing interface. The custom image can be cleared for a card by clearing the Image URL input field in this modal and saving.

I've tested this change extensively on my local machine, but it's far-reaching enough that it's certainly possible I've missed some edge cases. Please point those out as you see them via comments, and I'll address them.

The approach of allowing external image URLs was chosen primarily because it allows custom image configuration, a widely-requested feature, to be released before the complex issue of image uploading and storage is resolved. Additionally, the majority of changes in this pull request apply to both the unhosted and the hosted implementation, which means that most of the effort put toward this change does not become "wasted" when image hosting is implemented.

This change was made with an eye toward compatibility with a future image hosting implementation, and has a clear migration path both for initial deployment and for updating to hosted images. When this change is first deployed, all cubes without custom images configured (all cubes) will be unaffected. Users can "opt in" to using custom images by configuring them on a per-card basis. When image hosting is deployed atop this change, configured external images will not need to be overwritten on deploy. The newly added `imgUrl` database field can be overwritten when an image is uploaded, and the behavior implemented in this pull request will cause those uploaded images to display properly.

Some miscellaneous notes on the implementation:
* Cards with URL that point to invalid images don't get autocard hover images. This indicates the broken state of the URL to the user while also sidestepping the annoying problem of making it work properly.
* Custom images are not included in blog posts.
* Flip cards show their normal, non-custom back side.
* Custom images are excluded from group edits. This is because the use case of assigning the same image to multiple cards is not something I chose to support in this pull request.
* There has been some debate in the Discord about whether to stretch or letterbox images that are not the same aspect ratio as Magic cards. This pull request chooses to stretch them.

Overview of changes
* Adds the `imgUrl` String field to the Cube database model
* Adds a default Magic card back image to `public/content`
* Adds some CSS rules implementing a class that defaults missing card images to the Magic card back using `background-image`
* Changes most uses of `card.details.image_normal` to `card.details.display_image`
* Adds logic to context modal submission causing `null` to be sent as the value of an update field when the image URL is unset
* Sets `card.details.display_image` as `imgUrl` if defined, `card.details.image_normal` otherwise
* Adds a Card Image URL field to the modal edit screen and populates it with `imgUrl` if specified
* Adds `util.getCardImageURL` as a convenience function implementing image precedence
* Adds a wrapping `<div>` around each card in the visual spoiler to allow placeholder images to align properly

Here is a screenshot of a visual spoiler using this functionality. Note that although the custom images configured in this screenshot depict unofficial Magic cards, this pull request does not address the issue of "custom cards", covered in https://github.com/dekkerglen/CubeCobra/issues/124. These images were chosen for the screenshot because they are clearly non-default images that are also the same shape as a Magic card.
Screenshot from 2019-09-05 17-02-35


omniczech:
Should allow users to disable custom images when viewing cubes.

lane-eden:
if the image url breaks can we just show the original card art? It would be lame to go into someone’s cube that they haven’t maintained and see a bunch of card backs

emmett9001:
An implementation of custom art toggle for cube viewers as @omniczech mentioned also solves the issue @lane-eden brought up. For a cube with unmaintained, broken images, the viewer can toggle custom images off to fix the viewing experience.

Defaulting broken images to the normal image has the issue of not providing clear feedback to the editing user, since a failed edit can appear to have had no effect.

I’m not sure whether allowing the toggle should block this pull request or be opened as a separate issue post-merge.

emmett9001:
Considering the above a bit more, I believe it's acceptable to merge this without support for the custom image toggle and to open a separate issue for that feature.

emmett9001:
Opened an issue to track the custom image toggle https://github.com/dekkerglen/CubeCobra/issues/279

PR #276, opened 2019-09-06
Adds ability for users to choose if price data is public

omniczech:


omniczech:
Gonna do a quick revision to make sure the price always shows for the owner

PR #280, opened 2019-09-06
Delete button starts disabled, only enabled when user confirms

rudyards:
Addressing issue #7


dekkerglen:
I reverted it, as this merge was going into master. Please create a new PR into development

Issue #92, opened 2019-08-17
MongoDB Presets


dekkerglen:
As a developer, I want to have some preconfigured data in my database after running the initial setup steps.

This can be accomplished by one of:
-  use mongodb's import/export feature, and have a pre-configured DB exported into a separate linked repo. 
- Or, we could include a script that populates some fixed values into the DB

swood456:
I'll work on this

crookedneighbor:
@swood456 how's the progress on this? Was just thinking this project could use a good populate db script when I found this issue.

swood456:
Wow I completely forgot about this. I had gotten [node-mongo-seeds](https://www.npmjs.com/package/node-mongo-seeds) set up and working but had a bit of issue pulling out the formatted data. I had made some progress and should be able to have it set up and working today

crookedneighbor:
The one suggestion I’d have about this is to use the mongoose models rather than importing data via the seeds module you posted. That will allow the seed script to remain up to date when new fields with defaults are added, or new validation rules are added. 

Node-mongo-seeds will work, but will require more work to keep it synced up with the mongoose models. 

Ultimately though, do whatever the maintainers of the project want to do. Just giving my 2 cents about it :)

swood456:
I see what you're saying, I got this set up with [mongoose-seed](https://www.npmjs.com/package/mongoose-seed) instead so it should have validation done via mongoose. PR: https://github.com/dekkerglen/CubeCobra/pull/309

crookedneighbor:
👏 👏 👏

PR #309, opened 2019-09-09
Feature #92 - Added seed data and script

swood456:
Running the script will wipe all databases except session_data and passwordresets and repopulate them with some sample data that I put together (a simple test user, two cubes, and a few drafts). Right now you can run the script just via `node seed.js`, I'm not sure if any info should be added to readme or if the script should be hooked into anything in that pipeline.


dekkerglen:
Can you revise this so it didn't wipe the db, it just adds items if they are not already present?

swood456:
Done, now the databases do not get wiped when the script is run. If you already have seeded, some errors are thrown because of attempts to duplicate data, but nothing gets changed. Would you like me to add an optional argument to wipe the db when the script is run?

PR #331, opened 2019-09-11
Filter Search Functionality and Correctness Update

purxiz:
This PR accomplishes many things:
1) Move all token parsing logic into token parsing, before the parser
   would pass raw user strings to the filter logic, and it would fail
there if it was going to fail. Now all arguments are checked and put
into a form the filter can use without further processing.
2) Argument checking, all arguments for fields with only specifically
   formatted arguments allowed are now checked at the time of filter
generation. An invalid filter string error is supplied to the user if
their parsed filter fails any of these checks.
3) Implemented support for shard/wedge/guild names in color and color
   identity searches
4) fixed color searches to use the card's actual color, and
   color_identity to use either the color_identity, or the cube owner's
overridden color identity if available.
5) Handles the word 'and'. See the commit for a full description, but in short, it just ignores it, and that's okay, since and is the default, and OR has to be specified by 'OR'. I recommend we don't reference the word AND on our syntax page at all, but it will still work for people who are familiar with the scryfall syntax.

Barring anybody finding issues, I think this PR is ready for production with 2 addendums:
1) Regex attacks are possible with the current implementation, so a timeout function should be started when generateFilters is called, and the operation should be aborted if it goes over a certain number of seconds. I'm not sure how important that is specifically, since this is all happening on the client machine, and I'm not familiar enough with Regex to know what (if any) strings would trigger an issue. So far I've jammed a bunch of special regex characters into the search bar, and not run into any issues.
2) The UI needs to be implemented for advanced search and the syntax page, but @dekkerglen has said he would do that!

### DIFF
--- a/public/js/editcube.js
+++ b/public/js/editcube.js
@@ -52,7 +52,6 @@ $('#compareButton').click(function(e) {
 
 $('#filterButton').click(function(e) {
   var filterText = $('#filterInput').val();
-  console.log(filterText);
   updateFilters(filterText);
 });
 
@@ -1638,12 +1637,7 @@ function updateFilters(filterText) {
 
   if (filterText) {
     new_filters = [];
-    if (generateFilters(filterText.toLowerCase(), new_filters)) {
-      filters = new_filters;
-      updateCubeList();
-    } else {
-      //TODO: couldn't parse that query, display error?
-    }
+    generateFilters(filterText.toLowerCase(), new_filters)
   } else {
     document.getElementById('filterarea').innerHTML = '<p><em>No active filters.</em></p>';
   }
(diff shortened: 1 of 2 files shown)


public/js/editcube.js
```diff
@@ -1799,42 +1849,128 @@ function generateFilters(filterText) {
...
     }
   } else {
-    return false;
+    document.getElementById('filterarea').innerHTML = '<p><em>Invalid Filter String.</em></p>';
   }
-  //return result;
 }
 
 const verifyTokens = (tokens) => {
+  console.log(tokens);
```
lane-eden: Extra console.log

purxiz: Removed in commit 3.

PR #334, opened 2019-09-12
Pull playtest page into React world.

phulin:
This PR also fixes a variety of minor layout issues with the page; it should display cleanly on a wide variety of screen sizes now.

image

image

image

image


phulin:
Noting just for posterity that the modal is still in jqueryland

PR #336, opened 2019-09-12
Save "show tag colors" state

lane-eden:


models/user.js
```diff
@@ -26,6 +26,10 @@ const UserSchema = mongoose.Schema({
     type: String,
     required: false
   },
+  hide_tag_colors: {
```
purxiz: Just rename this show tag colors so it's consistent with the front end: "show tag colors." Having one be hide and one be show is unnecessary, unless I'm missing something

lane-eden: It is this way so that we don't have to update the database with a script. Using hide_tag_colors for the db key means we can default it to false and if it is undefined it will still count as false.

purxiz: This isn't a breaking change, so current users won't be able to see tag colors until they check a check box. Personally I think that's preferable, but I'm sure we can just let a third party tie-break

lane-eden: I think the argument was that people will complain that colors aren't showing up, not realizing you need to turn it on. Whereas if they turn them off they will know that they have to turn them back on to see them.

PR #338, opened 2019-09-12
Fix password reset validation 

lane-eden:


purxiz:
What was the issue here?

lane-eden:
Passwords no longer have these requirements but these checks were accidentally left in in the password reset endpoints

Issue #362, opened 2019-09-13
Implement Tags and Status filtering


purxiz:
I would like to be able to filter my cube by tags or status of cards using the new filter syntax:
all you need to do it go to `categoryMap` in `editcube.js` and add `['tag', 'tag']`
then go to `filterApply` in `sortfilter.js`, look at name for an example, and implement tag similarly
it would probably just be `res = card.tags.contains(filter.arg)`

PR #363, opened 2019-09-13
add tag, status as filtering fields

emmett9001:
This pull request addresses part of https://github.com/dekkerglen/CubeCobra/issues/362 by adding `tag` as a filtering field. This fixes a regression in which this functionality was broken by https://github.com/dekkerglen/CubeCobra/commit/f04e6ca3fb67d59274a93221b402379d8f369ff0.

### DIFF
--- a/public/js/sortfilter.js
+++ b/public/js/sortfilter.js
@@ -84,6 +84,9 @@ function filterApply(card, filter) {
   if (filter.category == 'name') {
     res = card.details.name_lower.indexOf(filter.arg) > -1;
   }
+  if (filter.category == 'status') {
+    res = card.status.toLowerCase() === filter.arg;
+  }
   if (filter.category == 'tag') {
     var lowerTags = [];
     card.tags.forEach(function(tag) {
--- a/public/js/sortfilter.js
+++ b/public/js/sortfilter.js
@@ -88,11 +88,9 @@ function filterApply(card, filter) {
     res = card.status.toLowerCase() === filter.arg;
   }
   if (filter.category == 'tag') {
-    var lowerTags = [];
-    card.tags.forEach(function(tag) {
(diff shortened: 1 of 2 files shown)


emmett9001:
This could use a review from @purxiz

public/js/sortfilter.js
```diff
@@ -84,6 +84,16 @@ function filterApply(card, filter) {
   if (filter.category == 'name') {
     res = card.details.name_lower.indexOf(filter.arg) > -1;
   }
+  if (filter.category == 'status') {
+    res = card.status.toLowerCase() === filter.arg;
+  }
+  if (filter.category == 'tag') {
+    var lowerTags = [];
```
purxiz: Rather than instantiate a new variable, you could just use:

If you want to match partial tags, i.e. `tok` matches `token`
```javascript
card.tags.some(element => {
    element.toLowerCase().contains(filter.arg);
});
```
If you want to match only complete tags:
```javascript
card.tags.some(element => {
    element.toLowerCase() === filter.arg;
});```

File: public/js/sortfilter.js [hunk 2]
```diff
@@ -84,6 +84,16 @@ function filterApply(card, filter) {
...
+    card.tags.forEach(function(tag) {
+      lowerTags.push(tag.toLowerCase());
+    });
+    res = lowerTags.indexOf(filter.arg) > -1;
```
purxiz: Rather than instantiate a new variable, you could just use:

If you want to match partial tags, i.e. `tok` matches `token`
```javascript
res = card.tags.some(element => {
    element.toLowerCase().contains(filter.arg);
});
```
If you want to match only complete tags:
```javascript
res = card.tags.some(element => {
    element.toLowerCase() === filter.arg;
});```

dekkerglen:
Duplicate PR

PR #364, opened 2019-09-13
Feature 353

dekkerglen:


app.js
```diff
@@ -416,6 +422,6 @@ schedule.scheduleJob('0 0 * * *', function(){
 */
 
 // Start server
-http.createServer(app).listen(5000, 'localhost', function() {
+http.createServer(app).listen(4999, 'localhost', function() {
```
phulin: This was mistakenly committed I think
```suggestion
http.createServer(app).listen(5000, 'localhost', function() {
```

public/js/sortfilter.js
```diff
@@ -207,6 +209,79 @@ function filterApply(card, filter) {
       }
     }
   }
+  if(filter.category == 'tag') {
+    card.tags.forEach(function(val, index){
```
purxiz: I'd rather see this as
```javascript
res = card.tags.some(element => {
    element.toLowerCase() == filter.arg;
});```

File: public/js/editcube.js [hunk 3]
```diff
@@ -1658,7 +1775,14 @@ let categoryMap = new Map([
...
   ['toughness', 'toughness'],
-  ['name', 'name']
+  ['name', 'name'],
+  ['tag', 'tag'],
+  ['price','price'],
+  ['pricefoil','pricefoil'],
+  ['p','price'],
+  ['pf','pricefoil'],
+  ['status','status'],
+  ['s','status']
```
purxiz: I'd remove this line, so we can reserve `s:` for `set:`

public/js/sortfilter.js
```diff
@@ -207,6 +209,79 @@ function filterApply(card, filter) {
...
+          break;
+      }
+    }
+  }
+  if(filter.category == 'pricefoil')
+  {    
+    var price = null;
+    if (card.details.price_foil) {
+      price = card.details.price_foil;
+    }
```
phulin: Could be written more succinctly as `price = card.details.price_foil || null`

app.js
crookedneighbor: This should probably be configurable by an ENV variable and default to a value.

```js
// put this at the top of the file
const PORT = process.env.PORT || 4999;

http.createServer(app).listen(PORT, 'localhost', function() {
  console.log(`server started on port ${PORT}...`);
```

views/cube/cube_list.pug
```diff
@@ -164,6 +165,162 @@ block cube_content
   .container#cubelistarea
   #react-root
 
+  // Filter Modal
```
phulin: It doesn't block this PR, but I would love if we could start trying to get new UI elements implemented in React instead of Pug. Especially for mostly-client-side features.

public/js/sortfilter.js
```diff
@@ -207,6 +209,79 @@ function filterApply(card, filter) {
...
+    var price = null;
+    if (card.details.price) {
+      price = card.details.price;
+    } else if (card.details.price_foil) {
+      price = card.details.price_foil;
+    }
+    if (price) {
+      switch(filter.operand)
+      {
+        case ':':
```
phulin: Not for this PR, but I think we could combine all this code into one switch statement instead of having all the copy-and-pasted code.

views/cube/cube_list.pug
dekkerglen: I haven't spent enough time with react to do this, but moving forward I will. I just wanted to get this done for the release tonight.

public/js/editcube.js
```diff
@@ -514,6 +506,131 @@ $('#showTagColorsCheckbox').change(function(e) {
...
+    str += ' pf'+$('#filterPriceFoilOp').val()+$('#filterPriceFoil').val();
+  }
+
+  //status
+  if($('#filterStatus').val().length > 0)
+  {
+    str += ' stat:"'+$('#filterStatus').val()+'"';
+  }
+
+  //loyalty
```
emmett9001: If these comments are meant to indicate TODO items, they would be better as one or more GitHub issues. Then, they could be removed from this code.

public/js/editcube.js
```diff
@@ -514,6 +506,131 @@ $('#showTagColorsCheckbox').change(function(e) {
...
+  //Type
+  if($('#filterType').val().length > 0)
+  {
+    str += ' t:'+$('#filterType').val();
+  }
+
+  //tags
+  if($('#filterTag').val().length > 0)
+  {
+    str += ' tag:"'+$('#filterTag').val()+'"';
```
emmett9001: The lack of spaces around `+` operators here indicate that this code has not had `beautify` run on it. Setting up a code editor hook that automatically runs this command on save is a good way to make it so you don't have to think about beautifying your code.

public/js/sortfilter.js
```diff
@@ -1,3 +1,5 @@
+
+
```
emmett9001: Unnecessary whitespace

views/info/filters.pug
```diff
@@ -0,0 +1,29 @@
...
+  br
+  dl.row
+    dt.col-3 General
+    dd.col-9
+      p You can combine any number of filters together, using AND, or OR.
+      p t:instant or t:sorcery //cards that are either instants or sorceries
+      p t:instant t:tribal //cards that are both instants and tribal
+      p You can also use quotes to ignore spaces.
+      p goblin blood //cards whose names contain both blood and goblin
+      p "goblin blood" //cards whose names contain goblin blood
```
emmett9001: Are these comments user-visible? If so, they could use some clarification.

dekkerglen: @omniczech Said he would fix this file after I pushed it

public/js/editcube.js
dekkerglen: More as comments to myself before I got onto my plane, I'll create an issue for these fields

PR #368, opened 2019-09-13
Addressing Issue343

Zenoguy3:
Added set_type to carddict and used it in new function notPromoOrDigitalOrTokenId to ensure default card is not token.


emmett9001:
Looks like this was based on `master` instead of `development`. The relevant commits should be cherrypicked to a branch based on `development` and a new pull request opened.

phulin:
Just as a heads-up, we've switched to basing on `master` instead of `development`. You'll want to rebase off the current master, only including your changes to `cube_routes.js`.

PR #379, opened 2019-09-14
curve view fixes

lane-eden:
this fixes a couple issues in curve view
- secondary sort is ignored and uses Creature/Non-Creature instead
- selecting headers in curve view now opens the modal with that group selected like it does in table view
- hide secondary sort selector in curve view


lane-eden:
this fixes a couple issues in curve view
- secondary sort is ignored and uses Creature/Non-Creature instead
- selecting headers in curve view now opens the modal with that group selected like it does in table view

PR #399, opened 2019-09-15
Fix crash when loading a draft that has picks in it

lane-eden:


purxiz:
Please reference an issue in the description of this PR, even if it's a bugfix

routes/cube_routes.js
```diff
@@ -1569,19 +1569,12 @@ router.get('/draft/:id', function(req, res) {
...
-        if (Array.isArray(card)) {
-          card.forEach(function(item, index2) {
-            if (item) {
-              console.log(item);
-              item.details = carddb.cardFromId(item).cardID;
-              item.details.display_image = util.getCardImageURL(item);
-            }
-          });
-        } else {
+      // TODO this only handles the user picks (item 0 of draft picks), so custom images won't work with bot picks.
```
purxiz: Create an issue for this broken functionality as well, please!

routes/cube_routes.js
```diff
@@ -1569,19 +1569,12 @@ router.get('/draft/:id', function(req, res) {
...
+      draft.picks[0].forEach(function(col, index) {
```
phulin: This is a nitpick, but I would love if we could move towards using for..of instead of forEach as it's cleaner

routes/cube_routes.js
lane-eden: Ok done

routes/cube_routes.js
lane-eden: ok sure - this wasn't fresh code i was just making what was already there work.

PR #412, opened 2019-09-17
move .js-beautify.json to .jsbeautifyrc

emmett9001:
This pull request moves js-beautify configuration from `.js-beautify.json` to `.jsbeautifyrc`. `.jsbeautifyrc` is also used by [beautify-lint](https://www.npmjs.com/package/beautify-lint), which this pull request uses to validate code style during Travis builds. This pull request also runs beautify on the repository to make the Travis build pass.

Fixes https://github.com/dekkerglen/CubeCobra/issues/417


emmett9001:
Updated, could use another review. Changes to app js files are beautify only and not substantive.

.js-beautify.json
```diff
@@ -1,19 +0,0 @@
...
-  "max_preserve_newlines": "5",
-  "preserve_newlines": true,
-  "keep_array_indentation": false,
-  "break_chained_methods": false,
-  "indent_scripts": "normal",
-  "brace_style": "collapse",
-  "space_before_conditional": true,
-  "unescape_strings": false,
-  "jslint_happy": false,
-  "end_with_newline": false,
```
phulin: Can we change this to true? I think this is what most editors want.

emmett9001: I deliberately made no changes to the existing code style definition, since style discussions can be a rabbit hole. I've opened https://github.com/dekkerglen/CubeCobra/issues/450 to track changes to this definition outside of this pull request, which simply moves the existing definition.

phulin: Fair enough

PR #416, opened 2019-09-17
Improve password reset errors.

phulin:
Both logging and user error messages were essentially nonexistent.

I haven't been able to test this locally, so we'll have to do it on the dev server.

### DIFF
--- a/routes/users_routes.js
+++ b/routes/users_routes.js
@@ -138,36 +138,37 @@ router.post('/lostpasswordreset', function(req, res) {
           if (!user) {
             req.flash('danger', 'No user with that email found! Are you sure you created an account?');
             res.render('user/passwordreset');
+            return;
           }
           if (req.body.password2 != req.body.password) {
             req.flash('danger', 'New passwords don\'t match');
             res.render('user/passwordreset');
-          } else {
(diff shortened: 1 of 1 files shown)


routes/users_routes.js
```diff
@@ -130,30 +130,43 @@ router.post('/lostpasswordreset', function(req, res) {
...
+          if (err) {
+            console.error('Password reset find user error:', err);
+            res.sendStatus(500);
+            return;
+          }
+          if (!user) {
+            req.flash('danger', 'No user with that email found! Are you sure you created an account?');
+            res.render('user/passwordreset');
+          }
+          if (req.body.password2 != req.body.password) {
```
dekkerglen: Because you are not using else, you can send multiple responses and crash the server in the case of multiple errors.

Remove 'return' and refactor this such that it deterministically sends exactly one response

phulin: Good catch, fixed. This function would turn into a giant tree without using return so I'd much prefer to leave that in.

dekkerglen: Be careful about using return with nested callbacks, it could just return the anonymous function, and still result in multiple responses. This looks fine to me though.

PR #424, opened 2019-09-17
Save position of cube list (List/Table/Curve View) in window hash.

### LINKED ISSUE #317: Put list view mode into the URL as a query param
As a user, I want to be able to go into Curve View or Visual Spoiler mode and refresh the page or share the link and still be in that mode. Similar to sort and filter, sort the list view mode in the URL for the page.

phulin:
Fixes #317, although not in the way suggested. Blocked on #391, #394, #408.

### DIFF
--- a/src/cube_list.js
+++ b/src/cube_list.js
@@ -14,9 +14,11 @@ class CubeList extends Component {
   constructor(props) {
     super(props);
 
+    const hash = window.location.hash.replace('#', '');
+    const options = ['table', 'curve', 'spoiler', 'list']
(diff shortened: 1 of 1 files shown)


purxiz:
any particular reason to use a hash rather than `?view=`? we already are planning to use URLparams for other features.

phulin:
Good point. I just did this because it was annoying me and the fix was extremely quick, but happy to change it. We might want to move all those query strings into the hash, as they don't actually need to go to the server.

PR #449, opened 2019-09-19
Validate colors on CSV upload.

### LINKED ISSUE #446: Context Modal Broken
On development, individual card context modal is broken, does not open, and hides the view.

On 2.1, the context modal doesn't work for lands specifically. Doesn't open modal.

phulin:
Fixes #446

### DIFF
--- a/one_shot_scripts/validate_colors.js
+++ b/one_shot_scripts/validate_colors.js
@@ -4,17 +4,18 @@ const config = require('../config/database');
 
 (async () => {
   mongoose.connect(config.database).then( async (db) => {
-    const cubes = await Cube.find();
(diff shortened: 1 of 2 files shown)


one_shot_scripts/validate_colors.js
```diff
@@ -0,0 +1,21 @@
+const mongoose = require('mongoose');
+const Cube = require('../models/cube');
+const config = require('../config/database');
+
```
dekkerglen: This script will OOM on live server. Can't load all cubes into memory at once.

phulin: Will fix

PR #459, opened 2019-09-20
Remove 'New' on adding cards

dekkerglen:
This updates remove the default adding of the 'New' tag.

Not 100% sure if this is how we want to do this, as we've received a lot of feedback complaining about the 'New' tag, but people who appreciate it probably haven't spoken up.

I personally think it's ok to remove, as you can still see what's new from the blog history.


purxiz:
I would rather just have an option to tag all cards that are on the add side on an add/remove operation.

pino-cc:
I think having a new tag and have it removed automatically when you add other cards or like stated above. When adding cards have the bulk enter with a tag of your choosing like "testing"

Issue #395, opened 2019-09-15
unit test coverage


emmett9001:
The project code should be factored into testable units. Each of these units should have at least one unit test associated with it. These tests should run automatically on each pull request when it's opened or has new commits added, and the results should be visible in GitHub.

This issue cannot be solved in a single pull request. It's meant to track the unit testing effort over the course of multiple pull requests and comments. Feel free to link to this issue when you make a change related to unit testing.

purxiz:
Can you either add a roadmap/progress report, or separate this into multiple issues, so we can assign people as appropriate?

emmett9001:
Splitting this into smaller issues is a great idea, but the reason I haven't already done that is that I'm not entirely sure how to go about it. What are some of the smaller issues you'd like to see related to this? Please open them if you'd like. I wanted to avoid making 20 issues that are all "Write unit tests for the draft code", "Write unit tests for the visualspoiler code", etc. The current untestable state of much of the code makes it hard to know along which lines the unit tests will be divided, since there's a refactoring process that has to happen almost concurrently with the test writing.

emmett9001:
My comment "Feel free to link to this issue when you make a change related to unit testing" is made in the hope that this ticket will become its own progress report over time.

emmett9001:
Pull requests now run the test suite automatically via Travis, and github flags test failures in the PR interface.

emmett9001:
`analytics.js` and the rest of the `cubefn.js` functions are the only pieces of `serverjs` that still don't have tests.

PR #461, opened 2019-09-20
cards.js tests and refactors

emmett9001:
This pull request refactors cards.js to make it testable and adds some test fixtures to support future unit tests. The fixtures are created by taking the first 100 entries in the `cards.json` file and using them to generate the other carddb files. The main refactors are code deduplications and the addition of promises to the file loading process. Promises support unit tests by providing a hook that the tests can use to know when the loading is complete.

File watcher registration is made optional so that tests can disable it. The open filehandles left by the watchers cause Jest to fail.

This pull request also fixes a few small bugs in the cards.js file that look to me like they were in previously unused codepaths.

Related to #395


emmett9001:
This is ready for review.

emmett9001:
Ready for review

PR #462, opened 2019-09-20
Feature 414

### LINKED ISSUE #414: Improve Sample Pack
There are three big improvements we can make to sample pack.

Currently, the random sample pack goes to route: https://cubecobra.com/cube/samplepack/:cubeid
we should instead make it go directly to the route under 'link pack' so users can immediately copy the url and link to that exact pack.

The metadata for the sample pack isn't working. We can use the metadata for cube overview as an example for how it should look.

Instead of showing 15 separate images, we should just display the image from the metadata link, so it's a single image users can copy/paste, no need for snip tools!

dekkerglen:
Fixes #414

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "selenium-webdriver": "^4.0.0-alpha.4",
     "serve-favicon": "^2.5.0",
     "shuffle-seed": "^1.1.6",
-    "website-scraper": "^4.0.1"
+    "website-scraper": "^4.0.1",
+    "zlib": "^1.0.5"
   },
   "devDependencies": {
     "@babel/core": "^7.5.5",
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -833,7 +833,7 @@ router.get('/analysis/:id', function(req, res) {
 });
 
 router.get('/samplepack/:id', function(req, res) {
-  res.redirect('/cube/samplepack/'+req.params.id+'/'+Date.now().toString());
+  res.redirect('/cube/samplepack/' + req.params.id + '/' + Date.now().toString());
 });
 
 router.get('/samplepack/:id/:seed', function(req, res) {
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -858,10 +858,10 @@ router.get('/samplepack/:id/:seed', function(req, res) {
           metadata: generateMeta(
             'Cube Cobra Sample Pack',
             `A sample pack from ${cube.name}`,
-            `https://cubecobra.com/cube/samplepackimage/${req.params.id}/${pack.seed}.png`,
-            `https://cubecobra.com/cube/samplepack/${req.params.id}/${pack.seed}`,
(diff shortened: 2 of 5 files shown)


routes/cube_routes.js
```diff
@@ -885,8 +857,8 @@ router.get('/samplepack/:id/:seed', function(req, res) {
           metadata: generateMeta(
             'Cube Cobra Sample Pack',
             `A sample pack from ${cube.name}`,
-            `https://cubecobra.com/cube/samplepackimage/${req.params.id}/${pack.seed}`,
-            `https://cubecobra.com/cube/samplepack/${req.params.id}/${pack.seed}`
+            `http://162.243.163.51/cube/samplepackimage/${req.params.id}/${pack.seed}.png`,
```
phulin: I assume you don't want this in production?

dekkerglen:
I'm not pushing this until I get it to work on dev server. Currently it's still having problems.

dekkerglen:
I believe it's an issue with the encoding we use to serve this image

dekkerglen:
Fixed

routes/cube_routes.js
dekkerglen: Yes, just doing some tests on dev server. PR is ready for review now

routes/cube_routes.js
```diff
@@ -915,9 +891,11 @@ router.get('/samplepackimage/:id/:seed', function(req, res) {
       }).then(function(image) {
         res.writeHead(200, {
           'Content-Type': 'image/png',
-          'Content-Length': image.length
+          'Content-Encoding': 'gzip'
+        });
+        zlib.gzip(Buffer.from(image.replace(/^data:image\/png;base64,/, ''), 'base64'), function(_, result) {
+          res.end(result);
```
phulin: Don't we want the webserver to be doing the compression? Also isn't it not going to accomplish much to compress an image that is already in a compressed format?

dekkerglen: What format is that? base64? Image previews require gzip or another specific compression method

dekkerglen: Refactored this, you're right that it's already compressed, as gzip. We just need to add the encoding into the header, and add og:width and og:height to make this work.

package.json
```diff
@@ -46,7 +46,8 @@
     "selenium-webdriver": "^4.0.0-alpha.4",
     "serve-favicon": "^2.5.0",
     "shuffle-seed": "^1.1.6",
-    "website-scraper": "^4.0.1"
+    "website-scraper": "^4.0.1",
+    "zlib": "^1.0.5"
```
phulin: Unnecessary now

routes/cube_routes.js
```diff
@@ -5,6 +5,7 @@ const fs = require('fs');
 const fetch = require('node-fetch');
 const rp = require('request-promise');
 const cheerio = require('cheerio');
+const zlib = require('zlib');
```
phulin: same

PR #467, opened 2019-09-21
updatecards.js tests and refactors

emmett9001:
This pull request refactors `serverjs/updatecards.js` for testability and adds unit tests for it.

Change overview
* Moves `fixtures` to toplevel to avoid `jest` attempting to run fixtures as tests
* Adds `updatecards.test.js` exercising the functionality in `updatecards.js`
* Adds `fixtures/cards_small.json` - a version of `private/cards.json` containing only 100 entries used for testing
* Adds `fixtures/examplecards.js` containing a few objects representing the output of `convertCard`
* Adds `rimraf` as a dependency used for clearing state between tests that write files to disk
* Makes internal data structures in `updatecards.js` public for visibility in tests
* Factors scryfall json download into its own function, enabling mocking in tests
* Deduplicates many redundant card name normalization calls
* Makes the `updateCardbase` source filepath configurable, allowing tests to use a fixture
* Abstracts internal data structure mutations in `updatecards.js` into `addCardToCatalog`, an independently testable function
* Abstracts file writing into a function returning a promise, allowing multiple file writes to be un-nested and the calling function to wait on one promise representing all file writes
* Factors duplicated code between `convertCard` and `convertExtraCard` into a single function
* Moves differing functionality between `convertCard` and `convertExtraCard` into attribute-specific `convertX` functions

The deduplication of `convertCard` and `convertExtraCard` is the most likely part of this pull request to contain subtle errors. The most helpful aspect of a review here is to double-check the format of the example objects in `updatecards.test.js` to ensure that they match the expected format.

Related to #395


emmett9001:
Ready for review

PR #474, opened 2019-09-22
Added counts to deckbuilder

Balkmeister:
I Added a count to the top of each column in the deck builder and a creature count total in the top row of the deck builder. These are things i was missing while using the service.

### DIFF
--- a/public/js/deckbuilder.js
+++ b/public/js/deckbuilder.js
@@ -175,45 +175,22 @@ function renderDraft() {
   var creatures = 0;
   //fill up deck
   deck.playerdeck.forEach(function(col, index) {
-    var colhtml = "";
-    var i = 0;
     col.forEach(function(card, index2) {
-      i= i+1;
       cards++;
       if (card.details.type.toLowerCase().includes('land')) {
         lands++;
       }
-      if (card.details.type.toLowerCase().includes('creature')){
+      if (card.details.type.toLowerCase().includes('creature')) {
         creatures++;
       }
-      if (card.details.card_flip) {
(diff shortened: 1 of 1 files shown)


public/js/deckbuilder.js
```diff
@@ -172,33 +172,47 @@ function renderDraft() {
 
   var lands = 0;
   var cards = 0;
+  var creatures = 0;
   //fill up deck
   deck.playerdeck.forEach(function(col, index) {
     var colhtml = "";
+    var i = 0;
```
phulin: You don't have to define a new variable here; you can just use `col.length` everywhere you are using `i`.

public/js/deckbuilder.js
```diff
@@ -172,33 +172,47 @@ function renderDraft() {
...
   deck.playersideboard.forEach(function(col, index) {
-    var colhtml = "";
+    var colhtml ='';
+    var i = 0;
     col.forEach(function(card, index2) {
+      i= i+1;
       if (card.details.card_flip) {
         colhtml += '<a style="z-index:' + index2 + '; position: relative; top:-' + 155 * (index2) + 'px;" class="autocard" card="' + card.details.display_image + '" card_flip="' + card.details.image_flip + '" href="#"><img class="sideboardcard" data-id="' + index2 + '" data-col="' + index + '" src="' + card.details.display_image + '" width="' + cardWidth + '" height="' + cardHeight + '"/></a>';
-      } else {
+      } else {
```
phulin: Make sure you run npm run-script beautify before submitting - that will eliminate whitespace errors like this one.

PR #481, opened 2019-09-22
[WIP] Beautify src directory.

phulin:
Something is still wrong with js-beautify, see e.g. PagedTable.js.


emmett9001:
I think js-beautify is attempting to beautify that file as javascript because of its extension. js-beautify allows one to specify html settings in addition to js in the `.jsbeautifyrc`, but I don't know if it can be made to handle the unusual react tag syntax in javascript files.

phulin:
Theoretically that’s what the e4x setting is for in jsbeautifyrc. It worked
on most of the files, just not that one.

phulin:
@emmett9001 Would you object to switching to a different beautify solution? I’ve used prettier in the past and it has good support for JSX.

emmett9001:
@phulin Two important considerations: a replacement should use the same code style as our current `.jsbeautifyrc`, and it should use a comparable configuration file so that contributors can easily set up their own tooling to use the same settings. If those two conditions can be satisfied, I think swapping is fine.

Issue #427, opened 2019-09-17
Put Prices outside of scroll area for context modal


dekkerglen:
As a user, I want to be able to see the price of selected cards at any point in the context modal. 

Move the price below the scrollable area, so I can always see the prices independent of the scroll position

omniczech:
This should be quick, I'm on it

PR #487, opened 2019-09-23
Adds Pricing to Group modals outside scroll area.

omniczech:
Fixes: #427

Returns pricing info to group modals.

### DIFF
--- a/src/components/GroupModal.js
+++ b/src/components/GroupModal.js
@@ -201,6 +201,11 @@ class GroupModal extends Component {
       return contextChildren;
     }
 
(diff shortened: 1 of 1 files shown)


src/components/GroupModal.js
```diff
@@ -201,6 +201,11 @@ class GroupModal extends Component {
       return contextChildren;
     }
 
+    const accumulator = (total, card) => {if(card.details.price){return total + card.details.price} return total}
```
phulin: This is kind of a nitpick but this would be cleaner as `total + (card.details.price || 0)`

omniczech: Fair point, made the change.

PR #497, opened 2019-09-24
Fixed #409

### LINKED ISSUE #409: Sort alphabetically at the last level.
All of the list views appear to sort by date added as a last-level sort. Instead, they should sort in alphabetical order.

dekkerglen:
This adds alphabetical to CMC groups, so the newest cards won't be at the bottom.
This also reworks list view a bit, it doesn't use CMC as a tertiary sort at all, but it does sort alphabetically after the 2nd level.
This also adds back in 'Unsorted' as a sort, this was removed at some point as part of a regression.
One thing you can do is set primary and secondary sort to 'Unsorted' and go to list view to see a list of your cube in alphabetical order:
image

### DIFF
--- a/src/components/AutocardListGroup.js
+++ b/src/components/AutocardListGroup.js
@@ -32,7 +32,7 @@ const AutocardListGroup = ({ cards, heading, primary, secondary, tertiary }) => 
                 groups[cmc].sort(function(a,b)
                 {
                   const textA = a.details.name.toUpperCase();
-                  var textB =  b.details.name.toUpperCase();
+                  const textB =  b.details.name.toUpperCase();
                   return (textA < textB) ? -1 : (textA > textB) ? 1 : 0;
                 }).map(card =>
                   (<AutocardListItem key={card.details.name} card={card} />)
--- a/src/components/AutocardListGroup.js
+++ b/src/components/AutocardListGroup.js
@@ -29,7 +29,12 @@ const AutocardListGroup = ({ cards, heading, primary, secondary, tertiary }) => 
           <Row key={cmc} noGutters className="cmc-group">
             <Col>
               {
-                groups[cmc].map(card =>
+                groups[cmc].sort(function(a,b)
+                {
+                  var textA = a.details.name.toUpperCase();
+                  var textB =  b.details.name.toUpperCase();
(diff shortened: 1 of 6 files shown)


src/components/AutocardListGroup.js
```diff
@@ -29,7 +29,12 @@ const AutocardListGroup = ({ cards, heading, primary, secondary, tertiary }) =>
           <Row key={cmc} noGutters className="cmc-group">
             <Col>
               {
-                groups[cmc].map(card =>
+                groups[cmc].sort(function(a,b)
+                {
+                  var textA = a.details.name.toUpperCase();
+                  var textB =  b.details.name.toUpperCase();
+                  return (textA < textB) ? -1 : (textA > textB) ? 1 : 0;
```
omnihound: This seems useful (and looks like you've used it again), maybe you would like to put this in a utility file to keep things DRY

I'm not a fan of the nested ternary personally, as they can get difficult to read.  I would split it out in to if statements personally but I'd respect if you wanted to keep it that way

phulin: I'm fine with leaving this here for now, but can you add a fixme to pull the sorting logic out into a separate file?

phulin: This is a nitpick so fine if you don't get a chance, but I've been trying to use arrow functions and `const` consistently in these files.

src/components/ListView.js
```diff
@@ -218,55 +215,58 @@ class ListViewRaw extends Component {
...
-                    deleteTag={this.deleteTag.bind(this, index)}
-                    reorderTag={this.reorderTag.bind(this, index)}
-                  />
-                </td>
-              </tr>
-            )
-          ))
+          groups[label1][label2].sort(function(a,b)
+          {
+            var textA = a.details.name.toUpperCase();
```
phulin:
```suggestion
            const textA = a.details.name.toUpperCase();
```

src/components/ListView.js
```diff
@@ -218,55 +215,58 @@ class ListViewRaw extends Component {
...
+            var textB =  b.details.name.toUpperCase();
```
phulin:
```suggestion
            const textB =  b.details.name.toUpperCase();
```

Issue #68, opened 2019-08-13
Card Middle Click


dekkerglen:
As a user, I would like to be able to middle click (or ctrl+click) a card on the list view to navigate towards a card-specific page. 

Use tcgplayer (w/ affiliate link) please. Later on, we can add a new route like '/card/:id' that will pull up a page for the card with information, a buy link, a list of cubes that run this card, etc.

When those list items are clicked, they have an event handler that is intercepts the navigation with e.preventDefault(), the default navigating to '#', we just need to change that URL to the card-specific page. This will only work when middle clicking (or ctrl+click) the item, it won't affect the modal popup.c

night-day:
Please assign this to me, thanks!

PR #539, opened 2019-09-29
This is for issue #68

night-day:
Created a new utiility js for creating the TCG affiliate link

Updated the CardModal to use new funciton

Added middle click check via OnAuxClick on the ListView and open TCG link in new window

### DIFF
--- a/src/components/AutocardListItem.js
+++ b/src/components/AutocardListItem.js
@@ -5,9 +5,12 @@ import { Button } from 'reactstrap';
 import CardModalContext from './CardModalContext';
 import DisplayContext from './DisplayContext';
 
+import Affiliate from '../util/Affiliate';
+
 const AutocardListItem = ({ card, noCardModal, children }) => {
   let { display_image, image_normal, image_flip, name } = card.details;
   let { tags } = card;
+  
   return (
     <DisplayContext.Consumer>
       {({ showCustomImages, showTagColors }) => {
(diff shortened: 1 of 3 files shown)


src/util/Affiliate.js
```diff
@@ -0,0 +1,15 @@
...
+    } else {
+      tcgplayerLink += `productcatalog/product/show?ProductName=${card.details.name}`;
+    }
+    tcgplayerLink += '&partner=CubeCobra&utm_campaign=affiliate&utm_medium=CubeCobra&utm_source=CubeCobra';
+  
+     return tcgplayerLink;
+}
+
+  
+export default { getTCGLink };
```
phulin: Can you change the indentation in this file to 2 spaces and remove the empty line at the start?

src/components/AutocardListItem.js
```diff
@@ -25,6 +28,7 @@ const AutocardListItem = ({ card, noCardModal, children }) => {
                 <a
                   href={noCardModal ? undefined : '#'}
                   className="d-block w-100"
+                  onAuxClick={noCardModal ? undefined : e => { e.preventDefault(); if (e.button == 1) { window.open(Affiliate.getTCGLink(card)); }}}
```
phulin: This handler is complicated enough that it should probably be broken out into a separate function.

src/util/Affiliate.js
```diff
@@ -0,0 +1,15 @@
+
+function getTCGLink(card) {
```
phulin: Might as well export this function as a named export, too.
```suggestion
export function getTCGLink(card) {
```

src/components/AutocardListItem.js
night-day: Thanks for the suggestion, where should I define the function?  in that same js?

night-day:
I updated and fixed based on phulin comments hoping it worked out

purxiz:
References #68

PR #542, opened 2019-09-29
Top Cards feature.

### LINKED ISSUE #43: Top Cards
As a user, I would like to be able to see what cards are picked the most in a cube and what cards a player most frequently picks.

### LINKED ISSUE #546: Status filter bad generation
The status filter is incorrectly case sensitive to lower case, and also needs to be surrounded with quotes when generated from advanced filter gui.

phulin:
This is a first attempt at the Top Cards feature, currently only measured by draft ratings. Issues people may want to discuss:

- The link to the page is currently in the About menu which doesn't make much sense, but I don't want to add it as a standalone link to the main navbar for space reasons. Maybe we should move the About section to the bottom of each page instead? Most websites seem to do it this way.
- This PR adds a slightly-janky system using `rollup` to use front-end modules like `Filter.js` in backend code. I think this is fine for now but it does change the things that `npm start` runs. I've added better instructions in README.md for our Windows friends.

[Demo](https://d3qbb95kugbas0.cloudfront.net/tool/topcards)

Partially fixes #43 and incidentally fixes #546. Next planned addition is the cube count which will require some thinking about how best to implement.

Outstanding issues:

- [x] Add autocard support.
- [x] Fix `rollup` configuration so it doesn't clear the screen whenever it reloads.

Screenshot: 
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,16 @@ Then, run the following commands in the root of the cloned repository:
     npm install
     npm install nodemon -g
     node force_update.js
+
+If you are on Windows, you will need to make sure you have `bash` installed somewhere and run the following command [with your `bash` path in place of the path below].
+
+    npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"
+
+Then you can start the program like so:
+
     npm start
 
-You can now open up a browser and connect to the app through: http://localhost:8080
+You can now open up a browser and connect to the app through: http://localhost:8080. Despite the fact that node says it is running on port 5000, you should use port 8080 to connect.
 
 Nodemon will restart the application anytime there is a change to a source file.
 
--- a/routes/tools_routes.js
+++ b/routes/tools_routes.js
@@ -36,7 +36,10 @@ function matchingCards(filter) {
 
 function makeFilter(filterText) {
   if (!filterText || filterText.trim() === '') {
-    return { err: false, filter: [] };
+    return {
+      err: false,
+      filter: []
+    };
   }
 
   const tokens = [];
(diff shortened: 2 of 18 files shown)


dekkerglen:
Is this ready for review?

phulin:
Yep

README.md
```diff
@@ -41,9 +41,16 @@ Then, run the following commands in the root of the cloned repository:
...
     npm install nodemon -g
     node force_update.js
+
+If you are on Windows, you will need to make sure you have `bash` installed somewhere and run the following command [with your `bash` path in place of the path below].
+
+    npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"
+
+Then you can start the program like so:
+
     npm start
```
dekkerglen: I prefer the change in #554, as it is requires less steps for new contributors to setup the application for the first time. Maybe we can add both, and label this as an alternative for more advanced users.

routes/tools_routes.js
```diff
@@ -0,0 +1,133 @@
...
+    // TODO: pull out and use notPromoOrDigitalId in cube_routes.js
+    let nonPromo = possible.find(card => !card.promo && !card.digital && card.border_color != 'gold');
+    return nonPromo || possible[0];
+  });
+
+  return CardRating.find({
+    'name': {
+      $in: names,
+    },
+  }).then(ratings => {
```
dekkerglen: How expensive is this DB call? Is it possible we should be using the npm library `mongoose-paginate`?

phulin: The expensive part isn't the returning of results - it's the call itself, so pagination wouldn't help. I haven't found it to be a problem on local but creation of an index on the name field should make it perfectly fine performance-wise. If it's causing problems we can revisit.

phulin: Actually, scratch that - pagination would do something, but we need to see all the results anyway so it's not useful here.

README.md
phulin: OK

phulin: If you merge that one first I can deconflict

phulin: Although I should also note that there are now three processes that run on npm start so it's going to be pretty annoying for windows people without this. I can put both in, though.

dekkerglen:
Can you rebase to fix conflict?

PR #549, opened 2019-09-30
#546 Status Filter in Advanced view issue

night-day:
I tried two approaches to get the status filter working with Advanced View

1) The one in this PR was to add the quotes to the statuses that have more than one word, then remove the quotes on the dropdown options.  This doesn't seem to work as the value which goes in the Filter textbox ends up with no quotes.  If you remove the replace function, then the quotes show up in the dropdown, but also correctly show up in the Filter textbox

2) Forcing the quotes to be added after the status is selected in the dropdown within the HandleChange function.  This worked in that the Filter textbox shows the status with quotes around it, but breaks the dropdown prepopulation presumably since it doesn't match the original text which has no quotes

### DIFF
--- a/src/components/FilterCollapse.js
+++ b/src/components/FilterCollapse.js
@@ -76,8 +76,8 @@ const AdvancedFilterModal = ({ isOpen, toggle, apply, values, onChange, ...props
             <InputGroupText>Status</InputGroupText>
           </InputGroupAddon>
           <Input type="select" name="status" value={values.status} onChange={onChange}>
-            {['', 'Not Owned', 'Ordered', 'Owned', 'Premium Owned'].map(status =>
-              <option key={status}>{status}</option>
+            {['', '\"Not Owned\"', 'Ordered', 'Owned', '\"Premium Owned\"'].map(status =>
+              <option key={ status }>{status.replace(/[""]+/g, '')}</option>
             )}
           </Input>
         </InputGroup>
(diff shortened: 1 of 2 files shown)


src/util/Filter.js
```diff
@@ -579,7 +579,7 @@ function filterApply(card, filter) {
     });
   }
   if (filter.category == 'status') {
-    if (card.status.toLowerCase() == filter.arg) {
+    if (card.status == filter.arg) {
```
dekkerglen:
```suggestion
    if (card.status.toLowerCase() == filter.arg.toLowerCase()) {
```
Let's do this instead to prevent any further issues with case sensitivity

night-day:
Fixed and pushed

src/components/FilterCollapse.js
```diff
@@ -131,6 +131,10 @@ class FilterCollapse extends Component {
     const tokens = [];
     for (const name of allFields) {
       if (this.state[name]) {
+        if (name == "status")
+        {
+          this.state[name] = "\"" + this.state[name] + "\"";
+        }
```
phulin: So this pattern doesn't work in React - you need to use `this.setState` to change component state in React, and anyway you don't want to change the component state here (you don't want to put the quotes in the Advanced modal). But since this change is included in #542, I think you can just discard this change and the other one in this file.

phulin: I think you still need to remove this change

src/components/FilterCollapse.js
```diff
@@ -130,7 +130,7 @@ class FilterCollapse extends Component {
     // Advanced Filter change. Render to filter input.
     const tokens = [];
     for (const name of allFields) {
-      if (this.state[name]) {
+      if (this.state[name]) {
```
phulin: Trailing space here, sorry for being annoying about this but this should get removed

PR #554, opened 2019-10-01
fix: update seed.js to include cubes

haganbmj:
`seed.js` wasn't accounting for the cube model and wasn't documented in the Readme.

I also included a blurb about how the concurrent execution currently in the `package.json` won't work on Windows without something like [npm-run-all](https://www.npmjs.com/package/npm-run-all). The workaround for Windows users is to run each command in its own command prompt/powershell instance.


README.md
```diff
@@ -25,23 +25,36 @@ Atom (optional): https://atom.io/
...
+```sh
+npm install
+npm install nodemon -g
+node force_update.js        # Download Scryfall assets.
+node seed.js                # Seed the mongo database.
+
+# Linux/OSX users
+npm start                   # Start nodemon for backend server and webpack for frontend assets.
+
+# Windows users will need to use 2 terminal instances.
```
phulin: I have a blurb about this in #542 - the alternate solution is to use bash to run the command via npm config.

haganbmj: Yeah, that's a fair option as well.

PR #568, opened 2019-10-02
Feature 543

night-day:
Adding to api to send cube list in plain text


routes/cube_routes.js
```diff
@@ -2025,6 +2025,17 @@ router.get('/api/getcardfromcube/:id', function(req, res) {
   });
 });
 
+router.get('/api/cubelist/:id', function(req, res) {
+  Cube.findOne(build_id_query(req.params.id), function(err, cube) {
+    var cardnames = "";
+    cube.cards.forEach(function(item, index) {
+      cardnames += carddb.cardFromId(item.cardID).name + "\n";
+    });
```
phulin: You could write this more concisely -

const names = cards.map(card => carddb.cardFromId(card.cardID).name);
res.contentType...
res.status(200).send(names.join(‘\n’));

Issue #374, opened 2019-09-14
Notifications


dekkerglen:
As a user, I would like to receive notifications, and have a log of notifications, for other's actions related to my Cubes including drafting and commenting (if commenting is implemented first), and rating (if rating is implemented first).

If this feature is implemented after commenting, comments and all comment replies should also yield notifications.

Don't send emails. Nobody will like that

Issue #49, opened 2019-08-07
Follow Cubes / Feed


dekkerglen:
As a user, I would like to be able to follow other's cubes, and have a feed of changes to all the cubes I follow.

omnihound:
Would follows be a new model in the database or apart of the user model?

Issue #48, opened 2019-08-07
Notifications for Drafts


dekkerglen:
As a user, I would like to receive notifications when someone drafts my cube.

This can be done by sending an email, or by adding some elaborate UI and a notifications array in the user object - preferably both.

PR #573, opened 2019-10-03
discussion: Activity Feed and Notifications

haganbmj:
Just wanted to open up the conversation to see if there were opinions about the implementation of Activity Feeds, Notifications, etc.

Most of my experience stems from relational databases, so I'm probably a bit off on models here.

I'll throw some comments within the files, but I imagine Notifications have to follow a similar model to an ActivityFeed, with the challenge being how items are marked eligible for addition to that list.

Relates to #49, #48, and #374.

### DIFF
--- a/models/activity.js
+++ b/models/activity.js
@@ -1,0 +1,19 @@
+let mongoose = require('mongoose');
+
+let activitySchema = mongoose.Schema({
+  actor: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'User',
+    required: false
+  },
+  timestamp: {
+    type: Date,
+    required: true
+  },
+  referenceId: {
+    type: String,
+    required: true
+  }
+});
+
+let Activity = module.exports = mongoose.model('Activity', activitySchema)
--- a/models/activityfeed.js
+++ b/models/activityfeed.js
@@ -1,0 +1,19 @@
+let mongoose = require('mongoose');
+
+let activityFeedSchema = mongoose.Schema({
+  owner: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'User',
+    required: true
+  },
+  timestamp: {
+    type: Date,
+    required: true
+  },
+  activityId: {
+    type: mongoose.Schema.Types.ObjectId,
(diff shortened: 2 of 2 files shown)


models/activity.js
```diff
@@ -0,0 +1,19 @@
+let mongoose = require('mongoose');
+
+let activitySchema = mongoose.Schema({
+  actor: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'User',
+    required: false
```
haganbmj: Unsure here if this is correct, there likely should be a consideration for system level activity items - like website blog posts or updates. Those could be tied to a system User though.

models/activity.js
```diff
@@ -0,0 +1,19 @@
...
+  },
+  timestamp: {
+    type: Date,
+    required: true
+  },
+  referenceId: {
+    type: String,
```
haganbmj: This likely has to be generic to account for a variation of activity types.

Given a Deck object the activity item can infer the type of message based on that model.
Similarly, a Blog object contains enough information to build a different format of activity message.

models/activityfeed.js
```diff
@@ -0,0 +1,19 @@
+let mongoose = require('mongoose');
+
+let activityFeedSchema = mongoose.Schema({
+  owner: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'User',
```
haganbmj: Having a User object here likely makes this more of a NewsFeed, this goes back again to my non-relational database inexperience though, unsure if this reference id should exist on the parent or child or both.

models/activityfeed.js
```diff
@@ -0,0 +1,19 @@
...
+    required: true
+  },
+  timestamp: {
+    type: Date,
+    required: true
+  },
+  activityId: {
```
haganbmj: The alternative here of course is to have an array of Activity.

dekkerglen:
Are you on the discord? Most dev talk occurs there. 
https://discord.gg/Hn39bCU

@ dekkaru once you join to gain access to contributor channels

PR #576, opened 2019-10-03
Fixing p1p1 seeding

joshfinnie:
Was reviewing the code, and it seems as though the seed variable for p1p1 generation is not using the request parameter and I think it should.


dekkerglen:
Took me a while to realize this was an api route, that's probably why this hasn't raised any red flags yet.

PR #577, opened 2019-10-03
*WIP* updating the cube route to show a different way of doing it

joshfinnie:
This PR is generated to allow for a conversation on a way to refactor the routes code that might help the project as it continues to grow.

This approach modularizes the cube routes breaking them into multiple files, 1 isolates all the api views for the cube, 1 isolates all the standard views for the cube. I also added a helpers file that keeps all the helper functions that were not specifically a route out of the route files.

I think this will be beneficial to this project for a few reasons, 1) there will be less merge conflicts when changing view logic 2) your frame of reference is more focused on a specific subsection of views (i.e. do not lose an API change because it was on line 3000 of a file) 3) it shows us what needs to still be tested in the application (helpers should probably have a high test coverage)

Hope that helps clear up my thought pattern here.


emmett9001:
A short summary of how your approach differs from the existing one would make it easier to comment on this relatively large diff.

PR #588, opened 2019-10-07
Break root routes into separate file.

phulin:
Self-explanatory. This should facilitate testing and also keeps stuff out of app.js, which now just has config + basic info.


app.js
```diff
@@ -129,308 +121,16 @@ let cubes = require('./routes/cube_routes');
 let users = require('./routes/users_routes');
 let devs = require('./routes/dev_routes');
 let tools = require('./routes/tools_routes');
+app.use('', require('./routes/root'));
```
dekkerglen: Shouldn't this come after the other app.use? Wouldn't all routes just get routed into here?

phulin:
No, it checks those routes and if none of them match, it falls through.

PR #594, opened 2019-10-11
Add featured cubes section

dekkerglen:
This adds featured cubes sections. This was a prerequisite to become SCG's cube partner, so I'll be looking to merge this despite the feature freeze.


dekkerglen:
Those are just from running npm update, so it shouldn't be an issue

dekkerglen:
I agree, this is fine for now, but we should switch to actual role based auth in the future

phulin:
@dekkerglen just want to flag for you that you can run `npm install` to get new packages without upgrading all of our packages and triggering package-lock changes, maybe you intended to do that but just in case you didn't

PR #597, opened 2019-10-13
Pull tag colors modal into React.

phulin:
Outstanding issues:

- [x] Fix compare view.


phulin:
I think this is good to go now

PR #617, opened 2019-10-19
RSS updates

omnihound:
Not that it should make much of a difference but I've noticed other feeds reverse order the dates so I've done that here

also I've appended the changelist from the blog post after noticing alot of posts being empty


routes/cube_routes.js
```diff
@@ -676,10 +676,24 @@ router.get('/blog/:id/rss', function(req, res) {
             site_url: 'https://cubecobra.com',
           });
 
+
+
```
phulin: remove these?

omnihound: 👍

PR #619, opened 2019-10-19
Moved blog posts to React, Add Commenting

### LINKED ISSUE #374: Notifications

dekkerglen:
Fixes #374, adds commenting.
Comments can have a maximum response depth of 6. Comments can be edited. Comments show how long ago the comment was posted, as well as a reference to the user's page. Comments can be deleted, reddit style, so the content and reference to the name will be removed, but the comment item + replies will continue to exist. Comments paginate with a max page size of 10.


src/components/BlogPost.js
```diff
@@ -0,0 +1,78 @@
...
+  constructor(props) {
+    super(props);
+    
+    this.toggle = this.toggle.bind(this);
+    this.state = {collapse: false};
+  }
+  
+  toggle() {
+    this.setState({ collapse: !this.state.collapse });
+  }
```
omniczech: `bind` is a bit fragile, if you just do `toggle = () => {` you get the same result and less lines

src/components/BlogPost.js
```diff
@@ -0,0 +1,78 @@
...
+                </p>
+                }
+                {!post.body ? '' :
+                <p className="card-text">{post.body}</p>
+                }
+                {!post.html ? '' :
+                <div className="card-text">
+                    <a dangerouslySetInnerHTML={{__html: post.html}}></a>
+                </div>
+                }
```
omniczech: Unless all of these are commonly blank, use phrasing similar to this `{post.changelist &&(<HTML>)}` so you're not returning blank strings for them.

src/components/Paginate.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+import { Pagination, PaginationItem, PaginationLink, Collapse } from 'reactstrap';
+
+class Paginate extends React.Component {
+  constructor(props) {
+    super(props);
+    
+    this.state = {collapse: false};
```
omniczech: State is unused, remove

src/cube_blog.js
```diff
@@ -0,0 +1,49 @@
...
+
+class CubeBlog extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+        posts: this.props.posts,
+        pages: this.props.pages,
+        canEdit: this.props.canEdit
+      };
```
omniczech: Not sure what this state is meant to accomplish, looks to be unused

src/cube_blog.js
```diff
@@ -0,0 +1,49 @@
...
+  componentDidMount() {
+    updateBlog();
+  }
+
+  select(nav) {
+    this.setState({ });
+  }
+
+  render() {
+    console.log(this.props.posts);
```
omniczech: remove log

src/cube_blog.js
```diff
@@ -0,0 +1,49 @@
...
+        <>
+          {!(this.props.pages && this.props.pages.length > 1) ? '' :
+            <Paginate pages={this.props.pages} />
+          }
+          {this.props.posts.map(post =>
+            <BlogPost post={post} canEdit={this.props.canEdit}/>
+          )}
+          {!(this.props.pages && this.props.pages.length > 1) ? '' :
+            <Paginate pages={this.props.pages} />
+          }
```
omniczech: Same as previous comment, check for presence not lack thereof

src/cube_blog.js
```diff
@@ -0,0 +1,49 @@
...
+
+  componentDidMount() {
+    updateBlog();
+  }
+
+  select(nav) {
+    this.setState({ });
+  }
+
+  render() {
```
omniczech: destructuring props would help here `const {pages, posts, canEdit} = this.props` means you can change these variables from `this.props.VAR` to `VAR` in the rest of the file

src/components/BlogPost.js
dekkerglen: I don't understand, what lines do I need to change?

omniczech: Ditch line 8, change like 12 to the commented code, should work the same.

phulin: I think we don't have that transform (class-props or whatever it's called) turned on in babel, unless it's changed what is on by default since the last time i checked. But we certainly could do it

src/components/BlogPost.js
```diff
@@ -0,0 +1,78 @@
...
+            </div>
+            }
+            {this.props.canEdit &&
+            <div className="card-footer">
+                <span>
+                <button className="btn btn-success mr-2 editBlogButton" role='button' data-id={post._id}>Edit</button>
+                <a> </a>
```
phulin: Nitpick but you can just use `{' '}` here

src/components/Paginate.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+import { Pagination, PaginationItem, PaginationLink, Collapse } from 'reactstrap';
```
phulin: You're not using the Collapse import here

src/cube_blog.js
```diff
@@ -0,0 +1,43 @@
...
+import BlogPost from './components/BlogPost';
+import Paginate from './components/Paginate';
+
+class CubeBlog extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  componentDidMount() {
+    updateBlog();
```
phulin: This is a pretty fragile way to do a trigger and will lead to weird issues when the component updates - can you move the functions into event handlers in `BlogPost` instead?

anderoonies: also could `blog.js` be a module that exports `updateBlog` rather than being on global namespace? to keep `updateBlog` from being a mystery function

routes/cube_routes.js
```diff
@@ -545,114 +545,92 @@ router.get('/blogsrc/:id', function(req, res) {
...
+    var pages = [];
+    var blog_page = [];
+    if (blogs.length > 0) {
+      blogs.reverse();
+
+      var page = parseInt(req.params.page);
+      if (!page) {
+        page = 0;
+      }
+      for (i = 0; i < blogs.length / 10; i++) {
```
anderoonies: `var i` to prevent global

routes/cube_routes.js
```diff
@@ -545,114 +545,92 @@ router.get('/blogsrc/:id', function(req, res) {
...
-      });
+        } else {
+          pages.push({
+            url: '/cube/blog/' + cube_id + '/' + i,
+            content: (i + 1)
+          });
+        }
+      }
+      blog_page = [];
+      for (i = 0; i < 10; i++) {
```
anderoonies: `var i` to prevent global

src/cube_blog.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  select(nav) {
```
anderoonies: is this doing anything?

src/cube_blog.js
```diff
@@ -0,0 +1,43 @@
...
+
+  render() {
+    const {pages, posts, canEdit} = this.props
+    return (
+      <>
+        {(pages && pages.length > 1) &&
+          <Paginate pages={pages} />
+        }
+        {posts.map(post =>
+          <BlogPost post={post} canEdit={canEdit}/>
```
anderoonies: these should have `key`s i think

src/cube_blog.js
phulin: Yeah my vision would be to get rid of updateBlog altogether and have all the DOM manipulation + setting of event handlers done by react

src/components/Paginate.js
```diff
@@ -0,0 +1,29 @@
...
+
+class Paginate extends React.Component {
```
phulin: This could just be a function component; doesn't need to be a full class.

src/components/Paginate.js
```diff
@@ -0,0 +1,29 @@
...
+  constructor(props) {
+    super(props);
+  }  
+  
+  render() {
+    return (
+        <>
+        <hr/>
+        <Pagination aria-label="Table page" className="mt-3">
+          {this.props.pages.map(page =>
```
phulin: This is a design point and not functional, but I generally try to avoid passing these kinds of non-trivial data structures down as props. Could you instead just pass a url prefix and the page count to this component and then have it do the rest?

src/cube_blog.js
anderoonies: gotcha, +1

src/components/Paginate.js
anderoonies: my understanding is that react passes everything other than primitives by reference, so there shouldn't be a cost of passing large structures as props?

phulin: Of course - it's not a performance issue, just awkward design from my point of view

src/cube_blog.js
dekkerglen: Yes, I'm slowly moving blog.js into react as well

src/cube_blog.js
dekkerglen: I don't understand, can you elaborate?

src/components/Paginate.js
dekkerglen: Sorry, I don't understand, how should I change it?

src/cube_blog.js
anderoonies: any elements in an array (like `BlogPost`) should have a `key` attribute to uniquely identify the element, i'm not positive what the data of `post` is (just `blogSchema?`) but something unique

src/components/Paginate.js
anderoonies:
```jsx
const Paginate = ({pages}) => {
  return (
    <html stuff>
  )
}

File: src/cube_blog.js [hunk 17]
```diff
@@ -0,0 +1,41 @@
...
+        }
+      </>
+    );
+  }
+}
+
+const posts = JSON.parse(document.getElementById('blogData').value);  
+const pages = JSON.parse(document.getElementById('blogPages').value); 
+const loggedIn = document.getElementById('userid') != null;
+var userid = '';
```
joshfinnie: this could also be written as a ternary operation...

```
const userid = loggedIn ? document.getElementById('userid').value : '';
```

dekkerglen: Right on, that's so much cleaner using ternary

src/components/AgeText.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+
+class AgeText extends React.Component {
+    constructor(props)
```
phulin:
```suggestion
    constructor(props) {
```

phulin: For consistency

src/components/AgeText.js
```diff
@@ -0,0 +1,97 @@
...
+            if(parseInt(str) > 1)
+            {
+                str += 's ago';
+            }
+            else
+            {
+                str += ' ago';
+            }
+        }
+        return (
```
phulin:
```suggestion
        return str;
```

src/components/BlogContextMenu.js
```diff
@@ -0,0 +1,74 @@
...
+    csrfFetch("/cube/blogsrc/" + post._id)
+      .then(response => response.json())
+      .then(function(json) {
+        if (json.src) {
+          $('#editor').html(json.src);
+        } else {
+          $('#editor').html(json.body);
+        }
+
+        $('#postBlogTitleInput').val(json.title);
```
phulin: Optimally this state would be controlled by react - we would have an element `<div contenteditable value={...} onChange={...} />`. But this is fine for now.

phulin: I've also been trying to keep jQuery out of the react code with the eventual goal of removing it as a dependency.

src/components/BlogPost.js
```diff
@@ -0,0 +1,116 @@
...
+import CommentEntry from './CommentEntry';
+
+class BlogPost extends React.Component {
+  constructor(props) {
+    super(props);    
+    
+    this.onPost = this.onPost.bind(this);
+    this.submitEdit = this.submitEdit.bind(this);
+
+    this.childElement = React.createRef();
```
phulin: This isn't a react-y way of making this work - instead of having a parent call a function on the child, have the parent manage the child's state and then pass down the value + a toggle callback. I've avoided using refs anywhere in the rest of the react code, and I think that's a good constraint to keep unless there's a really good reason not to.

See [here](https://reactjs.org/docs/lifting-state-up.html)

src/components/BlogPost.js
```diff
@@ -0,0 +1,116 @@
...
+
+  async submitEdit(comment, position)
+  {
+    //update current state
+    this.saveEdit(this.props.post.comments, position, comment);
+  }
+
+  render() {
+    var post = this.props.post;
+    return (
```
phulin: Can you make sure that the indentation is consistently two spaces here? Although maybe this comment should just wait until we have prettier set up.

src/components/CommentsSection.js
```diff
@@ -0,0 +1,52 @@
...
+    this.setState({ collapse: !this.state.collapse });
+}
+
+expand() {
+    this.setState({ collapse: true });
+}
+
+render() {
+    var comments = this.props.comments;
+    return (
```
phulin: Can just do `return comments.length == 0 ? '' : ....`; no need for the `<>`.

views/index.pug
```diff
@@ -26,7 +26,7 @@ block content
 										a(href='/cube/overview/'+get_cube_id(cube))
 											img.card-img-top(src=cube.image_uri style='width: 100%;height: auto;')
 										em.text-right(style='padding: 2px;') Art by: #{cube.image_artist}                    
-										a.btn.btn-success(href='/cube/overview/'+get_cube_id(cube)) View
+										a.btn.btn-success.rounded-0(href='/cube/overview/'+get_cube_id(cube)) View
```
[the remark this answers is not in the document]
dekkerglen: This is only applicable to these buttons, otherwise I would agree

src/components/BlogContextMenu.js
dekkerglen: I agree this is an ongoing goal, but my focus here is just to add commenting. It's already a bloated PR, so I will leave this here for now

src/components/BlogPost.js
dekkerglen: I'll wait for prettier - there is a lot of inconsistency between files for whitespace

routes/cube_routes.js
```diff
@@ -545,114 +545,92 @@ router.get('/blogsrc/:id', function(req, res) {
...
+      ),
+      loginCallback: '/cube/blog/' + req.params.id
+    });
+  } catch (err) {
+    return res.status(500).send(err);
+  }
 });
 
-router.get('/blog/:id/rss', function(req, res) {
+router.get('/rss/:id', function(req, res) {
```
joshfinnie: what's the motivation for the url change here? I don't think i minded `/blog/:id/rss`...

joshfinnie: there might be people where this is a breaking change?

src/components/PagedList.js
```diff
@@ -0,0 +1,56 @@
...
+    autocard_init('autocard');
+  }
+
+  render() {
+    const { pageSize, rows, children, ...props } = this.props;
+    const { page } = this.state;
+    const displayRows = rows.slice(page * pageSize, (page + 1) * pageSize);
+    const validPages = [...Array(Math.ceil(rows.length / pageSize)).keys()];
+
+    return <>
```
joshfinnie: this is a bit bike-shedding, but i think this would work and remove the need for a react.fragment here:

```
return {validPages.length === 1 
  ? <></>
  : <Pagination aria-label="Table page" className="mt-3">
        {validPages.map(page =>
          <PaginationItem key={page} active={page === this.state.page}>
            <PaginationLink tag="a" href="#" page={page} onClick={this.setPage}>
              {page + 1}
            </PaginationLink>
          </PaginationItem>
        )}
      </Pagination>}
```

This would need to be tested since it's kind of just off the top of my head...

PR #623, opened 2019-10-21
Adding the ability to follow cubes

joshfinnie:
Currently, this PR adds a follow link to cubes that are not yours. This will allow you to follow them and have them show up on your profile.

There are some things that are not perfect here, currently the follow/unfollow causes a page refresh because I am bad at front-end. If anyone wants to take a look at how to make that better, awesome!

Also, this is a very simplistic approach. Glad to discuss other options.


public/js/followcube.js
```diff
@@ -0,0 +1,11 @@
...
+  await fetch(`/cube/api/follow/${cubeID}`);
+  await console.log(`followed ${cubeID}`);
+  await location.reload();
+}
+
+const unfollow = async (cubeID) => {
+  await fetch(`/cube/api/unfollow/${cubeID}`);
+  await console.log(`unfollowed ${cubeID}`);
+  await location.reload();
+}
```
anderoonies: shouldn't need to `await location.reload()` because you're not doing anything after it, also remove console.log

models/user.js
```diff
@@ -30,7 +30,8 @@ const UserSchema = mongoose.Schema({
     type: Boolean,
     default: false
   },
-  edit_token: String
+  edit_token: String,
+  followed_cubes: [Number]
```
anderoonies: maybe call this `followed_cube_ids` to be more descriptive of what the data actually is

routes/cube_routes.js
```diff
@@ -2974,4 +2976,34 @@ router.get('/api/p1p1/:id/:seed', function(req, res) {
...
+router.get('/api/follow/:id', function(req, res) {
+  User.findOneAndUpdate({
+      username: req.user.username
+    }, {
+      $addToSet: {
+        followed_cubes: String(req.params.id)
+      }
+    },
+    function(err, success) {
+      if (err) console.log(err);
```
anderoonies: this will still send a success even if it errors which we might not want

routes/cube_routes.js
```diff
@@ -2974,4 +2976,34 @@ router.get('/api/p1p1/:id/:seed', function(req, res) {
...
+router.get('/api/unfollow/:id', function(req, res) {
+  User.findOneAndUpdate({
+      username: req.user.username
+    }, {
+      $pullAll: {
+        followed_cubes: [String(req.params.id)]
+      }
+    },
+    function(err, success) {
+      if (err) console.log(err);
```
anderoonies: same here about giving a success code even on error

dekkerglen:
I'm closing this PR, as I've incorporated most of the functionality into branch #feature-374

PR #634, opened 2019-10-26
Added isToken field to getTCGLink

### LINKED ISSUE #633: Token tcgplayer links should append 'token' if tcgplayer id is missing
The links in the token analysis point to the name of the token instead of "<token name> token".
so if i click the rat token i get a search of "rat" that does not find any tokens instead of the "rat token" search i need
they use the function that creates the tcgplayer link
which uses the tcgplayer ID to point to the actual card page, but if that is missing, then it does a search by card name

dekkerglen:
fixes #633


src/util/Affiliate.js
```diff
@@ -1,9 +1,16 @@
-export function getTCGLink(card) {
+export function getTCGLink(card, isToken) {
   let tcgplayerLink = 'https://shop.tcgplayer.com/';
   if (card.details.tcgplayer_id) {
     tcgplayerLink += `product/productsearch?id=${card.details.tcgplayer_id}`;
   } else {
-    tcgplayerLink += `productcatalog/product/show?ProductName=${card.details.name}`;
+    if(isToken)
```
phulin:
```suggestion
    tcgplayerLink += `productcatalog/product/show?ProductName=${card.details.name}`;
    if (isToken) {
      tcgplayerLink += ' Token';
    }
```

PR #639, opened 2019-10-28
feat(views): add link to drafter profile on decks

haganbmj:
Adds a link to the drafter's profile to the deck display page.

decks-profile-link

decks-profile-link-anon

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -2399,16 +2399,32 @@ router.get('/deck/:id', function(req, res) {
           req.flash('danger', 'Cube not found');
           res.status(404).render('misc/404', {});
         } else {
-          var owner_name = "Unknown";
-          var drafter_name = "Anonymous";
-          User.findById(deck.owner, function(err, drafter) {
-            if (drafter) {
-              drafter_name = drafter.username;
+          let owner = {
+            name: 'Unknown',
+            id: null,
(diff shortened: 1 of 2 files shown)


dekkerglen:
You can ignore the CI failure, it's been fixed in another PR.

I'm happy to merge this if you would like, but there are a couple changes I'd like to see us working towards with the route functions, specifically moving towards using `await` instead of callbacks. This avoids a lot of repeated code, and makes the functions a lot more readable.

We could also do several DB calls in parallel, and wait for all of them to finish by doing:
`[result1, result2] = await Promise.all([async1(), async2()]);`

Where `async1()` and `async2()` are the DB calls.

Don't feel obligated to make these changes now, this is still a good change. Just let me know if you want me to merge now, or if you want to make the suggested changes. (either is ok)

haganbmj:
I used the existing non-promisified queries since there were no changes to them and I have a habit of over-scoping my pull requests.
I'd be happy to go back and refactor this and a few more methods in a separate pull request.

haganbmj:
Taking a stab rewriting that method.
https://github.com/haganbmj/CubeCobra/commit/3ffdba9b5ef08d836da01487fbefba9b4fb2341e

Also taking the opportunity to add some mongoose virtuals, but with the way the fields are named it's probably better to just use populate for most things.

PR #644, opened 2019-10-31
inital setup of prettier and eslint using airbnb standards

joshfinnie:
This is the start of getting `prettier` and `eslint` set up for CubeCobra.

I went through `app.js` to show some of the changes that would be required to get a sense of them.


phulin:
This looks really great. I assume we need a second commit making all the changes in the rest of the codebase?

joshfinnie:
@phulin yea, we can either make a massive update or we can just submit these rules and have the developers start implementing the changes on PRs for other work.

phulin:
My instinct is that we want one massive PR that also puts it into the checks that Travis runs.

joshfinnie:
@dekkerglen this is ready for merger

PR #649, opened 2019-11-01
Fix advanced filter issues.

### LINKED ISSUE #632: Advanced filter issues
typing in the Mana Cost field "{W}" then apply searches for mana:{W}, which returns nothing, CMC "not equal to" 2 searches for "cmc!2" which returns nothing as well (should be -cmc=2 ?), including/exactly those colors/color identities doesn't fill up the filter.

### LINKED ISSUE #638: OR is case sensitive in filter syntax
Please .toLowerCase() or in the filter syntax

phulin:
Fixes #638. 
Fixes #632.

### DIFF
--- a/src/components/FilterCollapse.js
+++ b/src/components/FilterCollapse.js
@@ -36,7 +36,7 @@ const NumericField = ({ name, humanName, placeholder, valueOp, value, onChange, 
       <option value=">">greater than</option>
       <option value="<=">less than or equal to</option>
       <option value=">=">greater than or equal to</option>
-      <option value="!">not equal to</option>
+      <option value="!=">not equal to</option>
     </Input>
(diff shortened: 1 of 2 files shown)


src/components/FilterCollapse.js
```diff
@@ -36,7 +36,7 @@ const NumericField = ({ name, humanName, placeholder, valueOp, value, onChange,
       <option value=">">greater than</option>
       <option value="<=">less than or equal to</option>
       <option value=">=">greater than or equal to</option>
-      <option value="!">not equal to</option>
+      <option value="!=">not equal to</option>
```
dekkerglen: for not equal to, shouldn't we use '=' with '-' prepended to the entire term?

purxiz: yes, != doesn't work with the current filter syntax afaik. Or if it does, I have no idea how

PR #652, opened 2019-11-01
Rework draftbots.

phulin:
This PR reworks the draftbots to have a more fluid prioritization of on-color cards. They still prioritize on-color cards, but only by giving a boost to that card's ratings. Fetchlands that are single-on-color are special-cased, and lands that are single-on-color are prioritized ahead of spells that are single-on-color.

### DIFF
--- a/public/js/draft.js
+++ b/public/js/draft.js
@@ -88,6 +88,9 @@ function botRating(botColors, card) {
   const overlap = botColors.some(c => colors.includes(c));
   const isLand = card.type_line.indexOf('Land') > -1;
   const isFetch = fetchLands.includes(card.details.name);
+
+  // Prioritize on-color or overlapping fetches.
+  // Then overlapping lands, then overlapping spells.
   if (subset || (isFetch && overlap)) {
     rating -= 0.4;
   } else if (isLand && overlap) {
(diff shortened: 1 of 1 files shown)


public/js/draft.js
```diff
@@ -64,72 +64,66 @@ function arrayRotate(arr, reverse) {
   return arr;
 }
 
+function arrayIsSubset(needles, haystack) {
+  return needles.every(x => haystack.includes(x));
+}
```
dekkerglen: cute

public/js/draft.js
```diff
@@ -64,72 +64,66 @@ function arrayRotate(arr, reverse) {
...
+  'Arid Mesa',
+  'Bloodstained Mire',
+  'Flooded Strand',
+  'Marsh Flats',
+  'Misty Rainforest',
+  'Polluted Delta',
+  'Scalding Tarn',
+  'Verdant Catacombs',
+  'Windswept Heath',
+  'Wooded Foothills',
```
RKelln:
```suggestion
  'Wooded Foothills',
  'Bant Panorama',
  'Esper Panorama',
  'Grixis Panorama',
  'Jund Panorama',
  'Naya Panorama',
```
Do you want to include the 3 colour fetch panoramas?  From https://mtg.gamepedia.com/Fetch_land

dekkerglen: I don't think so. I think it might be worth to include prismatic vista though.

phulin: My assumption was people would have vista in the colorless section and therefore it wouldn't be a problem. This special case is specifically for one-on-color fetches, not for others.

Issue #552, opened 2019-10-01
Card Viewer Showing Blank Line


cyandev:
**Describe the bug**
When moving the mouse fast enough, the card display doesn't load an image and then doesn't disappear, and is a white line attached to the pointer.

**To Reproduce**
1. Go to a cube
2. Move your mouse very fast along the left edge of the white section, loading many card images, and quickly move it away.
3. The white line will be stuck to your pointer. It is very visible when you move your mouse near cards.

**Expected behavior**
The card display should disappear when hovering is done and the image isn't loaded.

**Additional context**
This may be an issue with events, I'm not familiar with the project and how the card display works, just a potential cause.

cyandev:
Somehow deleted the section saying this is on windows 10 desktop.

phulin:
Thanks - this is a duplicate of #488

dekkerglen:
Closed #488, use that issue as a reference when working on this one

cyandev:
If nobody is working on this can someone assign me? I haven't worked on anything here but I am interested in helping out. My experience is extremely limited.

PR #660, opened 2019-11-06
Fix bottom of autocard popup lingering

tspigot:
This issue (#552) was caused by the image unloading handler being run
erroneously after the autocard_hide handler attempts to remove its
events. This is difficult to spot in most cases but becomes easily
reproducible if the tester disables caching and sets throttling to
something very low (on Firefox/Windows 10, "Regular 2G" did the trick...
set this after the cube list loads or you could be waiting a while).
On fast network connections, this would rarely be an issue, as the image
would load properly faster than the user could move their mouse off the
image, and browser caching prevents them from trying again on the same
card.

The issue was that innerHtml on the image container was being cleared
before the event handlers were canceled. Clearing this property removes
the child nodes from the document, meaning the queries in the lines
afterward do not find them and thus do not remove the event handler. 
The event handler fires, showing the (now-empty) popup if the
tags variable was captured with something in it.

Clearing the events properly fixes the bug, and it can no longer be
reproduced even with throttling/nocache.

### DIFF
--- a/public/js/autocard.js
+++ b/public/js/autocard.js
@@ -138,14 +138,14 @@ function autocard_show_card(card_image, card_flip, show_art_crop, tags) {
 }
 
 function autocard_hide_card() {
-  document.getElementById("autocard_popup").innerHTML = '';
-  document.getElementById("autocard_popup2").innerHTML = '';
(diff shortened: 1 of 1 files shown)


phulin:
This is awesome, thanks!!

PR #661, opened 2019-11-06
Fix for missing tokens

### LINKED ISSUE #630: Missing tokens
Catch all bug report for missing or incorrect token suggestions.
Just comment below if you find more errors.

Missing tokens for card:
Force of rage.

Balkmeister:
Fixes #630


serverjs/updatecards.js
```diff
@@ -233,6 +244,55 @@ function addTokens(card) {
         }
       }
     }
+    if (_catalog.dict[card.id].oracle_text.includes('Ascend (')) {
+      mentionedTokens.push({
+        tokenId: getTokenIDForSpecialCaseToken("City's Blessing"),
+        sourceCardId: _catalog.dict[card.id]._id
+      });
+    }
+    if (_catalog.dict[card.id].oracle_text.includes('Infect (')) {
```
purxiz: Does scryfall/oracle add the reminder text to all static abilities? They aren't printed on every card. Also, do you handle other sources of poison counters like Caress of Phyrexia?

serverjs/updatecards.js
```diff
@@ -463,10 +539,10 @@ var specialCaseCardsList = {
     tokenId: "e6fa7d35-9a7a-40fc-9b97-b479fc157ab0"
   }],
   "Westvale Abbey": [{
-    tokenId: " 94ed2eca-1579-411d-af6f-c7359c65de30",
+    tokenId: "94ed2eca-1579-411d-af6f-c7359c65de30",
```
purxiz: Is there any way we can move this object to a data file i.e. json rather than in the source code?

serverjs/updatecards.js
```diff
@@ -463,10 +539,10 @@ var specialCaseCardsList = {
...
   }]
 };
-var specialCaseTokensList = ["Food"];
+var specialCaseTokensList = ["Food", "Treasure"];
 
 function getTokensForSpecialCaseCard(newCardid, card) {
```
phulin: This function could just be defined as an object and could also be moved into a separate json file.

serverjs/updatecards.js
Balkmeister: True, I changed the condition to "poison counter" since that is the way they are always talked about in the rules text. So this should cover all poison effects, even those without Infect.

Balkmeister: Oh, and Ascend is never mentioned without reminder text.

PR #662, opened 2019-11-06
Apply prettier to entire codebase.

phulin:
This PR applies `prettier` to the whole codebase. This brings all of the React/JSX code inside the beautify bubble. Prettier is also significantly more aggressive than js-beautify about formatting code, which should remove stylistic discrepancies between our wide array of our authors. The PR also updates Travis to test via prettier instead of the old js-beautify.


__tests__/serverjs/cards.test.js
```diff
@@ -109,106 +110,108 @@ test("cardFromId returns a placeholder card object when given a nonexistent ID",
...
   });
 });
 
-test("loadJSONFile loads a JSON file into the correct attribute", () => {
+test('loadJSONFile loads a JSON file into the correct attribute', () => {
   expect.assertions(1);
-  const attribute = "testAttribute";
-  return carddb.loadJSONFile(fixturesPath + "/names.json", attribute).then(function() {
+  const attribute = 'testAttribute';
+  return carddb.loadJSONFile(fixturesPath + '/names.json', attribute).then(function() {
```
joshfinnie: should this not be:

```
return carddb.loadJSONFile(`${fixturesPath}/names.json`, attribute).then(function() {
```

phulin: Generally yes but I don't think prettier does that transformation. We've got a lot of code written in the sums-of-strings style.

joshfinnie: that's fair... this looks really good @phulin

PR #671, opened 2019-11-08
Updated one shot scripts to cache and batch (batch and cache?)

dekkerglen:


one_shot_scripts/add_blog_names.js
```diff
@@ -7,22 +7,52 @@ const {
...
+                {
+                    let blog = await cursor.next()
+                    if(blog) {
+                        blogs.push(blog)
+                    }
+                }
+            }
+            await Promise.all(blogs.map(blog => addVars(blog)));
         }
+        console.log("Finished: " + count + " of " + count + " blogs");
```
RKelln: This log is out the for, the one in add_deck_names is inside, which do you prefer?

RKelln: Pretty sure you want inside?

PR #676, opened 2019-11-11
Feature 374

### LINKED ISSUE #48: Notifications for Drafts

dekkerglen:
Fixes #48 
Adds notifications for:

- Users replying to your comments
- Users drafting your cube
- Users commenting on your blogpost

Also:

- Added old notifications page (caps to 200)
- adds dev blog posts to everyone's activity
- fixes an issue where the oldest 50 blog posts would be pulled, not the most recent, for the activity feed
- added scroll bar and max height to blog posts
- improved cube preview css
- add route for viewing single blogposts
- Updated deck preview


routes/cube_routes.js
```diff
@@ -2074,8 +2117,10 @@ router.post('/api/editcomment', ensureAuth, async function(req, res) {
 });
 
 router.post('/api/postcomment', ensureAuth, async function(req, res) {
-  user = await User.findById(req.user._id);
-  post = await Blog.findById(req.body.id);
+  const userq = await User.findById(req.user._id);
+  const postq = await Blog.findById(req.body.id);
+
+  const [user, post] = await Promise.all([userq, postq]);
```
joshfinnie: i think we we do this, we do not need the `await`s above...

routes/cube_routes.js
```diff
@@ -2277,32 +2342,37 @@ router.get('/api/cubelist/:id', function(req, res) {
   });
 });
 
-router.post('/editdeck/:id', function(req, res) {
-  Deck.findById(req.params.id, function(err, deck) {
+router.post('/editdeck/:id', async (req, res) => {
+  try {
+    const deck = await Deck.findById(req.params.id);
+
     if (err || !deck) {
```
RKelln: Where does `err` come from?

public/css/stylesheet.css
```diff
@@ -451,3 +450,73 @@ body.busy-cursor {
...
+	left: 0;
+	bottom: 0;
+	right: 0;
+}
+.cube-preview-artist {
+	font-size: small;
+	position: absolute;
+	bottom: 0;
+	right: 2px;
+    color: white;
```
RKelln: Some indentation disagreement in this file (if that bothers you).

routes/dev_routes.js
```diff
@@ -110,19 +112,17 @@ router.post('/blogpost', ensureAuth, function(req, res) {
...
-      res.status(404).render('misc/404', {});
+      await blogpost.save();
+
+      req.flash('success', 'Blog post successful');
+      res.redirect('/dev/blog');
     }
-  });
+  } catch (err) {
+    res.status(500).send({
+      success: 'false'
```
RKelln: No `message: err`?

routes/users_routes.js
```diff
@@ -34,12 +34,59 @@ function checkUsernameValid(req) {
...
+    }
+
+    const notification = user.notifications.splice(req.params.index, 1)[0];
+    await user.save();
+
+    return res.redirect(notification.url);
+  } catch (err) {
+    res.status(500).send({
+      success: 'false'
+    });
```
RKelln: `message: err`?

src/notifications.js
```diff
@@ -0,0 +1,44 @@
...
+import Notification from './components/Notification';
+
+import { Card, Col, Row, CardHeader , CardBody, CardFooter} from 'reactstrap';
+
+
+class Notifications extends Component {
+  constructor(props) {
+    super(props);
+    
+    this.newCube = this.newCube.bind(this);
```
RKelln: This seems extraneous? (And `newCube()` below)

models/user.js
```diff
@@ -42,7 +42,27 @@ const UserSchema = mongoose.Schema({
   users_following: {
     type: [String],
     default: []
-  }
+  },
+  notifications: {
+    type: [{
+      user_from: String,
```
RKelln: I might rename to `from_user`, as this seems to be used below as:
`notification.from_user` is the user which sent the notification, which reads better than `notification.user_from`
Similarly, `from_username` (or with underscore if that is consistent elsewhere)

src/components/BlogPost.js
```diff
@@ -56,6 +56,16 @@ class BlogPost extends React.Component {
     this.saveEdit(this.props.post.comments, position, comment);
   }
 
+  componentDidMount() {
+    if(this.props.focused)
+    {
+        var $container = $("html,body");
+        var $scrollTo = $('.comment-highlighted');
+        
+        $container.animate({scrollTop: $scrollTo.offset().top - $container.offset().top + $container.scrollTop(), scrollLeft: 0},300);
```
RKelln: Extra fancy would make the duration slightly dependent on distance:
```javascript
const distance = $scrollTo.offset().top - $container.offset().top;
const duration = 200 + Math.min(200, distance / 10); // 200 - 400ms depending on distance 
$container.animate({scrollTop: distance + $container.scrollTop(), scrollLeft: 0}, duration);
```

models/user.js
purxiz: I might even use a more generic term like "source" and "source_description" in case of cases where the system generates notifications, etc.

public/css/stylesheet.css
```diff
@@ -451,3 +450,73 @@ body.busy-cursor {
...
 }
+.cube-preview-card {
+	width: 100%;
+	padding-top: 110%; /* 1:1 Aspect Ratio */
+	position: relative; /* If you want text inside of it */
+}
+.cube-preview-card:hover {
+	background: #EFEFEF;
+	cursor: pointer;
+    border: 1px solid #000000;
```
purxiz: prettier?

routes/cube_routes.js
```diff
@@ -2031,6 +2031,16 @@ function insertComment(comments, position, comment) {
...
 }
 
+function getOwnerFromComment(comments, position, comment) {
+  if (position.length <= 0) {
+    return '';
+  } else if (position.length == 1) {
+    return comments[position[0]].owner;
+  } else {
+    return getOwnerFromComment(comments[position[0]].comments, position.slice(1), comment);
+  }
```
purxiz: Why use recursion here instead of a for loop? it'll just muddy up the call stack.

purxiz: Also, what is the comment argument doing here?

routes/cube_routes.js
```diff
@@ -2102,6 +2150,23 @@ router.post('/api/postcomment', ensureAuth, async function(req, res) {
...
     });
+
+    //give notification to owner
+    if (req.body.position.length == 0) {
+      //owner is blog post owner
+      const owner = await User.findById(post.owner);
+      await util.addNotification(owner, user, '/cube/blogpost/' + post._id, user.username + " added a comment to " + post.title);
+    } else {
+      //need to find owner from comment tree
+      const owner = await User.findById(getOwnerFromComment(post.comments, req.body.position));
```
purxiz: 2 args vs 3 args in the function

routes/cube_routes.js
```diff
@@ -2331,16 +2401,20 @@ router.post('/submitdeck/:id', async function(req, res) {
     }
 
     cube.numDecks += 1;
-    await cube.save();
-    const user = await User.findById(deck.owner);
+    const userq = User.findById(deck.owner);
```
purxiz: maybe a more descriptive variable name than userq?

routes/cube_routes.js
```diff
@@ -2331,16 +2401,20 @@ router.post('/submitdeck/:id', async function(req, res) {
...
+    const cubeOwnerq = User.findById(cube.owner);
```
purxiz: same for cubeOwnerq, more descriptive names would be better

serverjs/util.js
```diff
@@ -112,6 +112,31 @@ function fromEntries(entries) {
...
+    text: text
+  });
+  user.old_notifications.push({
+    user_from: from._id,
+    user_from_name: from.username,
+    url: url,
+    date: new Date(),
+    text: text
+  });
+  if (user.old_notifications.length > 200) {
```
purxiz: I would change this to a while loop just to avoid any weird conditions where the array ends up with more notifications than it should have

src/components/BlogPost.js
```diff
@@ -67,39 +77,45 @@ class BlogPost extends React.Component {
                     }
                 </div></h5>
                 <h6 className="card-subtitle mb-2 text-muted">
-                    <a href={"/user/view/" + post.owner}>{post.username}</a>
+                    <a href={"/user/view/" + post.owner}>{post.dev=='true' ? 'Dekkaru' : post.username}</a>
```
purxiz: We really need to implement user roles instead of throwing Dekkaru all around the code lol

dekkerglen: This is a bit different though, as this just codes in the case of what to display for names for devblog posts

purxiz: For now that's fair, I think such a system is outside the scope of this PR anyway

PR #679, opened 2019-11-13
Fix Failure to Show New Cube Modal

### LINKED ISSUE #678: Fail to Create New Cube for new account
**Describe the bug**
Clicking on the `Add a new cube?` link under `Your Cubes` fails to open the new cube modal.

**To Reproduce**
1. Go to the dashboard with an account that has not created any cubes.
2. Click on `Add a new cube?` under `Your Cubes`.
3. Nothing happens visibly and exceptions show up in the console.

**Expected behavior**
The modal for creating a new cube should open.

 - OS: MacOS
 - Browser: Chrome
 - Version: 78.0.3904.70

 - OS: MacOS
 - Browser: FIrefox
 - Version: 69.0.3

 - OS: MacOS
 - Browser: Safari
 - Version: 13.0.3

Should be fixed by creating the modal in the same we do from the menu here and I'll put a PR for that shortly. Might point to some script ordering issues as well.

ruler501:
Fixes #678 by using the same method to show the modal we use in the dropdown menu.

### DIFF
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -12,14 +12,6 @@ import { Card, Col, Row, CardHeader , CardBody, CardFooter} from 'reactstrap';
 class Dashboard extends Component {
   constructor(props) {
     super(props);
-    
-    this.newCube = this.newCube.bind(this);
-  }
-
-  newCube(evt)
-  {
-    evt.preventDefault();
-    document.getElementById('cubeModal').modal('show');
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


dekkerglen:
Was this tested? I think the issue is that 'cubeModal doesn't actually exist in the dashboard page, and needs to be copied over. Or better yet, made into a react component.

ruler501:
I tested it in the three browsers I mentioned that the fix worked and was a problem without the change. In Chrome I verified that the element did exist, but did not have a modal property when loaded with getElementById.

dekkerglen:
Right, because the modal is defined in the layout.pug Looks good!

Issue #537, opened 2019-09-29
Rarity Filter 'Starts with'


dekkerglen:
As a user, I would like to be able to use a filter as: 'r:r" to filter rarity by rare.

We should just use starts with when checking rarity.

purxiz:
No, we should use the simplifyArg map to alias 'r' to rare, and same for the other arguments.

ruler501:
I can take this up since it seems like a relatively simple change.

PR #680, opened 2019-11-13
Enable Single Char Rarity Args With Map

ruler501:
Add a map to use in simplifyArg to enable queries using just the first character of a rarity as the argument. An example would be `r:m` for finding mythics. A possible implementation for #537.

### DIFF
--- a/src/util/Filter.js
+++ b/src/util/Filter.js
@@ -215,9 +215,6 @@ function simplifyArg(arg, category) {
       res = parseManaCost(res)
       break;
     case 'rarity':
-      console.log(rarityMap);
-      console.log(res);
-      console.log(rarityMap.get(res));
       res = rarityMap.get(res) || res;
       break;
   }
--- a/src/util/Filter.js
+++ b/src/util/Filter.js
@@ -194,6 +194,13 @@ const colorMap = new Map([
   ['temur', 'rug']
 ]);
 
+const rarityMap = new Map([
+    ['c', 'common'],
+    ['u', 'uncommon'],
+    ['r', 'rare'],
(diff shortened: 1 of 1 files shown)


src/util/Filter.js
```diff
@@ -210,6 +217,14 @@ function simplifyArg(arg, category) {
     case 'mana':
       res = parseManaCost(arg)
       break;
+    case 'rarity':
+      let argLower = arg.toLowerCase()
+      if (rarityMap.has(argLower)) {
+        res = rarityMap.get(argLower);
+      } else {
+        res = arg;
+      }
```
dekkerglen: I think this entire case could be simplified as:
```suggestion
res = rarityMap[arg.toLowerCase()] ? rarityMap[arg.toLowerCase()] : arg;
```

ruler501: I was trying to match the style of the block above for colors and wasn't sure on ternary operators. Would we want to change that block to this style as well? Possibly introduce a utility function for `getOrDefault` as well?

dekkerglen: Yeah ternary operators are great if they are clean, like this example

ruler501: Updated with ternary and combined the mapping cases to make future additions easier.

dekkerglen: I don't like adding the extra map, if it doesn't eliminate the switch statement. Can we also not use .get() and .has() to maintain a consistent style?

ruler501: Indexing doesn't work correctly with Map objects, at least in my testing on Chrome so we do need to use get. I did make it a little shorter and removed the extra map in the next commit.

Issue #681, opened 2019-11-13
Allow Negative Parameters for Power


ruler501:
**Describe the bug**
Can't give a negative argument to a power query even though [[Char-Rumbler]] and [[Spinal Parasite]] have negative power(and for aguments there's [[Half-Squirrel, Half-]]).

**To Reproduce**
1. Go to Filters.
2. Input `power=-1`.
3. See that it fails validation.

**Expected behavior**
Query works and finds the above mentioned cards if they're in the domain of the search.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 78
This should happen the same in every browser though.

**Additional context**
Current we match against a regex to match a series of digits `[0-9]` only, adding `-?` to the beginning of the regex should solve this.

I can take this on.

PR #682, opened 2019-11-13
Allow half values for cmc as well as negative values for power/toughness

ruler501:
Fix for #681 plus the fractional part. Note [[little girl]] does not show up with `power:0.5`, but will show up with `power:.5` which is why the regexes got so complicated.

### DIFF
--- a/src/util/Filter.js
+++ b/src/util/Filter.js
@@ -250,9 +250,13 @@ const verifyTokens = (tokens) => {
             return false;
           }
           break;
-        case 'cmc':
         case 'power':
         case 'toughness':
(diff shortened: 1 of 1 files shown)


purxiz:
I don't think there's any need to allow +, even if javascript does generally support it. Also, I'm not sure why we'd allow anything other than .5/0.5, so maybe just:
`^(0?\.5)|(-?\d+)$` is a little easier?

ruler501:
Can remove the plus for sure, but [[Cardpecker]] has 1.5 power, should work to only allow non-negative to have a half after it. Cmc seems to just be [[Little Girl]] so `^((0?\.5)|(\d+))$` should work for that.

purxiz:
Didn't know about cardpecker. If it's both let's just leave it as you had it to avoid adding an extra case

PR #705, opened 2019-11-25
Adventure fix

RKelln:
WIP: do not merge, just for showing my work.


serverjs/updatecards.js
```diff
@@ -347,7 +347,8 @@ function convertColors(card, isExtra) {
   } else {
     if (!card.card_faces || card.layout == 'flip') {
       return colors.concat(card.colors);
-    } else if (card.layout == 'split') {
+    } else if (card.layout == 'split' || card.layout == 'adventure') {
+      if (card.name.startsWith('Flaxen')) console.log(card);
```
dekkerglen: remove debug line

serverjs/updatecards.js
```diff
@@ -596,6 +597,9 @@ function convertCard(card, isExtra) {
   newcard.scryfall_uri = card.scryfall_uri;
   newcard.rarity = card.rarity;
   newcard.oracle_text = card.oracle_text;
+  if (card.layout == 'adventure') {
+    newcard.oracle_text = card.card_faces[0].oracle_text + '\n' + card.card_faces[1].oracle_text;
+  }
```
dekkerglen: can we do this for all cards with multiple 'faces' objects?

src/util/Filter.js
```diff
@@ -446,35 +446,32 @@ function filterApply(card, filter, inCube) {
   if (filter.category == 'oracle' && card.details.oracle_text) {
     res = card.details.oracle_text.toLowerCase().indexOf(filter.arg) > -1;
   }
-  if (filter.category == 'color' && card.details.colors) {
+  if (filter.category == 'color' && colors !== undefined) {
```
dekkerglen: 'colors' is defined above as the card's color_identiy, or override. We should be using card.details.colors for the color filter

src/util/Filter.js
```diff
@@ -695,14 +692,44 @@ function filterApply(card, filter, inCube) {
...
+          if (symbol.includes('-')) {
+            res = false;
+          }
+        });
+        break;
+
       case 'hybrid':
-      case 'phyrexian':
-          let type = filter.arg.substring(0,1).toUpperCase() + filter.arg.substring(1);
-          res = cardIsLabel(card, type, 'Manacost Type');
```
dekkerglen: why are we not using cardIsLabel for this? Shouldn't we be consistent?

RKelln: Yes, would prefer to reuse - but I got stuck trying to figure out how to require the file. This code is just for testing until we figure out how to best share code between client and server.

serverjs/updatecards.js
RKelln: Yup, good idea (for searching purposes). Is the oracle text used for anything else that would muck up if we concatenate it?

dekkerglen: Nope, just used for filtering

src/util/Filter.js
dekkerglen: Isn't this already included? It shows you've replaced this code with something else

dekkerglen:
New PR for this fix has been created

Issue #524, opened 2019-09-27
Draft Deck CSS Looks off


dekkerglen:
The bot drafts CSS looks off, the font and padding should be the same as the table view.
![image](https://user-images.githubusercontent.com/28238025/65794601-7b9ed080-e136-11e9-9b4d-cc5c3fa5268b.png)

Issue #511, opened 2019-09-27
Deck Arrangement is based on Original Drafter's Page Resolution


haganbmj:
**Describe the bug**
When constructing a deck the card arrangement is based on the available window size, with cmc columns being combined and multi-row arrangements being used.
This sets the final display of the deck, regardless of what the viewer's resolution is.

Ex: https://cubecobra.com/cube/deck/5d8d79c3884bf5349176e316

**To Reproduce**
1. Draft a deck using a smaller screen resolution.
2. Complete the draft and Save the deck.
3. View the page using a larger screen resolution.

**Expected behavior**
Deck arrangement should be based on structured rules based on device resolution, or in some future state based on one of multiple custom display options.

**Screenshots**

Draft Resolution features groupings of creatures and spells.
Page Width: 851px
<img width="796" alt="cubecobra-decks-1" src="https://user-images.githubusercontent.com/15820761/65738980-c1fe1c00-e0b1-11e9-8ae1-e75d4ef94356.png">

Increased screen resolution shifts some columns to the previous row.
Page Width: 1261px
<img width="826" alt="cubecobra-decks-2" src="https://user-images.githubusercontent.com/15820761/65738983-c3c7df80-e0b1-11e9-96e9-c196da40cc23.png">

**Desktop (please complete the following information):**
Firefox Nightly 71.0a1 (2019-09-25) (64-bit)

dekkerglen:
Duplicate of #78, but you are much more descriptive so I'll close the other issue.

dekkerglen:
Quoting original issue:

> When a deck is saved with a low screen width, then viewed with a larger screen width, the bottom columns move upwards and everything gets mismatched.
> 
> Create a variable that saves the number of columns in the draft object, and when the draft is loaded into a screen that is too wide, take note of this variable and add empty columns such that column at position == num_cols will be the first element in the second row.

haganbmj:
Ah whoops, didn't pick the right search terms I guess.
Thanks for linking the two.

Issue #453, opened 2019-09-19
Draft Auto-order uses details CMC


dekkerglen:
When drafting, the default CMC column for a card is based off the card's details' value of CMC, but should use the override value instead.

Issue #405, opened 2019-09-15
Double face cards not showing up on hover in draft


lane-eden:
Only the front side of cards show up when you hover over them in draft mode.

PR #711, opened 2019-11-27
Pull draft, deckbuilder, and deck view into React.

### LINKED ISSUE #398: Drag and Drop not working on touch devices
On mobile, in the draft/deckbuilder ui, drag and dropping cards does not work properly.

### LINKED ISSUE #733: DFCs backs in draft not visible
But I'm still not seeing the back half of DFCs when I'm doing playtest drafts. I'm on mobile but I tried it in mobile mode and desktop mode.

phulin:
Title of the PR says most of what it does, but a few things for other devs to note:

* The `display_image` field has now been eliminated. Use `util.getCardImageURL` on the server-side and the `CardImage` or `FoilCardImage` components on the client-side.
* There are a bunch of new React components for people to use.
* I've started adding `propTypes` to do type-checking on new React components. This is a best practice IMO and I'll be doing it with all my code going forward; I recommend others do the same as well.
* I'm now just typing scripts into the pug templates instead of making a bunch of random hidden inputs. I think this makes more sense.

With this PR, almost all of the dynamic frontend is in React. The next step is going to be to get all the miscellaneous static pages in, and then we'll be able to have the whole layout in one engine. After that my hope would be that we could start doing server-rendering of the React components + client-side hydration.

When you pull this, you'll have to run `npm install` as some package versions have changed.

Fixes #733, #524, #511, #453, #405. Probably fixes #398 and some other issues too.


phulin:
Working on this again now

PR #712, opened 2019-11-27
Import visible tags from CubeTutor.

### LINKED ISSUE #710: Import tags and colors from CubeTutor

phulin:
Fixes #710.


dekkerglen:
Note: this only works for visible tags, it will not work for tags that have been assigned a color on CT

phulin:
True

dekkerglen:
This PR is currently failing the prettier check

phulin:
I think that’s all in files i didn’t modify but i can merge in master if you want

PR #717, opened 2019-11-30
Card Pages with analytics

dekkerglen:
This PR is ready for review.

This adds the route '/tool/card/:id', where id can be the card name or the card id. If a cardname is provided, it is rerouted to the default 'reasonable' version of that card.

Card pages included counts for what kinds of cubes run this card, overall inclusion rate, cards found in decks with this card, and some cubes that run this card. The cubes are 12 random cubes taken from all cubes that include the card.

This PR includes a populate_analytics.js script, which is an expensive script to run and will likely need the flag: --max-old-space-size=8192 to avoid OOM.

I'll setup a chron job on my local machine to run this script regularly on the live database to keep these pages up to date.


populate_analytics.js
```diff
@@ -0,0 +1,283 @@
...
+}
+
+async function processDeck(deck) {
+    if(deck.playerdeck && deck.playerdeck.length > 0) {
+        //flatten array
+        const deckCards = [];
+        deck.playerdeck.forEach(function(col, index) {
+            if(Array.isArray(col)) {
+                col.forEach(function(row, index2) {
+                    deckCards.push(row.details.name.toLowerCase().normalize('NFD')
```
RKelln: Is there a reason not to use `carddb.normalizedName(row.details)` here? (And below with `col.details`?)

populate_analytics.js
```diff
@@ -0,0 +1,283 @@
...
+            }
+        }
+    }
+}
+
+async function processCube(cube) {
+
+    let cubeSizeDict = cardSizeUses.size180;
+    let cubeLegalityDict = cardSizeUses.vintage;
+
```
RKelln: I wonder if there should be some sort of looser/fuzzier boundary here? Are people sticklers for sizes or is close enough better for categorization? If looser, just by  a absolute amount? 10? Or percentage?
```js
let fuzzyBoundary = 1.05; // allow for sizes within 5% of card count
if (cube.card_count <= Math.floor(180 * fuzzyBoundary)) {
```

dekkerglen: the actual ranges are:
image

src/components/CountTableRow.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const CountTableRow = ({label, value, ...props }) => (
+    <tr>
+        <td>{label}:</td>
+        <td>
+            {Math.round(value[1] * 1000.0)/10}%                              
+            <span className="percent">{value[0]}</span>
```
RKelln: Probably a good idea to change this class name to "additional_detail", "number_detail" or something similar and update the "percent" analytics class name too.

src/cardpage.js
```diff
@@ -0,0 +1,151 @@
...
+                    ) : (
+                      <div className="card-price">Foil TCGPlayer Market: {prices[card.tcgplayer_id + '_foil'].toFixed(2)}</div>
+                    )}
+                  </div>
+                </Col>
+                <Col className='breakdown' xs="12" sm="8">
+                  <p>
+                    {'Played in '}                     
+                    {Math.round(data.total[1] * 1000.0)/10}%
+                    <span className="percent">{data.total[0]}</span>
```
RKelln: Potential class name change (see https://github.com/dekkerglen/CubeCobra/pull/717/files#r353357469)

routes/tools_routes.js
```diff
@@ -109,6 +113,30 @@ function topCards(filter, res) {
   });
 }
 
+function shuffle(a) {
```
RKelln: There is a `shuffle()` in `serverjs/utils.js`, reuse?

PR #722, opened 2019-12-03
Default versions now more reasonable

dekkerglen:


serverjs/cards.js
```diff
@@ -115,6 +115,24 @@ function unloadCardDb() {
...
+  return notPromoOrDigitalCard(cardFromId(id));
+}
+
+function notPromoOrDigitalCard(card) {
+  return !card.promo && !card.digital && card.border_color != 'gold' && card.language == 'en' && card.tcgplayer_id;
+}
+
+function getMostReasonable(cardname) {
+  const cards = data.nameToId[cardname];
+  for (let i = 0; i < cards.length; i++) {
```
phulin:
```suggestion
  for (const card of cards) {
```

serverjs/cards.js
```diff
@@ -115,6 +115,24 @@ function unloadCardDb() {
...
+    if (notPromoOrDigitalId(cards[i])) {
```
phulin:
```suggestion
    if (notPromoOrDigitalId(card)) {
```

serverjs/cards.js
```diff
@@ -115,6 +115,24 @@ function unloadCardDb() {
...
+      return cardFromId(cards[i]);
```
phulin:
```suggestion
      return cardFromId(card);
```

serverjs/updatecards.js
```diff
@@ -687,7 +689,15 @@ function convertCard(card, isExtra) {
   newcard.color_identity = newcard.color_identity.concat(card.color_identity);
   newcard.set = card.set;
   newcard.collector_number = card.collector_number;
-  newcard.promo = card.promo;
+  newcard.promo =
```
RKelln: I would put the promo detection in its own function to match the style of the `convertThing()` attributes and keep `convertCard()` as small as possible.

dekkerglen: I think I value not having these fields in the card object, to reduce the size of the actual object a lot more.

PR #731, opened 2019-12-05
Foils

anderoonies:
this is branched off of #730 

Adds the `finish = "Non-foil" | "Foil"` type to cards, the management tools to set, filter, and sort on `finish`, and a .png overlay for `finish=Foil` cards.
image

### Setting `finish` in CardModal
Confirm you can set/unset the `finish` in the CardModal
### Setting `finish` in ListView
Confirm you can set/unset the `finish` in ListView
### Autocard overlay
Confirm that Autocards have the png overlay 
### Sorting
Confirm you can sort on Finish
### Filtering
Confirm you can filter on Finish
### DFCs
Foil out some DFCs


serverjs/prices.js
```diff
@@ -79,7 +83,12 @@ async function GetPricesPromise(card_ids) {
     chunks.push(card_ids.slice(i * chunkSize, (i + 1) * chunkSize));
   }
 
-  const access_token = await GetToken();
+  let access_token;
```
anderoonies: this is from #730

anderoonies: everything in this file is

PR #811, opened 2020-01-07
modifying pagination to work with lots of blog pages

### LINKED ISSUE #600: Pagination doesn't line wrap
See: https://www.cubecobra.com/cube/blog/19h

joshfinnie:
This updates the pagination component to handle a lot of pages...

Closes #600 

image

### DIFF
--- a/src/components/Paginate.js
+++ b/src/components/Paginate.js
@@ -1,29 +1,112 @@
+/* eslint-disable react/jsx-filename-extension */
 import React from 'react';
 
 import { Pagination, PaginationItem, PaginationLink } from 'reactstrap';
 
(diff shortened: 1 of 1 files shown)


src/components/Paginate.js
```diff
@@ -1,29 +1,112 @@
+/* eslint-disable react/jsx-filename-extension */
```
phulin: I was assuming that we'd either move all the files or turn this on repo-wide once we get there.

joshfinnie: @phulin yea, I think I just have a must more strict lint notifier... I can remove this if it's seen as noise.

PR #817, opened 2020-01-08
Feature 792

### LINKED ISSUE #792: Card Links in Deck Page
As a user, I would like to be able to click on cards in a user's deck, and in the bot 'decks' to view that card's page

dekkerglen:
fixes #792
Also fixed a bug where anonymous decks can't be rebuilt, and removed the number above the column during the draft when the column count is not > 0

### DIFF
--- a/src/components/ImageFallback.js
+++ b/src/components/ImageFallback.js
@@ -13,17 +13,7 @@ const ImageFallback = ({ src, fallbackSrc, innerRef, ...props }) => {
     setFallback(false);
   }, [src]);
 
-  return (
-    <>
-      {props.cardid ? (
-        <a href={'/tool/card/' + props.cardid}>
-          <img src={fallback ? fallbackSrc : src} onError={handleError} ref={innerRef} {...props} />
-        </a>
-      ) : (
-        <img src={fallback ? fallbackSrc : src} onError={handleError} ref={innerRef} {...props} />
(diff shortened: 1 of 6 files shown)


src/components/ImageFallback.js
```diff
@@ -13,7 +13,17 @@ const ImageFallback = ({ src, fallbackSrc, innerRef, ...props }) => {
...
+    <>
+      {props.cardid ? (
+        <a href={'/tool/card/' + props.cardid}>
+          <img src={fallback ? fallbackSrc : src} onError={handleError} ref={innerRef} {...props} />
+        </a>
+      ) : (
+        <img src={fallback ? fallbackSrc : src} onError={handleError} ref={innerRef} {...props} />
+      )}
+    </>
+  );
```
phulin: This is supposed to be a generic component that is not necessarily specialized to card images. Can you add this to the `CardImage` component instead? Sorry to be a pain but i want to keep the basic components clean if possible

src/components/WithAutocard.js
```diff
@@ -28,6 +28,7 @@ const withAutocard = (Tag) =>
     back = back || card.details.image_flip;
     return (
       <Tag
+        cardid={card.cardID}
```
phulin: Can get rid of this one too

PR #820, opened 2020-01-08
Add card filters to custom drafts.

### LINKED ISSUE #433: Custom draft format with filter syntax
As a user, I would like to be able to use filter syntax to create my custom draft format. There should be a toggle to turn on the filter syntax, so the existing formats don't break.

RKelln:
Closes #433 

Adds card filter support for custom drafts but retain backwards compatibility (i.e. anything without filter syntax is assumed to be a tag and * still works for any card).

Tests still to come.

Has a few other fixes that were needed for this to work with the custom draft I wanted:
- fix tests (Pioneer support)
- add color and color identity count filtering: e.g. c>1 == cards with more than 1 color


routes/cube_routes.js
```diff
@@ -1725,19 +1725,43 @@ router.post('/startdraft/:id', function(req, res) {
...
+      seats: parseInt(req.body.seats),
+      packs: parseInt(req.body.packs),
+      cards: parseInt(req.body.cards),
+    };
+
+    // setup draft cards
+    let draftcards = cube.cards.slice();
+    if (draftcards.length === 0) {
+      req.flash('danger', 'No cards in cube. You must add cards before drafting.');
+      res.redirect('/cube/playtest/' + req.params.id);
```
dekkerglen: return res.redirect to prevent multiple results

routes/cube_routes.js
```diff
@@ -1725,19 +1725,43 @@ router.post('/startdraft/:id', function(req, res) {
...
+    let format = draftutil.getDraftFormat(params, cube);
+
+    try {
+      let draft = draftutil.createDraft(format, draftcards, bots, params.seats);
+      draft.cube = cube._id;
+      draft.save(function(err) {
+        if (err) {
+          throw err;
+        } else {
+          res.redirect('/cube/draft/' + draft._id);
```
dekkerglen: return res.redirect

routes/cube_routes.js
```diff
@@ -1725,19 +1725,43 @@ router.post('/startdraft/:id', function(req, res) {
...
+      card.details = carddb.cardFromId(card.cardID);
+    });
+    // set up bots and format
+    let bots = draftutil.getDraftBots(params);
+    let format = draftutil.getDraftFormat(params, cube);
+
+    try {
+      let draft = draftutil.createDraft(format, draftcards, bots, params.seats);
+      draft.cube = cube._id;
+      draft.save(function(err) {
```
dekkerglen: please don't use callback, let's use async/await. It is much cleaner. This line can be revised to:
`await draft.save();`
And make sure that the entire route is in a try/catch block to catch any errors. Then you don't need to have as much nested code

update_fixtures.js
```diff
@@ -15,6 +15,7 @@ const success = 0;
 
 // rebuild the fixtures based on the previous subset of cards
 (async function() {
+  console.log('Attempting to read', cardsPath);
```
dekkerglen: please remove this debug line

routes/cube_routes.js
RKelln: Sure, I was just copying style from nearby code to maintain style. Definitely prefer less nesting.

PR #821, opened 2020-01-08
Maybeboard feature.

### LINKED ISSUE #246: A maybeboard to hold potential cube inclusions
Have a dedicated area to put potential inclusions for the cube so that I don't forget about it. This area would not count towards any of the cube's statistics, but would be displayed on the cube listing page. Controls to easily move cards to and from this area would also be provided.

Currently users can use blog posts and the overview to track maybeboard cards. So this is not a pressing feature.

phulin:
This PR adds a "maybeboard" where you can store cards you are considering adding to a cube. The maybeboard is hidden by default and is exposed by an option in the "Display" menu. Clicking on a card in the maybeboard adds it to a changelist so you can add it to the cube.

Screenshot:
image

Closes #246.

### DIFF
--- a/models/cube.js
+++ b/models/cube.js
@@ -63,8 +63,14 @@ const cubeSchema = mongoose.Schema({
     type: [String],
     default: [],
   },
-  cards: [Card],
-  maybe: [Card],
+  cards: {
+    type: [Card],
+    default: [],
+  },
+  maybe: {
+    type: [Card],
+    default: [],
+  },
   tag_colors: [
     {
       tag: String,
--- a/models/cube.js
+++ b/models/cube.js
@@ -1,5 +1,18 @@
 const mongoose = require('mongoose');
 
+const Card = {
+  tags: [String],
+  finish: { type: String, default: 'Non-foil' },
+  status: String,
+  colors: [String],
+  cmc: Number,
+  cardID: String,
+  type_line: String,
+  addedTmsp: Date,
+  imgUrl: String,
+  details: {},
+};
+
 // Cube schema
 const cubeSchema = mongoose.Schema({
   name: {
@@ -50,20 +63,8 @@ const cubeSchema = mongoose.Schema({
     type: [String],
     default: [],
   },
-  cards: [
-    {
-      tags: [String],
-      finish: { type: String, default: 'Non-foil' },
-      status: String,
-      colors: [String],
-      cmc: Number,
-      cardID: String,
-      type_line: String,
-      addedTmsp: Date,
-      imgUrl: String,
-      details: {},
-    },
-  ],
+  cards: [Card],
+  maybe: [Card],
   tag_colors: [
     {
       tag: String,
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -796,6 +796,7 @@ router.get('/list/:id', async function(req, res) {
       defaultShowTagColors: !req.user || !req.user.hide_tag_colors,
       defaultSorts: cube.default_sorts,
       cards,
+      maybe: maybeCards(cube),
     };
 
     res.render('cube/cube_list', {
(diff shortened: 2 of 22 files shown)


routes/cube_routes.js
```diff
@@ -3160,6 +3165,65 @@ router.post('/api/updatecards/:id', ensureAuth, function(req, res) {
   });
 });
 
+function maybeCards(cube) {
+  const maybe = (cube.maybe || []).filter((card) => card.cardID);
```
anderoonies: are there negative or 0 cardIDs you're filtering out? do those just exist in the cube?

models/cube.js
```diff
@@ -50,20 +63,8 @@ const cubeSchema = mongoose.Schema({
...
-      cmc: Number,
-      cardID: String,
-      type_line: String,
-      addedTmsp: Date,
-      imgUrl: String,
-      details: {},
-    },
-  ],
+  cards: [Card],
+  maybe: [Card],
```
dekkerglen: I'd like to include default values for these

routes/cube_routes.js
```diff
@@ -3160,6 +3165,65 @@ router.post('/api/updatecards/:id', ensureAuth, function(req, res) {
...
+        success: 'false',
+        message: 'Cube not found.',
+      });
+    }
+
+    return res.status(200).send({
+      success: 'true',
+      maybe: maybeCards(cube),
+    });
+  }),
```
dekkerglen: can we wrap this in a try/catch?

routes/cube_routes.js
```diff
@@ -3160,6 +3165,65 @@ router.post('/api/updatecards/:id', ensureAuth, function(req, res) {
...
+    const withAdded = [...withRemoved, ...addCardsNoDetails];
+
+    cube.maybe = withAdded;
+    await cube.save();
+
+    return res.status(200).send({
+      success: 'true',
+    });
+  }),
+);
```
dekkerglen: I would also like to wrap this in a try/catch

routes/cube_routes.js
```diff
@@ -3160,6 +3165,65 @@ router.post('/api/updatecards/:id', ensureAuth, function(req, res) {
...
+  '/api/maybe/:id',
+  ensureAuth,
+  util.wrapAsyncApi(async function(req, res) {
+    const cube = await Cube.findOne(build_id_query(req.params.id));
+    if (!cube) {
+      return res.status(404).send({
+        success: 'false',
+        message: 'Cube not found.',
+      });
+    } else if (!req.user._id.equals(cube.owner)) {
```
dekkerglen: else not needed, as we return in the previous if block

routes/cube_routes.js
dekkerglen: nvm - I just saw the wrapper method

routes/cube_routes.js
dekkerglen: nvm - I just saw the wrapper method

src/components/EditCollapse.js
```diff
@@ -11,6 +11,24 @@ import ChangelistContext from './ChangelistContext';
...
+  if (name && name.length > 0) {
+    const normalized = encodeName(name);
+    const response = await fetch(`/cube/api/getcard/${normalized}`);
+    if (!response.ok) {
+      setAlerts((alerts) => [...alerts, { color: 'danger', message: `Couldn't get card: ${response.status}.` }]);
+      return null;
+    }
+
+    const json = await response.json();
+    if (json.success !== 'true' || !json.card) {
```
dekkerglen: In what situation will only one of these conditionals be true? Would it make sense to just use one if statement here?

routes/cube_routes.js
phulin: No, just trying to avoid server crashes if something gets corrupted.

src/components/EditCollapse.js
phulin: I don't think there is one, just trying to be robust to weird failures + bugs

routes/cube_routes.js
phulin: just habit and it lets me put it on the same line

models/cube.js
phulin: done

PR #831, opened 2020-01-10
Improve performance of draft view.

phulin:
This PR transmits only some of the card fields in draft view in order to improve performance and also uses server-side rendering for the draft page and playtest page.

I highly recommend other devs start using the SSR code (example here) for any new pages.


dekkerglen:
Conflicts

PR #833, opened 2020-01-11
Bring playtest page modals into react.

phulin:
Title self-explanatory. This is the last of the hard reactification stuff for the cube pages - next step is to move all of the cube layout into React, including the tab bar.


dekkerglen:
Conflicts

PR #835, opened 2020-01-11
Fix Microsoft Edge support.

phulin:
We have to transpile several of our node_modules in order to work on edge, as they use ES6 features that Edge doesn't support.

Important: This PR removes support for IE 11 as it significantly reduces our bundle sizes.


dekkerglen:
conflicts

PR #843, opened 2020-01-13
Move bulk upload page into React.

phulin:


routes/cube_routes.js
```diff
@@ -1230,40 +1237,40 @@ router.post('/uploaddecklist/:id', ensureAuth, async function(req, res) {
   }
 });
 
-router.post('/bulkupload/:id', ensureAuth, function(req, res) {
-  Cube.findOne(build_id_query(req.params.id), function(err, cube) {
+router.post('/bulkupload/:id', ensureAuth, async function(req, res) {
```
dekkerglen: it's great that these are async now, but I would really like to remove all the callbacks and make the body all use async when we are updating these methods

PR #847, opened 2020-01-13
Move draftutils into the client code

ruler501:
Move Draftutils to be usable from the React components. Needed for the Webworker-Analytics changes so we can calculate asfan on the client side.

Rebased on #846 assuming that will get merged.


src/util/Filter.js
```diff
@@ -42,7 +42,7 @@ let categoryMap = new Map([
 ]);
 
 const operators = ['>=', '<=', '<', '>', ':', '!=', '='];
-const operatorsRegex = new RegExp('(?:' + operators.join('|') + ')');
+export const operatorsRegex = new RegExp('(?:' + operators.join('|') + ')');
```
dekkerglen: we should only be using export once at the end of the file, if these functions should be accessible outside then add them to the object export at the end of the file

phulin: I've used this pattern elsewhere - no problem in ES6 modules with mixing named + default exports

phulin: I should also note that without specific exports like this one you _can't_ use the `import { operatorsRegex } from 'util/Filter'` pattern. The ES6 imports we use in the client code work differently from the commonjs ones in the server code.

PR #851, opened 2020-01-14
Move all the cube pages fully into React.

phulin:
This PR creates a new `CubeLayout` component that manages the navbar on the cube pages. Among other things, the change enables having the cube category override logic in the cube navbar. It also enables using non-relative paths in imports in the React components (`components/X.js` instead of `../components/X.js`).


dekkerglen:
Conflict

phulin:
Fixed again

PR #862, opened 2020-01-15
Cube Overview Bugfixes

### LINKED ISSUE #854: Capitals in custom id cause links to break
We need to use a .tolowercase() on the edit overview submit route in order to accommodate this

### LINKED ISSUE #861: Overview Modal doesn't load all fields
Overview Modal does not load the description correctly if the description is not using html.
Overview Modal also does not load the custom url alias

dekkerglen:
Fixes #861, Fixes #854

### DIFF
--- a/src/components/CubeOverviewModal.js
+++ b/src/components/CubeOverviewModal.js
@@ -26,13 +26,10 @@ class CubeOverviewModal extends Component {
   constructor(props) {
     super(props);
 
-    props.cube.descriptionhtml= props.cube.descriptionhtml && props.cube.descriptionhtml !== 'undefined' ? props.cube.descriptionhtml : props.cube.description;
-
     this.state = {
       isOpen: false,
       tags: props.cube.tags.map((tag) => ({ id: tag, text: tag })),
       cube: JSON.parse(JSON.stringify(props.cube)),
-      description: props.cube.descriptionhtml,
       image_dict: {},
     };
 
@@ -224,7 +221,6 @@ class CubeOverviewModal extends Component {
 
   render() {
     const { cube, cubeID, tags, isOpen } = this.state;
-    console.log(cube);
     return (
       <>
         <a className="nav-link" href="#" onClick={this.open}>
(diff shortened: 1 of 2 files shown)


src/components/CubeOverviewModal.js
```diff
@@ -26,14 +26,17 @@ class CubeOverviewModal extends Component {
   constructor(props) {
     super(props);
 
+    props.cube.descriptionhtml= props.cube.descriptionhtml && props.cube.descriptionhtml !== 'undefined' ? props.cube.descriptionhtml : props.cube.description;
```
phulin: This totally breaks React's data model and you should not do this (meaning modifying incoming props).

src/components/CubeOverviewModal.js
```diff
@@ -368,7 +368,7 @@ class CubeOverviewModal extends Component {
                 <h6>Description</h6>
                 <TextEntry
                   name="blog"
-                  value={this.state.cube.descriptionhtml ? this.state.cube.descriptionhtml : ''}
+                  value={cube.descriptionhtml && cube.descriptionhtml !== 'undefined' ? cube.descriptionhtml :cube.description}
```
phulin:
```suggestion
                  value={typeof cube.descriptionhtml !== 'undefined' ? cube.descriptionhtml :cube.description}
```

dekkerglen: What is this hack

phulin: This is actually how you are supposed to check if something is undefined

dekkerglen: The issue is descriptionhtml can be set to the string value of 'undefined' if any other field was updated with the new modal

dekkerglen: So this checks if it's actually undefined, or if it is equal to the string value of undefined

phulin:
Oh wat i don’t like that pattern but this pr is fine with me then

dekkerglen:
I need to beautify it, will do that once I get home

Issue #850, opened 2020-01-14
New Analytic: Prices


dekkerglen:
As a user I would like an analytics section that goes into detail about the price totals for my cube, taking into account card status and finish. I'd like to see several values, like $ amount to finish cube, $ amount to finish foiling cube, $ amount of currently owned cards, etc

ruler501:
Note for anyone starting this there's on progress work to greatly rework the analytics system. PR #875 is currently tracking that. The system is not compatible with the old one so when writing new analytics would likely want to base it off of that code, or wait for it to get reviewed and merged.

dekkerglen:
added

Issue #19, opened 2019-07-29
Comprehensive Analysis


dekkerglen:
As a user, I would like more options to slice my cube list's data in the analysis tab.

Here is an example of the kinds of data that would be useful: https://docs.google.com/spreadsheets/d/1s1BD4797XT--pDm_ZfJiDsnxIz_0DDx1b7rsbHHD1fk/edit

pozz94:
I would really like to see the average CMC of the cube and the average CMC of each color.
It would also be nice to have it visualized together with the mana curve and a pie chart of the type breakdown on the same page without having to switch between different sections of the analysis data

RKelln:
As a user I would really like to know the percentage of the different types of cards regardless of the total number of cards in each color.

Add percentage to the totals on the Type Breakdown stats tab. (Preferably leaving 0 and perhaps 1 card values without any percentage to reduce clutter). Round to 1 decimal place?

ruler501:
I can take this up. Would probably do best to do a series of PRs that only add a few things at a time since this should be able to be broken up quite a bit.

One possible architecture that would allow users to add their own functions would be a web worker system. That should isolate the user provided scripts and prevent adding extra load to the server, though I don't fully know all the security implications and would appreciate feedback. Eventually could have sharable functions, kind of like roll20 does with scripts. For a first iteration giving web workers a few query callbacks/messages, potentially just a callback/message that takes a string query like the web ui and gives back to the script a json list of matching cards and one to submit the final display string or html.

dekkerglen:
It's probably better just to use our tags + filters that we already have built to be able to slice the data onto different types of charts/tables in a flexible way.

I would also recommend against this as a first issue, maybe try one with the 'good first issue' tag to get started and learn the codebase a bit before taking on such a huge undertaking

ruler501:
I put together a rough proof of concept of one way to do user customizable analytics. Can test it out at https://cubecobra.ruler501.com/cube/analysis/packwars and the code is https://github.com/ruler501/CubeCobra/tree/webworker-analytics only supports tables at the moment, but graphs should be relatively easy to add next if this is a path that's deemed promising.

roborocket:
[...]

PR #875, opened 2020-01-15
Client-Side Analytics

ruler501:
Effectively the same as #736 to give somewhere to put discussion and work items to get it put back into master.
First steps to #19 and #850 

You can test this out at https://cubecobra.ruler501.com/cube/analysis/packwars.
The UI could use a decent amount of work. I'm relatively poor at designing and writing UI so have been focused mostly on getting the functionality working.

The overarching idea is to have analytics be run on the client side over the set of cards in the cube through WebWorkers which provide isolation from the page environment. These scripts would receive messages with the cards and respond with descriptions of the visualization they want for their analytic. I planned to allow responses to render tables, line graphs, bar graphs, pie charts, word clouds and card image with associated text grids(like the token analytic now). 
This also introduces a markdown lke language that supports rendering percentages, mana symbols, card images, and card links with autocard.

Further work intended is to add the other visualization supports mentioned, port over all existing analytic functions, add tests, and make the script requirements and how to write more discoverable. 

Custom analytics has been dropped from scope for now due to security concerns.

## Changes and Features:
- [x] Calculate analytics on the client side in isolated Webworker contexts. 
- [x] Allow running Filters on the cards before passing to the analytics. 
- [x] Allow custom draft formats for calculating Asfan the analytics can use.
- [x] Render tabular data returned from analytics including column headers(missing row headers).
- [x] Render arbitrary chart types returned by the analytics.
- [x] Render word cloud data returned by analytics.
- [x] Render card image with caption data returned by analytics. 
- [x] Add analytic script for calculating Asfan and count by color identity set inclusion.
- [x] Port analytic script for calculating Asfan and count by color identity set equality (UI change needs review). 
- [x] Port analytic script for calculating the type grid.
- [x] Port analytic script for calculating the cmc graph.
- [x] Port analytic script for calculating needed tokens and cards that need them.
- [x] Add analytic script for drawing a word cloud based on tags used in the cube.
- [x] Document expected inputs and outputs of analytic scripts. 
- [ ] Finalize UI and UX for the analytics page.
- [ ] Fix UX for the included analytics.
- [x] Separate Type Breakdown and Curve into two separate analytics to separated "Expected Count in Pool" and "Count."
- [x] Add descriptions to all analytic scripts.
- [x] Add visualization for Average CMC by color combination (Probably table, maybe bar chart).
- [x] Copy List navbar design for Filter/Draft Format input.
- [x] Enable multiple tables to be displayed together so Color Counts can be viewed without scrolling much/at all.

## Potential Future Work:
- Allow adding custom analytic scripts
- Persist custom analytic scripts.
- Allow editing, deleting, and duplicating analytic scripts.
- Allow Analytics scripts to have parameters for customizing behavior without editing code.
- Allow Analytic scripts to take other cubes as parameters.
- Allow comparing analytics before and after a suggested change from the Maybeboard.
- Display multiple analytics side by side. Needs a lot of UI design questions answered, but say having a graph next to a table of the data can be useful.
- Add other category for Type Breakdown(Vanguards, conspiracies, etc).
- Add a pricing analytic, documented in #850.


src/components/FilterCollapse.js
```diff
@@ -338,6 +334,7 @@ class FilterCollapse extends Component {
     if (!valid) return;
 
     if (tokens.length > 0) {
+      Query.set('f', this.state.filterInput);
```
phulin: Why this change?

public/js/analytics/averageCmc.js
```diff
@@ -0,0 +1,76 @@
+function GetColorCat(colors) {
```
phulin: Can you find a way to get all this code inside of webpack? That would involve putting it in the `src` directory instead and adding entry points in the `webpack.common.js` file for each analytic script. That way we would be able to avoid all the copying-and-pasting present throughout this PR. We'd also get minification and the ability to use node packages, all of which are good. I've been trying to eliminate all the code in the `public/js` dir and this goes in the opposite direction. Feel free to tag me on discord if you have questions. It would also make sure that this code is subject to eslint.

public/js/analytics/averageCmc.js
```diff
@@ -0,0 +1,76 @@
...
+    Green: { label: '{g}', asfan: 0, count: 0, totalAsfan: 0, totalCount: 0 },
+    Colorless: { label: '{c}', asfan: 0, count: 0, totalAsfan: 0, totalCount: 0 },
+    Multi: { label: '{m}', asfan: 0, count: 0, totalAsfan: 0, totalCount: 0 },
+    Total: { label: 'Total', asfan: 0, count: 0, totalAsfan: 0, totalCount: 0 },
+  };
+
+  for (const card of cards) {
+    const asfan = card.asfan || 15 / cards.length;
+    const colorCat = GetColorCat(card.colors || card.details.color_identity);
+    const cmc = card.cmc || card.details.cmc;
```
phulin: This won't work if the CMC is overridden to zero. See suggestion.
```suggestion
    const cmc = card.cmc !== undefined ? card.cmc : card.details.cmc;
```

public/js/analytics/colorCount.js
```diff
@@ -0,0 +1,113 @@
+function areArraysEqualSets(a1, a2) {
```
phulin: This function can be implemented more concisely as follows:
```javascript
const set1 = new Set(a1);
const set2 = new Set(a2);
return a1.every((x) => set2.has(x)) && a2.every((x) => set1.has(x));
```
If this code is all in webpack land, that function could be in a utility file as well.

public/js/analytics/colorCurve.js
```diff
@@ -0,0 +1,141 @@
...
+      case 'c':
+        category = curve.colorless;
+        break;
+      default:
+        break;
+    }
+    // const asfan = card.asfan || 15 / cards.length;
+    const asfan = 1;
+    if (category) {
+      if (card.cmc >= 9) {
```
phulin:
```suggestion
      if ((card.cmc !== undefined ? card.cmc : card.details.cmc) >= 9) {
```
Or you could make this a utility function and replace throughout this file

public/js/analytics/typeBreakdown.js
```diff
@@ -0,0 +1,202 @@
...
+        return 'Colorless';
+    }
+  }
+}
+
+onmessage = (e) => {
+  if (!e) return;
+  const cards = e.data;
+
+  const TypeByColor = {
```
phulin:
```suggestion
  const TypeByColor = Object.fromEntries(['Creatures', /* ... */ ]).map((category) => ({ /* ... */ })))
```

public/js/analytics/typeBreakdownCount.js
```diff
@@ -0,0 +1,202 @@
...
+        return 'Colorless';
+    }
+  }
+}
+
+onmessage = (e) => {
+  if (!e) return;
+  const cards = e.data;
+
+  const TypeByColor = {
```
phulin: Same as above. I'd like to be mindful of code size here if possible

routes/cube_routes.js
```diff
@@ -883,14 +881,58 @@ router.get('/analysis/:id', async (req, res) => {
...
+            type_line: tokenDetails.type,
+            addedTmsp: new Date(),
+            imgUrl: undefined,
+            finish: 'Non-foil',
+            details: { ...tokenDetails },
+          };
+        }
+      }
+    }
+    const priceDict = await GetPrices(pids);
```
phulin: You can use the `addPrices` function in `prices.js` to do exactly this

src/components/AnalyticsTable.js
```diff
@@ -0,0 +1,93 @@
...
+//     },
+//   ],
+// }
+const HeaderCell = ({ children, ...props }) => (
+  <th scope="row" {...props}>
+    {children}
+  </th>
+);
+const RegularCell = ({ children, ...props }) => <td {...props}>{children}</td>;
+RegularCell.propTypes = { children: PropTypes.shape({}).isRequired };
```
phulin:
```suggestion
RegularCell.propTypes = { children: PropTypes.node.isRequired };
```
Children should always be proptypes.node afaik

RKelln:
Regarding the analytics js files, I think we want those to look extremely minimal and "business logic" only.  Basically it should read as some math and how to display/format the results. There should be easy to use helpers that handle the thing that stats and analytics require.

Candidates include colorCombinations, etc but also typical data structures:
```js
analytics.getColorsArray({asfan: 0, count: 0, totalAsfan: 0, totalCount: 0 }) 
```
Could return an associative array with the objects initialized to the passed in param, and already comes with some standard properties like label: '{w}', etc.

public/js/analytics/typeBreakdown.js
ruler501: I was trying to think of a way to compress this down, thank you.

src/components/FilterCollapse.js
ruler501: Before we were updating the query as the user typed. I changed it so the query string is only ever what the user should be seeing at that moment.

ruler501:
@RKelln that was my goal with these, if you have some ideas on how to implement it that way I'd love to hear them. I'm not entirely clear on what we can pull into these scripts as far as library functions go (assuming they stay where they are) just because of how WebWorkers function.

RKelln:
Ah OK, wasn't aware of web worker restrictions. I don't know workers but a google suggests you can do: ` importScripts('foo.js'); ` So does that mean you could make resuable code in an `analytics.js` helper?

phulin:
Webpack will manage the imports for you - the webpack output will just
inline the necessary code.

src/components/CubeAnalysisNavbar.js
```diff
@@ -0,0 +1,118 @@
...
+const CubeAnalysisNavBar = ({
+  draftFormats,
+  formatId,
+  setFormatId,
+  filter,
+  setFilter,
+  numCards,
+  defaultFilterText,
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
```
phulin:
```suggestion
  const [isOpen, setIsOpen, toggleIsOpen] = useToggle(false);
```

src/components/CubeAnalysisNavbar.js
```diff
@@ -0,0 +1,118 @@
...
+  const [openCollapse, setOpenCollapse] = useState(null);
```
phulin:
```suggestion
  const [filterCollapseOpen, setFilterCollapseOpen, toggleFilterCollapseOpen] = useToggle(false);
```

src/components/CubeAnalysisNavbar.js
```diff
@@ -0,0 +1,118 @@
...
+            </NavItem>
+          </Nav>
+        </Collapse>
+      </Navbar>
+      <FilterCollapse
+        defaultFilterText={defaultFilterText}
+        filter={filter}
+        setFilter={setFilter}
+        numCards={numCards}
+        isOpen={openCollapse === 'filter'}
```
phulin:
```suggestion
        isOpen={filterCollapseOpen}
```

src/components/CubeAnalysisNavbar.js
```diff
@@ -0,0 +1,118 @@
...
+    <div className="usercontrols">
+      <Navbar expand="md" className="navbar-light">
+        <div className="d-flex flex-row flex-nowrap justify-content-between" style={{ flexGrow: 1 }}>
+          {dropdownElement}
+          <NavbarToggler onClick={toggle} />
+        </div>
+        <Collapse isOpen={isOpen} navbar>
+          <Nav className="ml-auto" navbar>
+            <NavItem>
+              <NavLink href="#" data-target="filter" onClick={handleOpenCollapse}>
```
phulin:
```suggestion
              <NavLink href="#" data-target="filter" onClick={toggleFilterCollapseOpen}>
```

src/components/CubeAnalysisNavbar.js
```diff
@@ -0,0 +1,118 @@
...
+        </Col>
+      </Row>
+    );
+  }
+  return (
+    <div className="usercontrols">
+      <Navbar expand="md" className="navbar-light">
+        <div className="d-flex flex-row flex-nowrap justify-content-between" style={{ flexGrow: 1 }}>
+          {dropdownElement}
+          <NavbarToggler onClick={toggle} />
```
phulin:
```suggestion
          <NavbarToggler onClick={toggleIsOpen} />
```

src/components/MagicMarkdown.js
```diff
@@ -0,0 +1,92 @@
...
+        }
+        if (section.startsWith('[[!')) {
+          const cardIndex = parseInt(section.substring(3, section.length - 2), 10);
+          const card = cube.cards[cardIndex];
+          return (
+            <a
+              key={/* eslint-disable-line react/no-array-index-key */ `card.cardID-${position}`}
+              href={getTCGLink(card)}
+            >
+              <img src={card.details.image_normal} className="card-img-top" alt={card.details.name} />
```
phulin: There's a FoilCardImage component you might want to use here that has autocard and stuff

src/components/FilterCollapse.js
```diff
@@ -283,8 +283,10 @@ class FilterCollapse extends Component {
     this.updateFilters(defaultFilter);
   }
 
-  componentDidUpdate() {
-    Query.set('f', this.state.filterInput);
+  componentDidUpdate(prevProps, prevState) {
+    if (prevProps.filter === this.props.filter) {
```
phulin:
```suggestion
    if (prevProps.filter !== this.props.filter) {
```

src/components/analytics/Chart.js
```diff
@@ -0,0 +1,43 @@
...
+// Data should be:
+// {
+//   type: 'chart',
+//   description: str,
+//   chartType: 'doughnut'|'pie'|'line'|'bar'|'horizontalBar'|'radar'|'polarArea'|'bubble'|'scatter'
+//   datasets: [], data field of Chart
+//   options: [], options field of Chart
+// }
+// See https://github.com/jerairrest/react-chartjs-2 for more information.
+class AnalyticsBarChart extends Component {
```
phulin:
```suggestion
const AnalyticsBarChart = ({ data }) => (
```

PR #882, opened 2020-01-16
Fixes for custom drafts:  UI, more feedback for filter errors, adfan...

RKelln:
- small fixes for custom draft UI
  - remove pack works again
  - can't remove last card in pack 
- check filters when saving a custom draft and show warnings if no cards for any of the filters
- change asfan custom draft code to use separate functions rather than polluting existing
  (note: quite likely I broke something, but code isn't being run yet)
- filterToString() help in Filter for turning parsed filters back into strings (not perfectly but close enough for warnings at this point)

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -191,6 +191,19 @@ router.post('/format/add/:id', ensureAuth, async (req, res) => {
       };
       message = 'Custom format successfully edited.';
     }
+    // check pack formats are sane
+    let draftcards = cube.cards.slice();
+    if (draftcards.length == 0) {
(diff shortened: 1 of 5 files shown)


RKelln:
Err, wait, caught a few more little bugs with this.

dekkerglen:
The 3rd bullet point stresses me out. Do you mean that those separate functions aren't in use yet anyway?

RKelln:
Correct functions are defined but not called from anywhere yet.

dekkerglen:
Is this ready to merge?

RKelln:
Yup, lemme just change that one cube save call.

PR #931, opened 2020-01-19
Fixed custom format modal state

### LINKED ISSUE #920: Edit custom format loads incorrect state
When I went to edit my ELD custom draft format, the radio button for set draft/cube draft was incorrectly set:

image

dekkerglen:
Fixes #920


mawills:
Did this actually fix the bug? It should now render with the correct `multiples` value, but there isn't any logic for saving the value of `multiples`. Should probably reopen. Working on a fix.

PR #936, opened 2020-01-19
Add a way to replace a cube with a new csv

ruler501:
Also generates a changelog.


routes/cube_routes.js
```diff
@@ -1389,6 +1344,120 @@ async function bulkuploadCSV(req, res, cards, cube) {
...
+          only_a,
+          only_b,
+          all_cards,
+        });
+      });
+    });
+  }
+}
+
+async function bulkReplaceCSV(req, res, list, cube) {
```
dekkerglen: helper functions are good - please keep them out of cube_routes. Move them into cubefn

src/components/CubeListNavbar.js
```diff
@@ -348,6 +380,7 @@ const CubeListNavbar = ({
                     <DropdownItem disabled>Import</DropdownItem>
                     <PasteBulkModalItem>Paste Text</PasteBulkModalItem>
                     <UploadBulkModalItem>Upload File</UploadBulkModalItem>
+                    <UploadBulkReplaceModalItem>Upload Replacement File</UploadBulkReplaceModalItem>
```
dekkerglen: Can you change this "Replace with File Upload"

routes/cube_routes.js
```diff
@@ -1389,6 +1344,120 @@ async function bulkuploadCSV(req, res, cards, cube) {
...
+
+function compareCubes(req, res, cubeA, cubeB) {
+  if (!cubeA) {
+    req.flash('danger', 'Base cube not found');
+    res.status(404).render('misc/404', {});
+  } else if (!cubeB) {
+    req.flash('danger', 'Comparison cube was not found');
+    res.redirect('/cube/list/' + cubeA._id);
+  } else {
+    return new Promise(function(resolve, reject) {
```
dekkerglen: Please use async/await - no callbacks, and no promises

ruler501: I hadn't realized there was a way to get prices without doing it this way. Updating it now.

src/components/CubeListNavbar.js
```diff
@@ -108,6 +108,38 @@ const UploadBulkModal = ({ isOpen, toggle }) => {
...
+  const { cubeID } = useContext(CubeContext);
+  return (
+    <Modal isOpen={isOpen} toggle={toggle} labelledBy="uploadReplacementModalTitle">
+      <ModalHeader id="uploadReplacementModalTitle" toggle={toggle}>
+        Bulk Upload - Replace with CSV File Upload
+      </ModalHeader>
+      <CSRFForm method="POST" action={`/cube/bulkreplacefile/${cubeID}`} encType="multipart/form-data">
+        <ModalBody>
+          <p>
+            Acceptable files are either .txt (plaintext) with one card name per line, or .csv with the exact format as
```
dekkerglen: This conflicts with the title of 'Replace with CSV File Upload'

routes/cube_routes.js
```diff
@@ -1248,10 +1199,61 @@ router.post('/bulkuploadfile/:id', ensureAuth, async (req, res) => {
...
+      }
+    }
+  } catch (err) {
+    console.log(err);
+    req.flash('danger', 'Error making bulk replacement');
+    return res.redirect(`/cube/list/${req.params.id}`);
+  }
+});
+
+async function CSVtoCards(cards) {
```
dekkerglen: this looks like a good method to move into cubefn

routes/cube_routes.js
```diff
@@ -1278,16 +1280,18 @@ async function bulkuploadCSV(req, res, cards, cube) {
...
       }
-      changelog += addCardHtml(carddb.cardFromId(card.cardID));
     } else {
       missing += card.name + '\n';
     }
   }
+  return { newCards: newCards, newMaybe: newMaybe, missing: missing };
+}
 
+async function generateBlogpost(req, res, cube, changelog, added, missing) {
```
dekkerglen: this looks like a good function to move into cubefn.js

routes/cube_routes.js
```diff
@@ -1339,6 +1342,82 @@ async function bulkuploadCSV(req, res, cards, cube) {
   }
 }
 
+async function bulkuploadCSV(req, res, cards, cube) {
```
dekkerglen: this looks like a good method to move into cubefn

routes/cube_routes.js
```diff
@@ -1339,6 +1342,82 @@ async function bulkuploadCSV(req, res, cards, cube) {
...
+  let changelog = '';
+  let { newCards, newMaybe, missing } = await CSVtoCards(cards);
+  newCards.forEach((card) => (changelog += addCardHtml(carddb.cardFromId(card.cardID))));
+  cube.cards.push(...newCards);
+  cube.maybe.push(...newMaybe);
+  await generateBlogpost(req, res, cube, changelog, newCards.concat(newMaybe), missing);
+}
+
+async function compareCubes(req, res, cubeA, cubeB) {
```
dekkerglen: this looks like a good method to move into cubefn

src/components/CubeListNavbar.js
```diff
@@ -108,6 +108,35 @@ const UploadBulkModal = ({ isOpen, toggle }) => {
...
+const UploadBulkReplaceModal = ({ isOpen, toggle }) => {
+  const { cubeID } = useContext(CubeContext);
+  return (
+    <Modal isOpen={isOpen} toggle={toggle} labelledBy="uploadReplacementModalTitle">
+      <ModalHeader id="uploadReplacementModalTitle" toggle={toggle}>
+        Bulk Upload - Replace with CSV File Upload
+      </ModalHeader>
+      <CSRFForm method="POST" action={`/cube/bulkreplacefile/${cubeID}`} encType="multipart/form-data">
+        <ModalBody>
+          <p>Acceptable files are .csv files with the exact format as our .csv export.</p>
```
dekkerglen: Can we add a new sentence before this one stating that this upload will *replace* the current contents of your cube?

routes/cube_routes.js
```diff
@@ -1339,6 +1342,82 @@ async function bulkuploadCSV(req, res, cards, cube) {
   }
 }
 
+async function bulkuploadCSV(req, res, cards, cube) {
+  let changelog = '';
+  let { newCards, newMaybe, missing } = await CSVtoCards(cards);
+  newCards.forEach((card) => (changelog += addCardHtml(carddb.cardFromId(card.cardID))));
+  cube.cards.push(...newCards);
+  cube.maybe.push(...newMaybe);
+  await generateBlogpost(req, res, cube, changelog, newCards.concat(newMaybe), missing);
```
dekkerglen: this looks like suspicious chaining to me. it might be a better pattern to return the last 4 parameters and then have those passed into generateBlogpost

routes/cube_routes.js
```diff
@@ -1248,103 +1203,72 @@ router.post('/bulkuploadfile/:id', ensureAuth, async (req, res) => {
...
+    if (!cube) {
+      req.flash('danger', 'Cube not found');
+      return res.status(404).render('misc/404', {});
+    } else if (cube.owner != req.user._id) {
+      req.flash('danger', 'Not Authorized');
+      return res.redirect('/cube/list/' + req.params.id);
+    }
+    cards = items.match(/[^\r\n]+/g);
+    if (cards) {
+      if (cards[0].trim() == CSV_HEADER) {
```
RKelln: Should we require the header or just remove it if present? Seems reasonable to make it optional, if not then specify in error message that the header is missing?

routes/cube_routes.js
```diff
@@ -1248,103 +1203,72 @@ router.post('/bulkuploadfile/:id', ensureAuth, async (req, res) => {
...
+            },
+            carddb,
+            prices,
+          );
+          let changelog = '';
+          only_a.forEach((card) => (changelog += removeCardHtml(carddb.cardFromId(card.cardID))));
+          only_b.forEach((card) => (changelog += addCardHtml(carddb.cardFromId(card.cardID))));
+          cube.cards = newCards;
+          cube.maybe = newMaybe;
+          await generateBlogpost(req, res, cube, changelog, only_b.concat(newMaybe), missing, carddb);
```
RKelln: In general I'm not a fan of passing request around, but defer to consensus.
I'd pass the info that the function needs (req.param.id) and throw errors. This means the route controller handles the request failures and the unit tests can be simpler. The flash message then becomes the error message.
Additionally if all errors are thrown and caught in the outer try/catch then there is a single place to flash and redirect.

routes/cube_routes.js
```diff
@@ -1248,103 +1203,72 @@ router.post('/bulkuploadfile/:id', ensureAuth, async (req, res) => {
...
+        } catch (e) {
+          console.error(e);
```
RKelln: rethrow to be caught later or redirect?

__tests__/serverjs/cubefn.test.js
```diff
@@ -204,3 +210,107 @@ test('addAutocard correctly replaces autocard format strings', () => {
...
+    expectSame(newCards[0], expectedCard);
+    expect(newMaybe.length).toBe(1);
+    expectSame(newMaybe[0], expectedMaybe);
+    expect(missing).toBe('');
+  });
+});
+
+describe('compareCubes', () => {
+  it('can calculate the diff between two cubes', async () => {
+    await carddb.initializeCardDb(fixturesPath, true);
```
RKelln: Might be better in a beforeAll block?

routes/cube_routes.js
ruler501: It's the main way we differentiate formats at this point in time. Eventually I'd imagine replacement would support all the other import methods we currently handle like CubeTutor and Plaintext.

routes/cube_routes.js
ruler501: I think this is something that would probably be better taken up as a separate task to restructure the code to whatever we think is best. I do agree that not passing them around would likely be better.

routes/cube_routes.js
ruler501: I think this was holdover code from  the previous implementation. We're already in a try block so I think just letting it propagate up is probably best here.

__tests__/serverjs/cubefn.test.js
ruler501: I was trying to follow the existing pattern, and I think there are some tests that don't use it currently, like `test('getBasics returns the expected set of basic lands', () => {` from line 117.

routes/cube_routes.js
dekkerglen: Ideally we would parse the header and map the column index to the corresponding field, so our csv import would be backwards- compatible

dekkerglen:
conflicts

routes/cube_routes.js
```diff
@@ -22,14 +22,19 @@ var {
...
 } = require('../serverjs/cubefn.js');
 const analytics = require('../serverjs/analytics.js');
 const draftutil = require('../dist/utils/draftutil.js');
 const cardutil = require('../dist/utils/Card.js');
 const carddb = require('../serverjs/cards.js');
 carddb.initializeCardDb();
 const util = require('../serverjs/util.js');
-const { GetPrices } = require('../serverjs/prices.js');
+const prices = require('../serverjs/prices.js');
+const { GetPrices } = prices;
```
dekkerglen: this is not a good pattern. Is this from mixing default and named exports? Can we not do this?

routes/cube_routes.js
```diff
@@ -422,42 +424,31 @@ router.get('/overview/:id', async (req, res) => {
...
-    }
-
+    const cardDetailsQ = populateCardDetails([cube.cards], carddb, prices, null, undefined, true);
     const userQ = User.findById(cube.owner);
     const blogsQ = Blog.find({
       cube: cube._id,
     }).sort('date');
-    const priceDictQ = GetPrices([...pids]);
-    const [user, blogs, priceDict] = await Promise.all([userQ, blogsQ, priceDictQ]);
+    const [user, blogs, cardDetails] = await Promise.all([userQ, blogsQ, cardDetailsQ]);
```
dekkerglen: cardDetails is not used anywhere in this route from what I can see. If populateCardDetails is modifying the collection that's passed in, that's an anti-pattern

routes/cube_routes.js
```diff
@@ -682,56 +673,8 @@ router.get('/compare/:id_a/to/:id_b', async (req, res) => {
...
-        only_b.splice(b_names.indexOf(card.details.name), 1);
-
-        a_names.splice(a_names.indexOf(card.details.name), 1);
-        b_names.splice(b_names.indexOf(card.details.name), 1);
-      }
-    });
-
-    let all_cards = in_both.concat(only_a).concat(only_b);
-
+    await populateCardDetails([cubeA.cards, cubeB.cards], carddb, prices);
```
dekkerglen: Modifying collections passed in as a side effect is an anti-pattern, please return the values you need and don't modify the passed in data

serverjs/cubefn.js
```diff
@@ -228,6 +231,218 @@ async function getElo(cardnames, round) {
...
+        },
+      );
+    });
+  }
+}
+
+// prices should be the prices module with the GetPrices function.
+// elo should be in the form { round: bool }.
+// requested details is a string to pass to carddb.cardFromId.
+async function populateCardDetails(
```
phulin: I think this function is a lot more complicated than it needs to be. I don't think it needs to do the allDetails thing in addition to everything else, since that's only used in one place. I'm worried about it being written in such an imperative style - instead, I'd rather have it as a pure function. I can give you some suggestions on the discord

routes/cube_routes.js
```diff
@@ -1248,103 +1159,59 @@ router.post('/bulkuploadfile/:id', ensureAuth, async (req, res) => {
...
+    }
+    cards = items.match(/[^\r\n]+/g);
+    if (cards) {
+      if (cards[0].trim() == CSV_HEADER) {
+        cards.splice(0, 1);
+        const { newCards, newMaybe, missing } = CSVtoCards(cards, carddb);
+        await populateCardDetails([cube.cards, newCards, newMaybe], carddb, prices);
+        const { only_a, only_b, in_both, all_cards } = await compareCubes(cube.cards, newCards);
+        let changelog = '';
+        only_a.forEach((card) => (changelog += removeCardHtml(carddb.cardFromId(card.cardID))));
```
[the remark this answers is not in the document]
ruler501: We should put this in a style guide or Contributing since it seems forEach was being used throughout much of the code which led to me copying it to use when I added things.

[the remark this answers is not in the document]
dekkerglen: This is something Patrick had confronted me about in person last weekend, so we decided to enforce it moving forward. Not a big deal either way as it's an easy change

serverjs/cubefn.js
dekkerglen: I think this PR is starting to suffer from trying to do too much - refactor routes, replacing duplicate code with helper functions, in addition to the new csv import. Going forward let's try to create more granular PRs - this will also make code reviews a lot easier, and conflicts less likely

routes/cube_routes.js
ruler501: I think I'll leave that for a separate PR since this is pretty large as is.

serverjs/cubefn.js
```diff
@@ -228,6 +252,172 @@ async function getElo(cardnames, round) {
...
+}
+
+// prices should be the prices module with the GetPrices function.
+// elo should be in the form { round: bool }.
+// requested details is a string to pass to carddb.cardFromId.
+async function populateCardDetails(cardLists, carddb, { GetPrices = null, elo = null, requested_details = undefined }) {
+  const pids = new Set();
+  const cardNames = new Set();
+  const lists = cardLists.map((list) => [...list]);
+  for (let i = 0; i < lists.length; i++) {
```
phulin:
```suggestion
  for (const cards of lists) {
```

serverjs/cubefn.js
```diff
@@ -228,6 +252,172 @@ async function getElo(cardnames, round) {
...
+    const cards = lists[i];
+    for (let j = 0; j < cards.length; j++) {
```
phulin:
```suggestion
    for (const card of cards) {
```

serverjs/cubefn.js
```diff
@@ -228,6 +252,172 @@ async function getElo(cardnames, round) {
...
+    }
+  }
+  req.flash('success', 'All cards successfully added.');
+  return res.redirect(`/cube/list/${req.params.id}`);
+}
+
+// prices should be the prices module with the GetPrices function.
+// elo should be in the form { round: bool }.
+// requested details is a string to pass to carddb.cardFromId.
+async function populateCardDetails(cardLists, carddb, { GetPrices = null, elo = null, requested_details = undefined }) {
```
phulin:
```suggestion
async function populateCardDetails(cardLists, carddb, { prices = false, elo = false, requestedDetails }) {
```
Why does this take a list of lists instead of just a list?

I'm also still not super happy with this giant function - I don't like things with side effects like this. I'd rather err on the side of having some slight repeated code elsewhere in our codebase. Instead, you could imagine a pattern where we have functions like getElo that return `id: { elo }` map and getPrices2 that returns `id: { price, price_foil }` map so that you can easily spread them into an existing card object. That architecture makes more sense to me.

routes/cube_routes.js
```diff
@@ -206,10 +210,11 @@ router.post('/format/add/:id', ensureAuth, async (req, res) => {
       message = 'Custom format successfully edited.';
     }
     // check pack formats are sane
-    const draftcards = cube.cards.map((card) => Object.assign(card, { details: carddb.cardFromId(card.cardID) }));
+    let draftcards = cube.cards.map((card) => Object.assign(card, { details: carddb.cardFromId(card.cardID) }));
```
phulin: this line is unnecessary

routes/cube_routes.js
```diff
@@ -206,10 +210,11 @@ router.post('/format/add/:id', ensureAuth, async (req, res) => {
...
     if (draftcards.length === 0) {
       throw new Error('Could not create draft: no cards');
     }
+    [draftcards] = populateCardDetails([draftcards], carddb, {});
```
phulin:
```suggestion
    [draftcards] = populateCardDetails([...cube.cards], carddb, {});
```

serverjs/cubefn.js
```diff
@@ -228,6 +252,172 @@ async function getElo(cardnames, round) {
...
+        newCards.push(card);
+      }
+    } else {
+      missing += `${card.name}\n`;
+    }
+  }
+  return { newCards, newMaybe, missing };
+}
+
+async function generateBlogpost(req, res, cube, changelog, added, missing, carddb) {
```
phulin: Can we name this function something more appropriate and put it back into cube_routes? It's doing stuff other than generating the blogpost and I don't like having what is effectively a route handler outside the routes files.

serverjs/cubefn.js
ruler501: It takes a list of lists for the cases where we're working with multiple lists like comparisons. Doing it all in one batch lets us batch things like the price api calls. I did try to prevent this from mutating any of the arguments, and return a new list, but unless we want to do a deep copy it'll have to modify the card objects.

dekkerglen:
failing ci, you probably need to merge master in
also conflicts

dekkerglen:
Conflicts

routes/cube_routes.js
```diff
@@ -970,6 +896,48 @@ router.get('/samplepackimage/:id/:seed', async (req, res) => {
...
+        : undefined,
+      reactProps: serialize(reactProps),
+      cube,
+      cube_id: req.params.id,
+      title: `${abbreviate(cube.name)} - Bulk Upload`,
+    });
+  }
+  await blogpost.save();
+  cube = setCubeType(cube, carddb);
+  try {
```
dekkerglen: we should wrap this entire function in a try/catch.

routes/cube_routes.js
```diff
@@ -1186,199 +1114,67 @@ router.post('/uploaddecklist/:id', ensureAuth, async (req, res) => {
...
-      cube,
-    );
-    req.flash('success', 'All cards successfully added.');
-    return res.redirect(`/cube/list/${req.params.id}`);
-  } catch (err) {
-    req.logger.error(err);
-    req.flash('danger', 'Error adding cards. Please try again.');
-    return res.redirect(`/cube/list/${req.params.id}`);
-  }
+  return updateCubeAndBlog(req, res, cube, changelog, added, missing);
```
dekkerglen: needs `await`

routes/cube_routes.js
```diff
@@ -1418,7 +1214,57 @@ router.post('/bulkuploadfile/:id', ensureAuth, async (req, res) => {
...
+  }
+});
+
+router.post('/bulkreplacefile/:id', ensureAuth, async (req, res) => {
+  if (!req.files) {
+    req.flash('danger', 'Please attach a file');
+    return res.redirect(`/cube/list/${req.params.id}`);
+  }
+
+  const items = req.files.document.data.toString('utf8'); // the uploaded file object
```
dekkerglen: include this in try catch

routes/cube_routes.js
```diff
@@ -1418,7 +1214,57 @@ router.post('/bulkuploadfile/:id', ensureAuth, async (req, res) => {
...
 
-    return await bulkUpload(req, res, items, cube);
+    return bulkUpload(req, res, items, cube);
+  } catch (err) {
+    return util.handleRouteError(req, res, err, `/cube/list/${req.params.id}`);
+  }
+});
+
+router.post('/bulkreplacefile/:id', ensureAuth, async (req, res) => {
+  if (!req.files) {
```
dekkerglen: include this is in try catch

routes/cube_routes.js
```diff
@@ -1418,7 +1214,57 @@ router.post('/bulkuploadfile/:id', ensureAuth, async (req, res) => {
...
+        }
+        for (const card of onlyB) {
+          changelog += addCardHtml(carddb.cardFromId(card.cardID));
+        }
+        cube.cards = newCards;
+        cube.maybe = newMaybe;
+        added.push(...onlyB);
+      }
+    }
+    return updateCubeAndBlog(req, res, cube, changelog, added, missing);
```
dekkerglen: needs `await`

serverjs/cubefn.js
```diff
@@ -1,6 +1,10 @@
 const sanitizeHtml = require('sanitize-html');
-const Cube = require('../models/cube');
+const serialize = require('serialize-javascript');
```
dekkerglen: Is this used?

serverjs/cubefn.js
```diff
@@ -232,6 +256,128 @@ async function getElo(cardnames, round) {
...
+// prices should be the prices module with the GetPrices function.
+// elo should be in the form { round: bool }.
+// requested details is a string to pass to carddb.cardFromId.
+async function populateCardDetails(cardLists, carddb, { GetPrices = null, elo = null, requested_details = undefined }) {
+  const pids = new Set();
+  const cardNames = new Set();
+  const lists = cardLists.map((list) => [...list]);
+  for (const cards of lists) {
+    for (const card of cards) {
+      card.details = {
```
dekkerglen: why not just do 
`card.details = carddb.cardFromId(card.cardID, requested_details)`

serverjs/cubefn.js
```diff
@@ -232,6 +256,128 @@ async function getElo(cardnames, round) {
...
+      missing += `${card.name}\n`;
+    }
+  }
+  return { newCards, newMaybe, missing };
+}
+
+// prices should be the prices module with the GetPrices function.
+// elo should be in the form { round: bool }.
+// requested details is a string to pass to carddb.cardFromId.
+async function populateCardDetails(cardLists, carddb, { GetPrices = null, elo = null, requested_details = undefined }) {
```
dekkerglen: You are using three different naming conventions for these flags, please commit to one

serverjs/cubefn.js
ruler501: We need a copy to prevent accidental modification of the carddb.

routes/cube_routes.js
ruler501: This should always be called from a route that's wrapped in a try catch block.

routes/cube_routes.js
ruler501: This is an async function so you don't need to await the return value, it's an eslint error to put an await here.

dekkerglen: If it's the last line in the route, use await instead of return

routes/cube_routes.js
dekkerglen: Doesn't matter, it costs nothing, and can help prevent issues in the future

dekkerglen:
Conflicts

routes/cube_routes.js
dekkerglen: Would like to see this change still

phulin:
Why aren't we just adding a checkbox to the existing modals? There's a ton of stuff going on in this PR that seems kind of unnecessary.

ruler501:
We aren't doing a checkbox partly because it wasn't brought up before and I didn't think of it. We could change it to work that way, but one benefit of the current method is that it does a lot to make it very clear that the operation will wipe all the cards in your cube and replace them to hopefully avoid people mistakenly using it.

This PR definitely ended up doing too much with the add card details stuff, and I'm tyring to avoid that in more recent PRs though the analytics one has grown quite a bit, partly because it's been open so long and any new work on analytics tends to build off of it.

dekkerglen:
CI Fails

PR #939, opened 2020-01-20
Add CustomFormat class and render an error message when not all of it…

mawills:
…s attributes are present


mawills:
Going to investigate using Typescript instead.

PR #941, opened 2020-01-20
Fix several issues with server-side rendering and reenable it.

phulin:
Note for future me and other developers: Server-side rendering will fail in very strange ways if the tree seen by `ReactDOM.hydrate` doesn't match what was rendered on the server. The most common ways this happens are using the URL hash or `localStorage` as inputs to the original render. Instead, we should use `componentDidMount` as the place to check `localStorage`.

I haven't been able to find the dev blog bug. If we haven't found it closer to next release, we can disable SSR for that page only.


dekkerglen:
Failing CI

phulin:
Fixed

PR #947, opened 2020-01-21
Elo updates

### LINKED ISSUE #872: Card ELO sort option
As a user, I would like to be able to sort my cube by the cards' ELO.

### LINKED ISSUE #873: More Information on Card Pages
As a user, on card pages I would like to see information about the card including:
Price, Foil Price (if it exists), ELO

### LINKED ISSUE #874: View ELO on card modal
As a user, I would like to be able to check the ELO of a card easily.

Can we put it right underneath where we put the price in the context modal? Makes sense since it's a field we can't edit, like price.

### LINKED ISSUE #877: Use Elo for draft bots
The Elo data looks good enough now that I think we're ready to use Elo for our draft bots.

dekkerglen:
Fixes #877, Fixes #874, Fixes #872, Fixes #873

### DIFF
--- a/src/util/Draft.js
+++ b/src/util/Draft.js
@@ -47,21 +47,26 @@ const fetchLands = [
 function botRating(botColors, card) {
   let rating = draft.ratings[card.details.name];
   const colors = card.colors || card.details.color_identity;
-  const subset = arrayIsSubset(colors, botColors);
+  const subset = arrayIsSubset(colors, botColors) && colors.length > 0;
   const overlap = botColors.some((c) => colors.includes(c));
   const typeLine = card.type_line || card.details.type;
   const isLand = typeLine.indexOf('Land') > -1;
   const isFetch = fetchLands.includes(card.details.name);
 
-  // Prioritize on-color or overlapping fetches.
-  // Then overlapping lands, then overlapping spells.
-  if (subset || (isFetch && overlap)) {
-    rating -= 0.3;
-  } else if (isLand && overlap) {
-    rating -= 0.2;
+  if (isLand) {
+    if (subset) {
+      rating *= 1.4;
+    } else if (overlap) {
+      rating *= 1.2;
+    } else {
+      rating *= 1.1;
+    }
+  } else if (subset) {
+    rating *= 1.3;
   } else if (overlap) {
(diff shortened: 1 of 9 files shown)


src/util/Draft.js
```diff
@@ -81,12 +88,13 @@ function botPicks() {
...
 
     ratedPicks.sort((x, y) => {
-      return botRating(botColors, pack[x]) - botRating(botColors, pack[y]);
+      return botRating(botColors, pack[y]) - botRating(botColors, pack[x]);
     });
     arrayShuffle(unratedPicks);
 
     const pickOrder = ratedPicks.concat(unratedPicks);
     pick = pack.splice(pickOrder[0], 1);
+    console.log(pick[0].details.name + ' - ' + botColors);
```
phulin: Take this out

src/util/Sort.js
```diff
@@ -335,6 +336,31 @@ export function getLabels(cube, sort) {
...
+    cube.forEach(function(card, index) {
+      if (card.details.elo) {
+        if (!items.includes(card.details.elo)) {
+          items.push(card.details.elo);
+        }
+      }
+    });
+    return items.sort(function(x, y) {
+      if (!/^\d+$/.test(x) || !/^\d+$/.test(y)) {
+        if (x > y) {
```
phulin: Shouldn't all the elos be integers? Why do you need to do this?

src/util/Sort.js
```diff
@@ -335,6 +336,23 @@ export function getLabels(cube, sort) {
     return labels;
   } else if (sort == 'Unsorted') {
     return ['All'];
+  } else if (sort == 'Elo') {
+    var items = [];
+    cube.forEach(function(card, index) {
```
RKelln: Buckets?
- find elo range from cards (min to max)
- bucket_size = 10 * elo_range / num_cards_in_cube   
- bucket size min 1

phulin: Just so it's written down somewhere, the Elo system should be centered at 1200 with a range of roughly 400-2000, so buckets can be made accordingly.

dekkerglen: I'll do buckets of 50, and just use math to divide it up

PR #948, opened 2020-01-21
Move util dir to utils to avoid name conflict.

phulin:
For some bizarre reason one of our dependencies has the module name `util` which was causing name conflicts.


dekkerglen:
This is failing CI checks

phulin:
Should be fixed now?

PR #958, opened 2020-01-22
Improve cube prices.

phulin:
This PR adds an as-owned and a to-purchase price for each cube.

image

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -406,43 +406,58 @@ router.post('/unfeature/:id', ensureAuth, async (req, res) => {
 
 router.get('/overview/:id', async (req, res) => {
   try {
-    var split = req.params.id.split(';');
-    var cube_id = split[0];
-    var currentUser;
-    admin = false;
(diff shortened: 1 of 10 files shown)


routes/cube_routes.js
```diff
@@ -406,43 +406,61 @@ router.post('/unfeature/:id', ensureAuth, async (req, res) => {
...
+    const priceDictQ = GetPrices([...pids]);
+    const [user, blogs, priceDict] = await Promise.all([userQ, blogsQ, priceDictQ]);
+
+    let totalPriceOwned = 0;
+    let totalPricePurchase = 0;
+    for (const card of cube.cards) {
+      if (card.status !== 'Not Owned') {
+        let priceOwned = 0;
+        if (card.finish === 'Foil') {
+          priceOwned = priceDict[card.details.tcgplayer_id + '_foil'] || 0;
```
ruler501: In the case foil price isn't available would it make sense to fall back on non-foil pricing?

PR #961, opened 2020-01-23
Maybeboard UX improvements

phulin:
Move the add function to a "+" button and make it only fill in the add input and move focus to the remove input. Enable the card modal so people can change versions and other data around maybeboard cards.

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -3081,7 +3081,7 @@ router.post(
       return res.status(400).send({
         success: 'false',
         message: 'Bad request.',
-      })
+      });
     }
     const newVersion = updated.cardID && updated.cardID !== card.cardID;
     for (field of ['cardID', 'status', 'finish', 'cmc', 'type_line', 'imgUrl', 'colors']) {
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -3056,6 +3056,53 @@ router.post(
   }),
 );
 
+router.post(
+  '/api/maybe/update/:id',
+  ensureAuth,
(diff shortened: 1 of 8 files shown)


routes/cube_routes.js
```diff
@@ -3056,6 +3056,53 @@ router.post(
...
+      }
+    }
+    await cube.save();
+
+    if (newVersion) {
+      return res.status(200).send({
+        success: 'true',
+        details: carddb.cardFromId(card.cardID),
+      });
+    } else {
```
dekkerglen: else is not necessary 
(removing this removes a bit of nesting which is a plus)

phulin: This is a stylistic point but I prefer else here... if you want me to change it i can

ruler501: Chiming in I prefer having the else as well. It's a little more nesting but it helps make the control flow explicit.

dekkerglen: To me, all of the 3 earlier branches that return follow the pattern of not using else, and if they used else it would cause a lot of chained nesting. This is the last branch so it looks like a different case but it's isn't actually. I'd prefer to be consistent avoiding unnecessary else - and make the code read as linear as possible

phulin: I'll submit to your demands this time but next time.............

phulin: Also, while acknowledging the fact that this is an extremely unimportant argument, the point of this style is to point out that there are two alternate paths which both constitute "success." Normally, the if(x) return; pattern indicates that you're exiting due to some abnormal condition, but these are both equally successful return statements.

phulin:
Fixed fwiw

PR #973, opened 2020-01-28
linting the serverjs analytics files

joshfinnie:
Adding the following rules to eslint.json to have it match our coding style better:

```
"no-plusplus": ["error", {"allowForLoopAfterthoughts": true}],
"no-restricted-syntax": ["error", "FunctionExpression", "WithStatement", "BinaryExpression[operator='in']"]
```


serverjs/analytics.js
```diff
@@ -1,112 +1,65 @@
...
-  }
-  return isValid;
-}
-
 function GetColorCat(type, colors) {
   if (type.toLowerCase().includes('land')) {
     return 'l';
-  } else if (colors.length == 0) {
+  }
+  if (colors.length === 0) {
```
ruler501: I'm not a fan of removing the else's here.

joshfinnie: Then I would not use a `return`... if `type.toLowerCase().includes('land')` the rest of the function does not execute because of the explicit `return 'l';` making the else redundant.

joshfinnie: we can also so something like:

```
let colorCat;

if (<something>) {
  ...
} else if (<another thing>) {
  ...
}
return colorCat;
```

dekkerglen: I like removing elses when not necessary

joshfinnie:
Close due to #977 getting in... should encompass a lot of the same changes...

joshfinnie:
ah, my bad... this is not handled in #977 ... serverjs still needs linting.

PR #979, opened 2020-01-30
Eslint for src directory.

phulin:
This PR enables eslint (blocking commits) for the entire `src` directory. For now, all existing components are whitelisted, but hopefully we can start trimming that list down over time. I've gone through everything outside `src/components` and fixed all the errors.

The pr is based on #977 and should only be merged after that one.


dekkerglen:
Conflicts

phulin:
Conflicts fixed

PR #1005, opened 2020-02-01
Add redirect for non-english cards.

### LINKED ISSUE #997: Card Pages for non-english cards broken
We currently do not index non-english cards from scryfall, but non-english cards still have a unique scryfall id. Our card pages use this scryfall id to find the data, but for these non english cards, the card won't be found and the page won't load. We need to get all language data IDs - and store those in our ID map.

https://scryfall.com/card/dom/101/fr/nich%C3%A9e-de-rats

phulin:
Important note: Using scryfall's all-cards download causes the node default heap to overflow. We'll have to make the heap size bigger on production if we aren't already.

Fixes #997.


dekkerglen:
Our droplets will be gated by RAM, I'll need to increase that before we can use this solution

phulin:
Ok i’ll think about this

dekkerglen:
I'm not opposed to increasing the RAM of droplets to 4GB, it would just cost me an extra ~20 a month. not a big deal

phulin:
We could also just use this: https://www.npmjs.com/package/JSONStream

phulin:
I think i have a solution, will push when i fix up the tests

phulin:
Should be good to go now. JSON object now streams through memory

dekkerglen:
Awesome, I can merge after prettier

PR #1006, opened 2020-02-02
Fix CubePlaytestPage.js

ruler501:
Removes a warning from the console of cards being null and fixes the error with useAlerts preventing it from rendering.


src/pages/CubePlaytestPage.js
```diff
@@ -345,7 +345,7 @@ const CubePlaytestPage = ({ cube, cubeID, canEdit, decks, draftFormats }) => {
 
 CubePlaytestPage.propTypes = {
   cube: PropTypes.shape({
-    cards: PropTypes.arrayOf(PropTypes.object).isRequired,
+    cards: PropTypes.arrayOf(PropTypes.object),
```
phulin: Why this change?

ruler501: I was getting warnings in the browser that we were passing a cube without cards and I checked and we aren't using cards in this file so it didn't seem to need to be required.

PR #1017, opened 2020-02-06
Change custom drafts to use id instead of index

RKelln:
Changes custom drafts to use id instead of index.

- Adds draft format helpers to `cubefn.js`.
- 2 other react errors on console for playtest page fixed
- improves format delete route so there aren't weird parameters

### DIFF
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -259,7 +259,6 @@ function removeDraftFormat(cube, id) {
 
 // returns -1 if format id not found
 function getDraftFormatIndex(cube, id) {
-  console.log('getDraftFormatIndex', cube, id);
   if (!cube.draft_formats) {
     throw new Error('Cube has no draft formats');
   }
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -235,7 +235,7 @@ function addDraftFormat(cube, { title, multiples, html, packs }) {
   cube.draft_formats.push({ title, multiples, html, packs });
 }
 
-function editDraftFormat(cube, id, { title, multiples, html, packs }) {
+function updateDraftFormat(cube, id, { title, multiples, html, packs }) {
   if (!cube.draft_formats) {
     throw new Error('Cube has no draft formats');
   }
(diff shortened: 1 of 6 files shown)


src/pages/CubePlaytestPage.js
```diff
@@ -272,38 +273,26 @@ const CubePlaytestPage = ({ cube, cubeID, canEdit, decks, draftFormats }) => {
...
-          color: 'danger',
-          message: 'Failed to delete format.',
-        });
+        addAlert('danger', `Failed to delete format. ${err.message}`);
       }
     },
     [addAlert, cubeID, formats],
   );
 
-  // Sort formats alphabetically.
```
phulin: I’d still prefer to do all sorting on the client, even with robust IDs. Layout should be done client side.

RKelln: Up to you, I agree that sorting should be possible on the client, but generally you should be able to request an order from a server or have it defaulted to a particular deterministic order. If the client changes that order everything should still work (hence the use of IDs and never index).

serverjs/cubefn.js
```diff
@@ -228,6 +228,48 @@ async function getElo(cardnames, round) {
   return result;
 }
 
+function addDraftFormat(cube, { title, multiples, html, packs }) {
```
joshfinnie: Could we add tests for these new functions?

RKelln: Yup can do.
Can I get some feedback first that this is the style that we want and this is the place where they should live? 
I'm used to CRUD functions being attached to controllers/models, which we don't have, so this is something like that but less elegant.
Also using some newfangled js style here, with destructuring the params and but I think it reads pretty well. Thoughts?

phulin:
It would definitely make sense to me to start attaching these to models.

serverjs/cubefn.js
RKelln: Tests added.

ruler501:
Now that the new analytics got merged, you may need to check what this change impacts there.

RKelln:
> Now that the new analytics got merged, you may need to check what this change impacts there.

Yup, thanks, will resolve and let you know if I have any questions. :+1:

dekkerglen:
Conflicts

RKelln:
After solving some conflicts I found a few more bugs, but it now makes sense to put the add/update/deleteDraftFormat functions into draftutls.js which will cause conflicts with @dekkerglen's change, so suggest leaving this until after that change is merged in and then I'll fix everything up.

RKelln:
Note to self: analytics also uses custom draft index.

dekkerglen:
Conflicts from draft and deck changes

RKelln:
Yup, this actually has a lot of work left, no need for reviews or anything yet. Will update when this is ready to be looked at again. Thanks.

dekkerglen:
Conflicts from deck and draft updates

PR #1019, opened 2020-02-07
Add setting for default printing + status.

### LINKED ISSUE #665: Always show oldest/newest print (edition) of a card for the whole cube
While importing or editing a cube it would be nice to be able to choose which print is used in general. For example:
- Always display first print of a card.
- Always display latest print of a card.
- Always display latest print of a card but only regular sets (no pimp ect.).

Its really a pain to edit every single card of a large cube to lets say the first print.

Also Rankings like:
Pimp > Old Border > Latest Print
Regular Set > First Print
PS19 > ELD > Alpha > M10 > ...
ect. would be a possibility

### LINKED ISSUE #1001: Remove MYB from default most reasonable
Make sure default most reasonable does not count MYB. we can add 'MYB' to the list of 'promo' sets for the isPromo field during card import.

phulin:
I've also started using the express request validator in this PR; we should replicate that elsewhere in the routes files.

Fixes #1001, fixes #665.


RKelln:
:+1: Settings UI sorely needed. Maybe some other existing stuff should move/copy here? Things currently under display? Perhaps Set Tag Colors?

Tests for `reasonable` functions?

phulin:
I moved all the things that I thought made sense; the display stuff is mostly local settings, and I want the tag colors to be on the list page

phulin:
Added some tests, this should be ready to go now

PR #1020, opened 2020-02-08
Add files for dockerization

ruler501:
First steps towards enabling dockerization. Have tested these to successfully run under microk8s and docker. Don't have to migrate to using these on prod, but can at least give a consistent environment for debugging with the docker images.


ruler501:
Closing until a time we're more able to address the architecture choices for it.

PR #1022, opened 2020-02-09
Add new social page to see followed users + cubes.

phulin:
Based on #1021. Don't merge until that one's in.

image


dekkerglen:
Is everyone an ambush viper?

phulin:
There's a field in the DB for it but i guess there's no way to change it? I
can add that to this PR

phulin:
Actually I'd like to get this PR in by itself since there will be significant work in getting the profile pics working.

PR #1030, opened 2020-02-11
Lint/serverjs cards

joshfinnie:
This PR gets both `serverjs/cards.js` and `__tests__/serverjs/cards.test.js` passing our ESLint rules.


dekkerglen:
failing ci, you probably need to merge master in

__tests__/serverjs/cards.test.js
```diff
@@ -126,21 +128,21 @@ test('cardFromId returns only selected fields', () => {
...
 
 test('getCardDetails returns a well-formed card object', () => {
   expect.assertions(1);
-  const _id = _RankleMasterofFixtures._id;
-  let expected = _RankleMasterofFixtures;
-  var promise = carddb.initializeCardDb(fixturesPath, true);
-  return promise.then(function() {
+  const { _id } = _RankleMasterofFixtures;
+  const expected = _RankleMasterofFixtures;
+  const promise = carddb.initializeCardDb(fixturesPath, true);
```
ruler501: For all these couldn't we just do `await carddb.initializeCardDb(fixturesPath, true)` to avoid callbacks? Also is that something we can move into a setup block instead of duplicating it in each test?

joshfinnie: yea, that's probably a good update...

joshfinnie: but maybe on a different PR just to keep things simple...

PR #1034, opened 2020-02-12
Correctly handle registration requests.

ruler501:
When creating a registration look at whether it was invalid for whether to send back to existing page, otherwise do the registration flow.


dekkerglen:
failing ci

PR #1064, opened 2020-02-15
Allow Specifying a Default Draft Format

ruler501:
Allows users to specify a default draft format they want to use for their cube for sample packs and p1p1. Makes that format always appear on top in playtesting.

Introduces a seed to draft creation to allow for repeatable p1p1 generations for custom draft formats.

Depends on #1218.


src/pages/CubePlaytestPage.js
```diff
@@ -292,12 +321,58 @@ const CubePlaytestPage = ({ cube, cubeID, canEdit, decks, draftFormats }) => {
     [addAlert, cubeID, formats],
   );
 
+  const handleSetDefaultFormat = useCallback(
+    async (event) => {
+      const formatIndex = parseInt(event.target.getAttribute('data-index'), 10);
+      try {
+        const response = await csrfFetch(`/cube/defaultdraftformat/${cubeID};${formatIndex}`, {
```
RKelln: I prefer cube/cubeID/defaultdraftformat/formatIndex  but realize that is a bit different than standard. I really don't like the `;` so happy if that gets removed.

routes/cube_routes.js
```diff
@@ -3121,6 +3122,30 @@ router.delete('/format/remove/:cubeid/:index', ensureAuth, param('index').toInt(
   }
 });
 
+router.post('/defaultdraftformat/:id', ensureAuth, async (req, res) => {
+  try {
+    const cubeid = req.params.id.split(';')[0];
```
RKelln: I prefer to let the variables be set in the route: e.g.
`:cubeId/defaultdraftFormat/:formatId`

src/utils/draftutil.js
```diff
@@ -241,23 +248,17 @@ function createPacks(draft, format, seats, nextCardFn) {
...
 
 // NOTE: format is an array with extra attributes, see getDraftFormat()
-export function populateDraft(draft, format, cards, bots, seats) {
+export function populateDraft(draft, format, cards, bots, seats, seed = false) {
   let nextCardFn = null;
 
   if (cards.length === 0) {
     throw new Error('Unable to create draft: no cards.');
   }
-  if (bots.length === 0) {
```
RKelln: If you remove these checks then they should be re-added to (before) the routes/calls that do require them.

src/hooks/UseAlerts.js
```diff
@@ -12,7 +12,10 @@ export const Alerts = ({ alerts, ...props }) =>
 const useAlerts = () => {
   const [alerts, setAlerts] = useState([]);
 
-  const addAlert = useCallback((color, message) => setAlerts((oldAlerts) => [...oldAlerts, { color, message }]), []);
```
phulin: This issue has been fixed in master; no need to include this change.

dekkerglen:
Conflicts from draft and deck changes

src/utils/draftutil.js
```diff
@@ -238,25 +244,19 @@ function createPacks(draft, format, seats, nextCardFn) {
 }
 
 // NOTE: format is an array with extra attributes, see getDraftFormat()
-export function populateDraft(format, cards, bots, seats, user) {
+export function createDraft(format, cards, bots, seats, user, seed = false) {
```
dekkerglen: Please keep PRs more granular, and state all changes in description. There is no indication in the PR description that this PR adds a seed to draft creation. Why do we need this?

babel.config.js
```diff
@@ -2,12 +2,16 @@ module.exports = function(api) {
...
+        '@babel/preset-env',
+        {
+          useBuiltIns: 'usage',
+          corejs: 3,
+          exclude: ['es.promise'],
+        },
+      ],
       '@babel/preset-react',
     ],
+    plugins: ['@babel/plugin-proposal-optional-chaining', '@babel/plugin-proposal-nullish-coalescing-operator'],
```
dekkerglen: you did not state that this PR was dependent on #1168, please explicitly state if you have PR dependencies so we can address them one by one

routes/cube_routes.js
```diff
@@ -1580,6 +1580,10 @@ router.post('/startdraft/:id', async (req, res) => {
       cards: parseInt(req.body.cards, 10),
     };
 
+    if (params.seats < 2) {
```
dekkerglen: This should be validated using the express body validator

routes/cube_routes.js
```diff
@@ -3125,6 +3129,36 @@ router.delete('/format/remove/:cubeid/:index', ensureAuth, param('index').toInt(
   }
 });
 
+router.post('/:id/defaultdraftformat/:formatId', ensureAuth, async (req, res) => {
```
dekkerglen: for api error handling, please follow the pattern we have used for other api routes using `util.wrapAsyncApi`

serverjs/cubefn.js
```diff
@@ -315,15 +316,13 @@ const methods = {
     if (!seed) {
       seed = Date.now().toString();
     }
-
-    const pack = util
-      .shuffle(cube.cards, seed)
-      .slice(0, 15)
-      .map((card) => carddb.getCardDetails(card));
-
+    cube.cards = cube.cards.map((card) => ({ ...card, details: { ...carddb.getCardDetails(card) } }));
```
dekkerglen: we don't know if cube has been called with `.lean()`, let's not populate the details here

src/components/CubeAnalysisNavbar.js
```diff
@@ -45,27 +45,27 @@ const CubeAnalysisNavBar = ({
 
   if (draftFormats) {
     dropdownElement = (
-      <Row>
-        <Col>
-          <h5>{formatId >= 0 ? draftFormats[formatId].title : 'Default Draft Format'}</h5>
-        </Col>
-        <Col>
-          <UncontrolledDropdown nav inNavbar>
+      <Col>
```
dekkerglen: Why is `<Col>` wrapping `<Row>`? Should be the other way around. There is no point to having a `<Row>` with no `<Col>` children

src/pages/CubePlaytestPage.js
```diff
@@ -98,14 +98,23 @@ LabelRow.propTypes = {
...
+}) => {
   const { cubeID, canEdit } = useContext(CubeContext);
   const { index } = format;
   return (
     <Card {...props}>
       <CSRFForm method="POST" action={`/cube/startdraft/${cubeID}`}>
         <CardHeader>
-          <CardTitleH5>{format.title} (custom draft)</CardTitleH5>
+          <CardTitleH5>
+            {format.title} (Custom Draft) {defaultDraftFormat === index && '(Default Format)'}
```
dekkerglen: Having two sets of parens trailing the format titles feels wrong

babel.config.js
ruler501: It makes the same changes, but was made prior to #1168. I made that PR since I noticed I had made the same changes on a few branches and it seemed best to pull it out.

src/utils/draftutil.js
ruler501: We have to have a seed to support consistent p1p1 creation for the cube. This was a required change for the stated purpose to work. I can add a description as to how the default format works with p1p1 though.

serverjs/cubefn.js
ruler501: We have to populate details before calling `createDraft` in case it is a custom draft format where filters refer to card details. We might need to populate elo and price as well, but I think that since those are dynamic values not supporting them in this case for custom draft formats is likely okay.

src/pages/CubePlaytestPage.js
ruler501: I'm open to suggestions on other ways to demarcate the Default format.

src/components/CubeAnalysisNavbar.js
ruler501: There does seem to be a slight visual difference when taking them out, though it looks better without the row tags I think. Removing them adds more of a margin between the text and the left edge of the enclosing box. I'll update the PR with the Row tags removed.

src/pages/CubePlaytestPage.js
dekkerglen: How about: Default Format: {name} ({custom draft})

routes/cube_routes.js
```diff
@@ -1813,69 +1813,74 @@ router.get('/deck/download/cockatrice/:id/:seat', async (req, res) => {
...
 });
 
-router.post('/startdraft/:id', async (req, res) => {
-  try {
-    const cube = await Cube.findOne(buildIdQuery(req.params.id), '_id name draft_formats card_count type cards').lean();
+router.post(
+  '/startdraft/:id',
+  body('id').toInt(),
+  body('seats').toInt({ min: 2, max: 16 }),
+  body('packs').toInt(),
```
dekkerglen: I like including min and max here as well

dekkerglen:
Conflicts

PR #1067, opened 2020-02-15
Enable using APM for performance stats

ruler501:
For discussion on whether we want this. Would hopefully help debugging performance issues and optimizing the server side code easier in the future. Would require setting up an APM server, elasticsearch, and probably kibana to use.


ruler501:
Closing until a time we're more able to address the architecture choices for it.

PR #1073, opened 2020-02-16
Add JaceVP emblem

### LINKED ISSUE #966: Jace, Vryn's Prodigy emblem missing from tokens
At least in [my cube](https://cubecobra.com/cube/list/dillon), the generated tokens misses the emblem for Jace, Vryn's Prodigy. 

To reproduce, see here: https://cubecobra.com/cube/analysis/dillon?nav=tokens.

timfreilly:
fixed #966 Adds flip Jace's emblem to the special list of tokens.

### DIFF
--- a/serverjs/updatecards.js
+++ b/serverjs/updatecards.js
@@ -191,7 +191,7 @@ const specialCaseCardsList = {
   ],
   "Jace, Vryn's Prodigy": [
     {
-      tokenId: '458e37b1-a849-41ae-b63c-3e09ffd814e4'
+      tokenId: '458e37b1-a849-41ae-b63c-3e09ffd814e4',
     },
   ],
   "Legion's Landing": [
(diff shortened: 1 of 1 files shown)


timfreilly:
Any idea why the build failed?

dekkerglen:
> Code style issues found in the above file(s). Forgot to run Prettier?

Do `npm run beautify` to fix.

Also, is this case being missed by the parser, or are we just not parsing the back side of flip cards? If it's the latter, this isn't the correct fix

timfreilly:
I looked at the other items in that exceptions list and a lot of them are for flip cards. There's a comment on line 142 talking about it:
  // the cards below are transform cards that are on here due to the way
  // we currently do not populate the oracle text of transform cards.

I was adding this one missing card to that list. Do you think the issue should be changed to  rewrite the code to understand flip cards?

dekkerglen:
Yeah this is fine for now

timfreilly:
Made it pretty

PR #1077, opened 2020-02-16
Improvements to Analytics

ruler501:
Allow prerendering without doing any calculations and add a tagCount analytic.


routes/cube_routes.js
```diff
@@ -737,8 +739,9 @@ router.get('/list/:id', async (req, res) => {
       cardNames.push(card.details.name);
     });
 
-    const priceDict = await GetPrices([...pids]);
-    const eloDict = await getElo(cardNames, true);
+    const priceDictQ = GetPrices([...pids]);
```
dekkerglen: we should just pass `pids` into here

routes/cube_routes.js
```diff
@@ -850,12 +853,14 @@ router.get('/playtest/:id', async (req, res) => {
 
 router.get('/analysis/:id', async (req, res) => {
   try {
-    const cube = await Cube.findOne(buildIdQuery(req.params.id)).lean();
+    const fields = 'cards name owner defaultDraftFormat draft_formats card_count type';
+    const cube = await Cube.findOne(buildIdQuery(req.params.id), fields).lean();
```
dekkerglen: why create a variable here, would be cleaner to just pass the string in directly

routes/tools_routes.js
```diff
@@ -58,7 +58,8 @@ async function matchingCards(filter) {
       cards = Filter.filterCardsDetails(cards, filter);
     }
     if (Filter.filterUses(filter, 'price') || Filter.filterUses(filter, 'price_foil')) {
-      const withPrices = await addPrices(cards);
+      const withPricesPre = await addPrices(cards.map((card) => ({ details: card })));
```
dekkerglen: why are you doing this

ruler501: If we want to be able to use addPrices with cube card objects it needs to take them this way.

routes/cube_routes.js
ruler501: I believe that didn't work because pids is a Set and GetPrices only works with an array.

phulin:
Can you try and separate these into smaller PRs in the future? Very hard to review something with this many moving parts.

ruler501:
Sure

dekkerglen:
Conflicts

PR #1085, opened 2020-02-17
Don't require an X for plain text uploads and pastes.

ruler501:
Removes the requirement since no cards start with integers.


routes/cube_routes.js
```diff
@@ -1275,10 +1275,15 @@ async function bulkUpload(req, res, list, cube) {
   let changelog = '';
   for (let i = 0; i < cards.length; i++) {
     const item = cards[i].toLowerCase().trim();
-    if (/([0-9]+x )(.*)/.test(item)) {
-      const count = parseInt(item.substring(0, item.indexOf('x')), 10);
+    const numericMatch = item.match(/([0-9]+)x? (.*)/);
+    console.warn(numericMatch);
```
timfreilly: Leftover console message?

PR #1087, opened 2020-02-17
255 Encode search parameters

### LINKED ISSUE #255: Encode Search Param
Search params need to be encoded to avoid issues with special characters

gabyreilly:
fixes #255 

There seemed to be two separate issues with special characters in search params. 

1 - Search params that were already HTML special character strings (like %20) were not escaped in the search URL.  Note that the server was already decoding these parameters, resulting in some unusual results:

image

2 - Search params that were special characters for regex were not escaped in the actual search method

image

This PR resolves both issues, and I've validated locally.  I don't see a great spot to write a unit test but if I've missed the right place for a test, please let me know!

### DIFF
--- a/routes/root.js
+++ b/routes/root.js
@@ -294,10 +294,12 @@ router.get('/landing', async (req, res) => {
 router.post('/advanced_search', (req, res) => {
   let url = '/search/';
   if (req.body.name && req.body.name.length > 0) {
-    url += `name${req.body.nameType}${req.body.name};`;
+    const encoded = encodeURIComponent(req.body.name);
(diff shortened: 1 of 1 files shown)


routes/root.js
```diff
@@ -322,6 +324,10 @@ router.get('/search/:id', (req, res) => {
   let page = parseInt(rawSplit[1], 10);
   let query = {};
   const terms = [];
+
+  function regexEscape(input) {
```
ruler501: Could this be named something a little more descriptive? Is it escaping regexes?

PR #1116, opened 2020-02-19
Further refine logging.

phulin:


app.js
```diff
@@ -224,10 +236,25 @@ app.use('/user', users);
 app.use('/dev', devs);
 app.use('/tool', tools);
 
+app.get('/500', (req, res, next) => {
```
ruler501: Should this be taken out?

PR #1119, opened 2020-02-19
1083 Add card meta for use in social media

### LINKED ISSUE #1083: Card page meta data
Currently, card page meta data doesn't have any contextual information. It should have the card image and card name displayed when you share the links on discord etc. Similar to how cube metadata works.

Current:
Screenshot_20200216-160628_Discord

gabyreilly:
fixes #1083 

Social media (like Discord) requires a set of meta tags in the head to be able to show the contextual card. However these meta tags were missing for the single-card analysis pages.

Luckily we had already written a method to generate meta for other pages, so I only had to pull it in for the card analytics page. 

Along the way, I had to refactor this page to use window.reactProps.

I can see the meta tags in my local browser, and I'll also retest with Discord after it's pushed to Dev.

### DIFF
--- a/routes/tools_routes.js
+++ b/routes/tools_routes.js
@@ -271,7 +271,7 @@ router.get('/card/:id', async (req, res) => {
       metadata: generateMeta(
         `${card.name} - Cube Cobra`,
         `Analytics for ${card.name} on CubeCobra`,
-        card.art_crop,
+        card.image_normal,
         `https://cubecobra.com/card/${req.params.id}`,
       ),
     });
--- a/routes/tools_routes.js
+++ b/routes/tools_routes.js
@@ -7,6 +7,7 @@ const cardutil = require('../dist/utils/Card.js');
 const { addPrices, GetPrices } = require('../serverjs/prices');
 const Filter = require('../dist/utils/Filter');
 const { getElo } = require('../serverjs/cubefn.js');
+const generateMeta = require('../serverjs/meta.js');
 
 const CardRating = require('../models/cardrating');
 const Card = require('../models/card');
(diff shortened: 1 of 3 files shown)


views/tool/cardpage.pug
```diff
@@ -2,12 +2,12 @@ extends ../layout
...
-  input(type='hidden', id='prices', value=JSON.stringify(prices))
-  input(type='hidden', id='related', value=JSON.stringify(related))
-  input(type='hidden', id='cubes', value=JSON.stringify(cubes))
+
   #react-root
+
+block scripts
   include ../react
+  script(type='text/javascript').
+    window.reactProps = !{reactProps};
```
gabyreilly: I copied this from other pages, but I'm not sure what this syntax means. If someone could explain it to me, I'd love to learn :)

src/cardpage.js
```diff
@@ -70,9 +70,9 @@ class CardPage extends Component {
             </Col>
             <Col className="breakdown" xs="12" sm="8">
               <p>
-                Played in
+                Played in &nbsp;
```
joshfinnie: I am not sure on the best practice here, but I think instead of adding `&nbsp;` we'd want `{' '}`... I am up to discuss this though as like I said, I am not sure the actual best practices here.

gabyreilly: Thanks for the idea! I hadn't seen that before.  

I added these in, and the `prettier` plugin complained about the first one, so i took its suggestion and reformatted a bit.

Thanks again!

joshfinnie: yea, it's a weird one for sure. This gives a little bit more information as to my thought process...

https://stackoverflow.com/questions/40264084/best-practice-when-adding-whitespace-in-jsx

dekkerglen: I definitely prefer using {' '} for white space in JSX

routes/tools_routes.js
```diff
@@ -257,12 +258,22 @@ router.get('/card/:id', async (req, res) => {
...
       cubes,
       related: data.cubedWith.map((name) => carddb.getMostReasonable(name[0])),
+    };
+    return res.render('tool/cardpage', {
+      reactProps: serialize(reactProps),
+      title: `${card.name}`,
+      metadata: generateMeta(
+        `${card.name} - Cube Cobra`,
+        `Analytics for ${card.name} on CubeCobra`,
+        card.art_crop,
```
dekkerglen: I think I like using the full card art here

gabyreilly: Done!  I flipped back and forth a few times.  Let's see how it looks on the dev server -- i wasn't sure if the dimensions would be right.

src/cardpage.js
gabyreilly: Done, thanks!  (I forgot to push changes earlier, lol)

PR #1121, opened 2020-02-19
Buff fetches and colorless ratings

dekkerglen:


ruler501:
@andymangold I believe colorless lands haven't changed, fetches went from 1.2 to 1.3. Colorless nonlands went from 1.1 to 1.2.

dekkerglen:
The comment is not correct, which is why I removed it

PR #1122, opened 2020-02-20
[WIP] New Parsing Implementation

ruler501:
This is for paving the way to autocomplete and hopefully better extensibility and easier maintainability.


mawills:
Needs test coverage

src/parsing/filterVisitor.js
```diff
@@ -0,0 +1,92 @@
...
+                return result;
+              };
+            }),
+          );
+        }
+      }
+      return (card) => filters.every((filter) => filter(card));
+    }
+
+    // eslint-disable-next-line class-methods-use-this
```
mawills: why are we ignoring the eslint here?

ruler501: Because the visit methods need to be class methods to be found correctly, but the class is stateless and they don't recurse so they never use `this` which is an eslint error.

ruler501:
> Needs test coverage

The goal would be to have this as a drop in replacement for the current system in which case that should provide us with a good start for testing.

mawills:
The new functions added in parsingHelpers should all be pretty straight-forward to add unit tests for, at the very least.

PR #1127, opened 2020-02-20
Alternate way to tweak draftbots.

phulin:
Lowers boost for on-color cards and hard-codes the fetchland color override.

### DIFF
--- a/src/utils/Draft.js
+++ b/src/utils/Draft.js
@@ -31,41 +31,39 @@ function arrangePicks(picks) {
   draft.picks[0] = [...picks];
 }
 
-const fetchLands = [
-  'Arid Mesa',
-  'Bloodstained Mire',
-  'Flooded Strand',
-  'Marsh Flats',
-  'Misty Rainforest',
-  'Polluted Delta',
-  'Scalding Tarn',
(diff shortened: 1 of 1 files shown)


src/utils/Draft.js
```diff
@@ -31,41 +31,40 @@ function arrangePicks(picks) {
...
+  'Arid Mesa': ['W', 'R'],
+  'Bloodstained Mire': ['B', 'R'],
+  'Flooded Strand': ['W', 'U'],
+  'Marsh Flats': ['W', 'B'],
+  'Misty Rainforest': ['U', 'G'],
+  'Polluted Delta': ['U', 'B'],
+  'Scalding Tarn': ['U', 'R'],
+  'Verdant Catacombs': ['B', 'G'],
+  'Windswept Heath': ['W', 'G'],
+  'Wooded Foothills': ['R', 'G'],
```
ruler501: Should probably include Prismatic Vista here as colorless so it shows up as a fetch.

dekkerglen: Can you make a separate issue for this? Also include the ELD rainbow fetch

PR #1129, opened 2020-02-20
1092 upload page state

### LINKED ISSUE #1092: Confirm Bulk Import not showing login state
This was reported:

This is not possible due to this screen only being accessible when logged in. The user reported that after finishing the import, the user was logged in correctly.

image

gabyreilly:
fixes #1092 

This work is related to the page that appears when you attempt to import card names into a cube, but a few card names are not found.

In the reported bug, there are two issues which I initially thought were related:

image

1: The user appears to be logged out on this page.  The user was logged in before and after this page
2: The various tabs on the cube navigation page went to /undefined instead of the actual cube ID.

Here are the fixes:

Issue 1: 
In most pages on CC, POSTS are immediately redirected to a GET.  This is not the case for the import error page -- the HTML returned to the user is actually the response to the POST.

In app.js, we already had code to make the user's login info "sticky" across GET requests.  I added the similar block below, to make the login sticky for POST as well.

Issue 2:
The cube_layout.pug layout controls the cube navigation, and the layout was not receiving cube_id as part of the react props.  I initially passed in cube_id as a sibling of the cube object, but I noticed that other pages actually used get_cube_id(cube) instead of passing both, so I updated the layout and tested the other cube pages.

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -1051,10 +1051,9 @@ router.post('/importcubetutor/:id', ensureAuth, body('cubeid').toInt(), flashVal
         blogpost: blogpost.toObject(),
       };
       return res.render('cube/bulk_upload', {
-        reactHTML:
-          NODE_ENV === 'production'
-            ? await ReactDOMServer.renderToString(React.createElement(BulkUploadPage, reactProps))
-            : undefined,
+        reactHTML: BulkUploadPage
+          ? await ReactDOMServer.renderToString(React.createElement(BulkUploadPage, reactProps))
+          : undefined,
         reactProps: serialize(reactProps),
         cube,
         cubeID: req.params.id,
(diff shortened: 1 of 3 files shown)


routes/cube_routes.js
```diff
@@ -1051,10 +1051,9 @@ router.post('/importcubetutor/:id', ensureAuth, body('cubeid').toInt(), flashVal
         blogpost: blogpost.toObject(),
       };
       return res.render('cube/bulk_upload', {
-        reactHTML:
-          NODE_ENV === 'production'
```
gabyreilly: These conditional seemed redundant since they duplicate this block:

https://github.com/dekkerglen/CubeCobra/blob/2728c61e0113f0a89bde8cdb2b240eeb77b53f99/routes/cube_routes.js#L64-L69

And they were confusing while i was debugging, so i cleaned them up!

Issue #1133, opened 2020-02-21
Should we allow CMC to be removed from cards?


gabyreilly:
Follow up from issue https://github.com/dekkerglen/CubeCobra/issues/1123

In the edit modal and the bulk edit screen, I can delete the CMC of a card.  This saves it as `null` in the db.

Is this something we should allow? Is it something that is desired by players? is `null` useful in some way?

If not, we should ensure the inputs are stricter and avoid these kind of issues in the future.

dekkerglen:
We should definitely validate these inputs - and also have more strict schemas for the cube. #974 shows a way to have strict schemas for card objects that we could use (once that is merged). We can force those fields to required so that null won't be a valid value to save.

PR #1134, opened 2020-02-21
1123 draft should allow cmc null

### LINKED ISSUE #1123: Cards that appear more than once in playtest mode can't be picked.
**Describe the bug**
In playtest mode, multiples of the same card appearing in a pack result in neither being pickable due to a JS error:
`Util.js:83 Uncaught (in promise) TypeError: Cannot read property 'indexOf' of null
    at d (Util.js:83)
    at CubeDraftPage.js:170
    at Object.Bi (react-dom.production.min.js:175)
    at Ai (react-dom.production.min.js:14)
    at Di (react-dom.production.min.js:14)
    at wf (react-dom.production.min.js:14)
    at Ei (react-dom.production.min.js:176)
    at zd (react-dom.production.min.js:15)
    at uc (react-dom.production.min.js:15)
    at aj (react-dom.production.min.js:41)`

**To Reproduce**
1. Go to [https://cubecobra.com/cube/draft/5e4e88db18368406b4324502](https://cubecobra.com/cube/draft/5e4e88db18368406b4324502)
2. Pick any card (I picked Grusilda) followed by either copy of Watermarket.

**Expected behavior**
Watermarket should be picked.

 - OS: W10
 - Chrome 80.0.3987.106 (64-bit)

gabyreilly:
fixes #1123 but we might want some follow up work, i opened https://github.com/dekkerglen/CubeCobra/issues/1133 for later discussion.

In the reported issue, both cards that were undraftable had CMC `null` instead of 0 as expected.

On my local database, the cards initially had CMC zero, but I could reproduce this issue using bulk edit:

image

This quick-fix allows CMC `null` cards to be drafted, and places them in the 0 CMC column.

### DIFF
--- a/src/utils/Draft.js
+++ b/src/utils/Draft.js
@@ -79,7 +79,7 @@ function botPicks() {
     const ratedPicks = [];
     const unratedPicks = [];
     for (let cardIndex = 0; cardIndex < botPack.length; cardIndex++) {
-      if (draft.ratings[botPack[cardIndex].details.name]) {
+      if (draft.ratings && draft.ratings[botPack[cardIndex].details.name]) {
         ratedPicks.push(cardIndex);
       } else {
         unratedPicks.push(cardIndex);
--- a/src/utils/Util.js
+++ b/src/utils/Util.js
@@ -79,6 +79,10 @@ export function alphaCompare(a, b) {
 
 export function cmcColumn(card) {
   let cmc = Object.prototype.hasOwnProperty.call(card, 'cmc') ? card.cmc : card.details.cmc;
+  // double equals also handles undefined
+  if (cmc == null) {
+    cmc = 0;
+  }
   if (!Number.isFinite(cmc)) {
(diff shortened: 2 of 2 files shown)


src/utils/Draft.js
```diff
@@ -79,7 +79,7 @@ function botPicks() {
     const ratedPicks = [];
     const unratedPicks = [];
     for (let cardIndex = 0; cardIndex < botPack.length; cardIndex++) {
-      if (draft.ratings[botPack[cardIndex].details.name]) {
+      if (draft.ratings && draft.ratings[botPack[cardIndex].details.name]) {
```
gabyreilly: This isn't related to the filed issue, but it caused some problems for me locally because the draft didn't have ratings.

dekkerglen: This happens b/c you don't have enough draft data in your local DB - and mongoose just saves the empty dict as null. This change is fine, just wanted to share why this is happening

PR #1138, opened 2020-02-21
clarify set up steps

mawills:
While troubleshooting issues with my local dev environment, I had a few friends try setting up CubeCobra on their own machines (linux and mac), and these were issues that all of them ran into.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ You will need start a MongoDB background process. Refer to the official document
 
 VSCode (not required): https://code.visualstudio.com/
 
-### Clone Project
+### Initial Set Up
 
 Clone the project into a folder of your choice. 
 
@@ -39,6 +39,11 @@ cd CubeCobra/..
 ln -s CubeCobra/cubecobrasecrets.example cubecobrasecrets
 ```
 
+In a separate terminal window, make sure your MongoDB database instance is up and running
+```sh
+sudo mongod
+```
+If you get an error such as "Data directory /data/db not found", you can fix this simply by going to your root directory `cd /` and creating those directories `sudo mkdir -p data/db`
 
 Then, run the following commands in the root of the cloned repository:
 
@@ -69,6 +74,14 @@ Then you can start the program like so:
 You can now open up a browser and connect to the app through: http://localhost:8080. Despite the fact that node says it is running on port 5000, you should use port 8080 to connect.
 
 Nodemon will restart the application anytime there is a change to a source file.
+
+After these initial set up steps, you can start up your development environment by opening two terminals and running one of the following commands in each:
+```sh
+sudo mongod
+```
+```sh
+npm start
+```
 
 ### Running tests
 


ruler501:
I think giving instructions or a link to setting up mongo with systemd would be preferable here. You don't really want to be manually starting and stopping it yourself.

mawills:
> I think giving instructions or a link to setting up mongo with systemd would be preferable here. You don't really want to be manually starting and stopping it yourself.

You think instructions to start up the database on boot are preferable? That sounds a little intrusive to me.

ruler501:
You can use systemd to start and stop only, you don't have to enable the service.

mawills:
Aren't you manually starting and stopping it still in that case? Running `mongod` in another terminal window is OS agnostic and is similar enough to the `npm start` the dev is already running.

README.md
```diff
@@ -39,6 +39,11 @@ cd CubeCobra/..
 ln -s CubeCobra/cubecobrasecrets.example cubecobrasecrets
 ```
 
+In a separate terminal window, make sure your MongoDB database instance is up and running
```
RKelln: Convert MongoDB to a link or add a link to install instructions?
https://docs.mongodb.com/manual/installation/

RKelln:
I'm no sysadmin but I agree with ruler501, feels awkward to have daemon run manually in another terminal. I like the idea of OS-agnostic instructions, but perhaps a "sidebar" note for Linux/Ubuntu systems with systemd style instructions?

ruler501:
Overall I feel these are questions we shouldn't really need to be answering. There should be some setup instructions we can link to instead of having to write our own for mongo.

mawills:
Having an instance of mongodb running is a requirement for our development set up to function at all. I had 3 different people (some linux, some mac) try cloning the repo and get set up. _All_ of them got stuck on this step because they didn't realize that. If we have requirements, they should be stated imo.

ruler501:
I think we can state that Mongo is a requirement but I don't think we
should be writing install instructions for Mongo, there must be some
official instructions we can link to so we aren't responsible for
maintaining them and it's clear where the best place to go with questions
on it is.

mawills:
I think you might be misunderstanding. `sudo mongod` doesn't install mongo, it's the command to start up an instance of mongo after you've already installed it.

README.md
mawills: This isn't an instruction to install mongoDB, it's an instruction to start an instance of it. Running seed.js will fail unless there is an instance of mongod running (since that means the database isn't up)

devonrichardsz:
Still there should be official docs for how to run it. We should be able to just say something like `You need to install and have MongoDB running. Documentation on MongoDB can be found at...`

mawills:
We already do say that, but are commonly missing it and getting stuck because seed.js fails, hence making it explicit and placing it here.

RKelln:
Explicit is better. mawills (and myself) have seen people fail at getting things installed and running. I suggest adding a link to this: https://docs.mongodb.com/manual/administration/install-community/   and also a quick summary of how to run it after install for each OS.

dekkerglen:
I have a readme update here: https://github.com/dekkerglen/CubeCobra/pull/1174/files#diff-04c6e90faac2675aa89e2176d2eec7d8

Does this satisfy your concerns? There are notes in `Prerequisites` for setting up mongo.

mawills:
> I have a readme update here: https://github.com/dekkerglen/CubeCobra/pull/1174/files#diff-04c6e90faac2675aa89e2176d2eec7d8
> 
> Does this satisfy your concerns? There are notes in `Prerequisites` for setting up mongo.

This is better, but I think it would be best to add a comment at the beginning of the code block after "Then, run the following commands in the root of the cloned repository:" so that it was something like
```sh
# Ensure you have an instance MongoDB running and then these scripts
npm install
npm install nodemon -g
npm install rollup -g
npm run setup               # This will bundle modules and download Scryfall assets.
node seed.js                # This will create and seed the mongo database specified in cubecobrasecrets.

ruler501:
Minor nitpick but I think `mongo daemon` would be better than `mongo background process`.

Vurolock:
since the seeding scripts haven't been updated with the new format, can we change `node seed.js` to `node seed.js && node ./one_shot_scripts/update_deck_format.js && node ./one_shot_scripts/update_draft_format.js`. Otherwise there are going to be app breaking issues.

ruler501:
An additional problem I found is mongo refusing IPv6 by default which for some reason became the default mongoose used on my system. I have a PR up to fix the default connection string. PR for changing the default connection string is #1215.

ruler501:
Something else I thought of today is this is a good place to plug dockerizing since that would help reduce any pain in setup since the only software you'd need to install on your system would be docker.  The `dockerize` branch on my fork has setup for running it in docker which I semi-routinely check against master to ensure it still works by updating my test server https://cubecobra.ruler501.com which runs the docker container in a Kubernetes cluster.

PR #1148, opened 2020-02-25
[WIP] Sort .cvs export

bramleyjl:
Changes the .csv export file to sort by Color -> CMC -> Alphabetical


src/utils/Sort.js
```diff
@@ -727,3 +727,16 @@ export function countGroup(group) {
   }
   return group.length;
 }
+
+export function downloadSort(cards) {
```
dekkerglen: Can you rename this to sortForCSVDownload? downloadSort sounds like it should download a sort.

src/utils/Sort.js
```diff
@@ -727,3 +727,16 @@ export function countGroup(group) {
...
+  var exportCards = [];
+  cards = sortDeep(cards, 'Color', 'CMC');
+  cards.forEach((firstGroup) => {
```
dekkerglen: can you use for(const .. of ..) for this instead of .forEach()? This is a style change we've recently been moving towards

dekkerglen:
Passing the current selected sorts world be very nice

bramleyjl:
Got a bit stuck on how I'm supposed to set the primary and secondary sort values in the proper scope from my callback function; there doesn't seem to be a state to `SortCollapse` that I can set?

src/utils/Sort.js
```diff
@@ -727,3 +727,16 @@ export function countGroup(group) {
   }
   return group.length;
 }
+
+export function sortForCSVDownload(cards) {
+  var exportCards = [];
+  cards = sortDeep(cards, 'Color', 'CMC');
+  for (const firstGroup of cards) {
+    for (const secondGroup of firstGroup[1]) {
```
dekkerglen: This doesn't look right... we are we doing `firstGroup[1]`?

bramleyjl: Each group produced by `sortDeep` is an array where the 0 index is the category (like `White` or `3`) and the 1 index is a subarray with either second groups or the cards themselves.

dekkerglen: Right, because that was used to write the headers. Looks good to me then

PR #1149, opened 2020-02-26
Pick by Pick Breakdown

### LINKED ISSUE #659: Pick by pick breakdown
As a user, I would like to be able to view the pick by pick breakdown of any given draft. For each pick, show the pack, and what the drafter picker from that pack, and maybe the draft rating for each other card in the pack?

We could use the carousel bootstrap element to good use here.

dekkerglen:
Fixes #659


src/components/DeckCard.js
```diff
@@ -160,12 +169,36 @@ DeckCard.propTypes = {
...
+        name: PropTypes.string.isRequired,
+      }),
+    ).isRequired,
+    cube: PropTypes.string.isRequired,
+    comments: PropTypes.arrayOf(PropTypes.object).isRequired,
+  }).isRequired,
+  seatIndex: PropTypes.number.isRequired,
+};
+
+DeckCard.defaultProps = {
```
joshfinnie: since `bot` is no longer required, would we not need a `defaultProp` for it?

src/components/DecksPickBreakdown.js
```diff
@@ -0,0 +1,147 @@
...
+    this.click = this.click.bind(this);
+    this.setPickIndex = this.setPickIndex.bind(this);
+
+    this.state = {
+      index: 0,
+    };
+  }
+
+  setPickIndex(index) {
+    console.log(`Setting index to ${index}`);
```
joshfinnie: do we want/need this `console.log`?

src/pages/CubeDeckPage.js
```diff
@@ -36,12 +40,21 @@ const CubeDeckPage = ({ cube, deck, canEdit, userid }) => {
             </Label>
             <Input type="select" id="viewSelect" value={seatIndex} onChange={handleChangeSeat}>
               {deck.seats.map((seat, index) => (
-                <option key={seat.userid || seat.bot} value={index}>
+                <option key={/* eslint-disable-line react/no-array-index-key */ index} value={index}>
```
joshfinnie: not sure, but could this be solved by something like this:

```
<option key={"seat"+index} value={index}>
```

src/components/DeckCard.js
dekkerglen: It's not that it's not required, it's just that it can be null (or absent). We don't want a defaultProp for it because being null means its a human

src/pages/CubeDeckPage.js
dekkerglen: We use this current pattern elsewhere as well. I personally think we should remove this rule. What is the reasoning behind it exactly?

joshfinnie: react component keys should be unique to the item itself for dom change reasons. if you use an index, it's not unique to the object itself but its location in a list.

dekkerglen: Makes sense, revised

joshfinnie: this is a bit more information that I found...

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

PR #1152, opened 2020-02-27
Fix the case where CMC is undefined preventing making a deck from draft

ruler501:
I believe this happens with Vanguards specifically. Also happens with Art cards which act different from other printings since they don't have proper details.


dekkerglen:
I believe a better way to fix this is during the card import script

PR #1153, opened 2020-02-27
moving the time component to react's timeago

joshfinnie:
This replaces the `AgeText` component and replaces it with `TimeAgo` [found here](https://www.npmjs.com/package/react-timeago)


dekkerglen:
RIP AgeText

PR #1155, opened 2020-02-28
Add an endpoint to provide CardRatings data and a script to consume it

ruler501:
Will enable local testing of things that depend on elo values.

### DIFF
--- a/routes/tools_routes.js
+++ b/routes/tools_routes.js
@@ -286,7 +286,7 @@ router.get('/api/cardratings', async (req, res) => {
   try {
     res.status(200).send({
       success: 'true',
-      data: await CardRating.find({}).lean(),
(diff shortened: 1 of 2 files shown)


routes/tools_routes.js
```diff
@@ -282,4 +282,18 @@ router.get('/card/:id', async (req, res) => {
   }
 });
 
+router.get('/api/cardratings', async (req, res) => {
+  try {
+    res.status(200).send({
+      success: 'true',
+      data: (await CardRating.find({}).lean()).map(({ name, elo, value, picks }) => ({ name, elo, value, picks })),
```
dekkerglen: This is not the correct way to do this.
Please use `.find({query}, 'name', ....`

PR #1156, opened 2020-02-28
Introduce new way to assign colors to bots

ruler501:
Calculates a rating for different colors being distributed amongst the seats, tries 1000 possibilities and chooses the best from among them. Appears to be very fast and number of iterations can be tweaked. Rating is based on trying to maximize the number of cards where there is a seat that has the colors to play it, maximizing the "available elo" which is the sum of the card elo, times the number of colors in its identity, over the number of colors for the seat, over the number of seats trying to draft that card, and maximizing the minimum "available elo" over all seats.

This should go beyond just making the bots more aware of what colors are available, but also let them key in on what colors seem most powerful based on elo. It should also help reduce the number of multicolor cards that table since it'll be more likely someone will be in the colors to play them.

This does allow bots to go beyond 2 colors, potentially all the way to 5 colors, given that there are cards with exactly those colors in their identity in the packs and even then they're disincentivized towards it since the value of their cards goes down as the number of colors they play increases.

This is difficult to test without elo data, see #1155 for that, but it does seem to have bots picking colors better so colors don't get ignored. I'll have a demo up on https://cubecobra.ruler501.com/cube/playtest/1/ if you want to try it out.


dekkerglen:
Conflicts - I'm also not sold on this solution. I think the direction we should move is dynamic color selection as draft progresses, based on picks and cards seen.

phulin:
I agree with Gwen. If we're going to be adding 800 lines of code here, let's have them be for dynamic selection of bot colors.

ruler501:
Thinking more about this and I still think this is a good first step. I'm fine working on dynamic selection next, but I think at the very least this is a good way to seed that and it does seem to produce better results in my testing. I've simplified the PR somewhat to try and make it more focused on just the relevant changes. Note that a good portion of the new lines of code are from package-lock.json and comments. There's only 200 or so lines of meaningfully changed code I'm pretty sure.

ruler501:
Closing since I can admit dynamic choice would be better and I found a good way to do it that doesn't depend on this, see #1239

PR #1169, opened 2020-03-06
Have prettier run on root level .js files as well.

ruler501:
If there are some we don't want it running we should add them to .prettierignore.

We have eslint cover some root level files like app.js and since it runs prettier checks those for correctness. This has the prettier fix, `npm run prettier`, and check, `npm run lint`, commands check every js file at the root as well as everything else it already checked unless they are in the `.prettierignore` file


dekkerglen:
Conflicts

dekkerglen:
Conflicts

PR #1170, opened 2020-03-07
Prevent card.cmc from being null

ruler501:
You will need to run `one_shot_scripts/validate_cmcs.js` to fix existing cubes
and `force_update.js` to fix the existing carddb.

### DIFF
--- a/one_shot_scripts/validate_cmcs.js
+++ b/one_shot_scripts/validate_cmcs.js
@@ -1,0 +1,28 @@
+const mongoose = require('mongoose');
+
+const mongosecrets = require('../../cubecobrasecrets/mongodb');
+
+const Cube = require('../models/cube');
+
+(async () => {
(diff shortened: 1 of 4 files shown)


one_shot_scripts/validate_cmcs.js
```diff
@@ -0,0 +1,28 @@
...
+    const totalCubes = await Cube.count();
+    for (let start = 0; start < totalCubes; start += 100) {
+      // eslint-disable-next-line no-await-in-loop
+      const cubes = await Cube.find()
+        .skip(start)
+        .limit(100);
+      const collected = [];
+      for (const cube of cubes) {
+        for (const card of cube.cards) {
+          card.cmc = card.cmc || 0;
```
dekkerglen: No! This will cause cascading mongoose cloning. We should use `.lean()`, and then you can use `await Cube.updateOne({_id:cube._id}, cube)`

PR #1180, opened 2020-03-09
Add deck downloading

### LINKED ISSUE #618: Export for draft decks
As a user, I'd like to be able to export my draft decks in various formats.

dekkerglen:
Fixes #618

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -1560,6 +1560,261 @@ router.get('/download/plaintext/:id', async (req, res) => {
   }
 });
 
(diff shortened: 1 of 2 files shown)


joshfinnie:
@dekkerglen this looks good from a code perspective. However there seems to be a lot of duplicated code throughout these routes. Not sure if you want to see if there's some way of abstracting this out or if that's just too much work in the end.

bramleyjl:
Agree with @joshfinnie, in particular the naming for maindecks/sideboards has a lot of repetition, could that be abstracted out to a single function with a platform parameter to allow for the small differences in naming between platforms?

PR #1186, opened 2020-03-11
Explore and Search Updates

dekkerglen:
This PR reworks the explore and search pages, bringing them into react:

Explore:
image
image

Search:
image
image
image

Cubes with card has been moved away from card analytics, instead there is a button linking to the correct search query:
image

This PR also improves performance with several queries, adds a new cube carousel component but was not used.


src/components/CubeSearchNavBar.js
```diff
@@ -0,0 +1,245 @@
...
+      case 'include':
+        setInclude(value);
+        break;
+      default:
+        break;
+    }
+  };
+
+  const submit = () => {
+    console.log(name, owner, decks, cards);
```
joshfinnie: is this needed?

dekkerglen: darn, I'm surprised that wasn't caught by linter though

joshfinnie: we globally allow `console` since that was the logging system at the start:
https://github.com/dekkerglen/CubeCobra/blob/master/.eslintrc.yml#L18

PR #1189, opened 2020-03-12
Add basic sealed

dekkerglen:
Adds a basic sealed implementation

Sealed card UI: (also note the reordered elements here)
image

After firing, the user will be brought to the deckbuilder with the following autosort:
image


routes/cube_routes.js
```diff
@@ -1556,6 +1556,119 @@ router.get('/download/plaintext/:id', async (req, res) => {
...
+
+router.post('/startsealed/:id', async (req, res) => {
+  try {
+    const user = await User.findById(req.user);
+
+    if (!user) {
+      req.flash('danger', 'Please Login to build a sealed deck.');
+      return res.redirect(`cube/playtest/${req.params.id}`);
+    }
+    const params = {
```
ruler501: Couldn't you add `body('packs').toInt(), body('cards').toInt()` to the route definition? With any required validations like `{ min: 1}` added there?

routes/cube_routes.js
```diff
@@ -1556,6 +1556,119 @@ router.get('/download/plaintext/:id', async (req, res) => {
...
+    if (!cube) {
+      req.flash('danger', 'Cube not found');
+      return res.status(404).render('misc/404', {});
+    }
+
+    if (cube.cards.length < numCards) {
+      throw new Error('Could not create sealed pool: not enough cards');
+    }
+
+    const source = shuffle(cube.cards).slice(0, numCards);
```
ruler501: Why can't you use the pack generation code we use for drafts?

dekkerglen: This is just a basic implementation that can be extended later

routes/cube_routes.js
dekkerglen: Good call

routes/cube_routes.js
```diff
@@ -1556,6 +1556,123 @@ router.get('/download/plaintext/:id', async (req, res) => {
...
+    if (!cube.decks) {
+      cube.decks = [];
+    }
+
+    cube.decks.push(deck._id);
+    if (!cube.numDecks) {
+      cube.numDecks = 0;
+    }
+
+    cube.decks.push(deck._id);
```
phulin: This line is duplicated. I also think you need to move it below to after the deck is saved; it won't have an `_id` until then.

routes/cube_routes.js
```diff
@@ -1556,6 +1556,123 @@ router.get('/download/plaintext/:id', async (req, res) => {
...
+  return a;
+}
+
+router.post('/startsealed/:id', body('packs').toInt({ min: 1, max: 16 }), body('cards').toInt(), async (req, res) => {
+  try {
+    const user = await User.findById(req.user);
+
+    if (!user) {
+      req.flash('danger', 'Please Login to build a sealed deck.');
+      return res.redirect(`cube/playtest/${req.params.id}`);
```
phulin:
```suggestion
      return res.redirect(`/cube/playtest/${req.params.id}`);
```

routes/cube_routes.js
```diff
@@ -1556,6 +1556,123 @@ router.get('/download/plaintext/:id', async (req, res) => {
...
+    }
+    const params = {
```
phulin: Why is this an object? I assume just because we're copying-and-pasting but not necessary.
```suggestion
    const packs = parseInt(req.body.packs, 10);
    const cards = parseInt(req.body.cards, 10);
```

routes/cube_routes.js
```diff
@@ -1556,6 +1556,123 @@ router.get('/download/plaintext/:id', async (req, res) => {
...
+      packs: parseInt(req.body.packs, 10),
+      cards: parseInt(req.body.cards, 10),
+    };
+
+    const numCards = params.packs * params.cards;
```
phulin:
```suggestion
    const numCards = packs * cards;
```

PR #1199, opened 2020-03-13
Rewrite autocard code without jquery.

phulin:
This removes the last of the direct usages. Next step is to remove all the native bootstrap js—anything using `data-target`—and then we can remove jquery/bootstrap js which will speed up loads.


phulin:
Integration failure appears to be due to a Travis bug, lint + test pass on my local.

PR #1200, opened 2020-03-13
Fix autocard issues.

phulin:
This PR doesn't change the size, something we're discussing on #1192. It does fix several issues with the autocard code: overlapping flip cards and wrong aspect ratio, in particular.


dekkerglen:
Conflicts with jquery change

phulin:
Unneeded with jquery change.

PR #1204, opened 2020-03-13
Fix memory error on Top Cards page.

### LINKED ISSUE #1197: Top cards page is broken
The top cards page (https://cubecobra.com/tool/topcards) only displays a "Internal Server Error"

phulin:
Fixes #1197.

Also adds a longer-term fix that should make Top Cards faster by adding indexed cubesLength field.


dekkerglen:
Needs prettier

PR #1207, opened 2020-03-13
strip details on deck submit

dekkerglen:
Prettier hit a bunch of random files too for some reason

### DIFF
--- a/one_shot_scripts/update_deck_format.js
+++ b/one_shot_scripts/update_deck_format.js
@@ -193,11 +193,24 @@ async function buildDeck(cards, bot) {
   }
 }
 
-async function update(deck) {
-  if (deck.seats && deck.seats.length > 0) {
-    return Deck.updateOne({ _id: deck._id }, deck);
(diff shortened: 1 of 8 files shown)


one_shot_scripts/update_deck_format.js
```diff
@@ -193,11 +193,24 @@ async function buildDeck(cards, bot) {
...
+function convertDeckCard(card) {
+  if (Array.isArray(card)) {
+    return card.map(convertDeckCard);
+  }
+  if (card.cardID) {
+    return card;
   }
+  return {
+    tags: [],
+    colors: card.details.colors,
```
phulin:
```suggestion
    colors: card.details.color_identity,
```

phulin: idk if this matters but it's correct

dekkerglen: for this case, it's fine -  these colors aren't used for any analytics or anything

PR #1215, opened 2020-03-14
Remove the need to install global packages.

ruler501:
Reduce the global impact of setting up and running the project on systems by moving rollup and nodemon into dev dependencies instead of asking devs to install them globally.


phulin:
We're no longer using rollup so it's not necessary to have it as a dependency. Can you check to make sure the rest of the instructions work with nodemon not in the global path?

ruler501:
Everything I've tested has worked fine. Running one shot scripts, seed, setup, running the project, and making changes to src or routes did live reload.
This actually fixed a problem I was having where a globally installed nodemon was having trouble finding node due to not including `.exe` in its path.

PR #1216, opened 2020-03-14
Default to connecting over 127.0.0.1 to force IPv4.

ruler501:
MongoDB by default refuses IPv6 queries which mongoose will use by default on some systems.


dekkerglen:
Important to note: this only changes the default value for the secrets, this value should be changed according to the environment

ruler501:
Yes. The goal was to reduce potential issues people might encounter when first starting to build the project.

PR #1217, opened 2020-03-14
Introduce generic way to access card properties

ruler501:
This fixes a variety of small bugs where cube overrides weren't respected and makes an easier way to future proof potential future additional overrides without having to find all usages.

Goes through filters and sorts and changes them to use the new function.

The filter tests were failing since they ignored overrides and this fixes that and updates them to be able to pass ESLint. Having ESLint check the file is left for another PR.

Depends on #1218.


ruler501:
@phulin the idea was to have a function we could use for almost every property and if we add overrides for a new function we would only need to update one place to get everywhere to use the new functionality.

phulin:
Yeah, I think I get the idea - but there's always a balance between extensibility and simplicity in code, and I think I'd prefer a simpler solution here.

src/components/Maybeboard.js
```diff
@@ -134,6 +134,7 @@ MaybeboardListItem.defaultProps = {
 };
 
 const Maybeboard = ({ filter, ...props }) => {
+  console.log(filter);
```
RKelln: debugging

RKelln:
I'd like to see this cleaned up, but agree the function feels awkward, although better than what we have now. This seems like a good place to use getters and setters rather than a function? We may want to think about what is most clear naming-wise. I'd suggest:

card.property = overridden property
card.original.property = card's original property value  (not sure if original is the right word? "prime"?)
card.override.property = overridden property or NULL if not overriden?

It would be nice if these could be dynamically created, but I don't think that's possible, but likely worth the effort of defining the getters and setters.

phulin:
The cards are currently just plain javascript objects everywhere they're
used, so no way to define methods on the objects - that would definitely be
optimal if it were possible. I don't have any good ideas as to how to give
them a prototype but definitely open to suggestions.

ruler501:
My first thought was getters and setters but that would require a very large overhaul to convert the cards everywhere into class objects.
I'm very open to suggestions, but thought this was a good way to make some progress and get the conversation started on a good future proofed way to handle accessing potentially overridden properties.

RKelln:
I haven't used get/set in js before, but seems like you can define get and set on objects:
```js
const o = {a: 0};
Object.defineProperty(o, 'b', { get: function() { return this.a + 1; } });
```
So could we have a function that sets up a card when we load into carddict? (Or a loop through all the cards after loading json, etc).

Would we need an equivalent when loading a card from mongoose, or is that just getting an id that we go to the carddict for? I'm not quite sure how the mongoose and carddict stuff connect.

In previous projects I usually went json -> classed object instance. But something like Go makes the trivially easy. Maybe there is an equivalent in js? Ya seems like you can use a `reviver` function in `JSON.parse` but it looks like a mess.

Hmm, ya, not sure best next step. Something simple that would need minimal changes when we do it correctly in the future, but not sure that exists.

ruler501:
The problem is doing that on the client side. We currently serialize the cards to json and embed it in the web page, we'd have to change that as well as all the api call points to add the getters and setters there

Issue #1230, opened 2020-03-16
Blog Post Confirm delete accept enter


dekkerglen:
As a user, It'd be awesome if when the "are you sure you want to delete?" dialogue pops up, I could hit enter or spacebar or something to confirm deletion rather than clicking the button.

PR #1238, opened 2020-03-29
Add Blog Post Confirm Delete via Enter keystroke

vhaso:
Issue: #1230

Solution: 
When the delete modal is opened, an event is added, which is triggered by keyup of the Enter key and which clicks the delete button programatically. This event is removed when the delete modal is closed.


phulin:
Could you instead move the delete modal into react? I'd really like to avoid creating new jquery code.

vhaso:
@phulin Yes, I could. There was some JQuery code in the same file so I thought it was an acceptable solution. I don't have that much experience with react, but I'll figure it out.

phulin:
I thought I had a pending PR to get rid of all the remaining jquery code but I don't remember what happened to it

vhaso:
Now there is a different bug: closing the modal does not remove the keybinding.

src/components/BlogDeleteModal.js
```diff
@@ -0,0 +1,68 @@
...
+      } else {
+        window.location.href = '';
+      }
+    });
+  }
+
+  render() {
+    const { isOpen } = this.state;
+    return (
+      <>
```
phulin: You don't need the `<>` fragment tag wrapping the modal here.

src/components/BlogDeleteModal.js
```diff
@@ -0,0 +1,68 @@
...
+import { csrfFetch } from '../util/CSRF';
+
+class BlogDeleteModal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isOpen: Boolean(props.isOpen),
+    };
+
+    this.close = this.close.bind(this);
```
phulin: Instead of having close here, have close be a function on the context menu passed down to the modal. Have the isOpen state be managed by the context menu.

src/components/BlogDeleteModal.js
```diff
@@ -0,0 +1,68 @@
...
+    this.confirm = this.confirm.bind(this);
+    this.keyPress = this.keyPress.bind(this);
+  }
+
+  close() {
+    document.removeEventListener('keyup', this.keyPress);
```
phulin: No need for an event listener on the whole document. You can add a `componentDidUpdate` method that uses a ref to put focus on the delete button when the modal opens. That avoids the cleanup problems.

src/components/BlogContextMenu.js
```diff
@@ -1,26 +1,38 @@
...
       dropdownOpen: !this.state.dropdownOpen,
     });
     updateBlog();
   }
 
+  openDeleteModal() {
+    this.deleteModal.current.setState({
+      isOpen: true,
+    })
+    document.addEventListener('keyup', this.deleteModal.current.keyPress);
```
phulin: See comment below.

phulin:
Can you fix the merge conflicts?

PR #1244, opened 2020-04-02
Improve dynamic color choices for bots

ruler501:
Calculates off of ELO, not count for openness and precomputes most of the values to avoid doing as many duplicate calculations.

### DIFF
--- a/src/utils/Draft.js
+++ b/src/utils/Draft.js
@@ -24,6 +24,7 @@ function init(newDraft) {
   for (const seat of draft.seats) {
     seat.seen = fromEntries(COLOR_COMBINATIONS.map((comb) => [comb.join(''), 0]));
     addSeen(seat.seen, seat.packbacklog[0].slice());
+    seat.picked = fromEntries(COLOR_COMBINATIONS.map((comb) => [comb.join(''), 0]));
   }
   draft.overallPool = fromEntries(COLOR_COMBINATIONS.map((comb) => [comb.join(''), 0]));
   addSeen(draft.overallPool, draft.initial_state.flat(3));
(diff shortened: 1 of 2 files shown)


src/utils/Draft.js
```diff
@@ -281,6 +290,9 @@ async function finish() {
       const colors = botColors(null, null, draft.seats[i].pickorder, cards);
       draft.seats[i].name = `Bot ${i + 1}: ${colors.join(', ')}`;
       draft.seats[i].description = `This deck was drafted by a bot with color preference for ${colors.join('')}.`;
+    } else {
+      const colors = botColors(null, null, draft.seats[i].pickorder, cards);
+      draft.seats[i].name = `${draft.seats[i].name}: ${colors.join(', ')}`;
```
RKelln, line 295: No description?

src/utils/Draft.js
```diff
@@ -1,14 +1,32 @@
...
+import { COLOR_COMBINATIONS } from 'utils/Card';
 
 let draft = null;
 
+const toValue = (elo) => 10 ** (elo / 400);
+
+function addSeen(seen, cards) {
+  for (const card of cards) {
+    const colors = card.colors ?? card.details.colors ?? [];
+    if (colors.length > 0) {
```
RKelln, line 12: No card rating needs to be added for colorless cards or do they have a color?  'C'?

RKelln:
If you have time I'd like to see more comments added to describe some of the more complicated sections of the code. Pretty hard to read though quickly and understand what is going on and helpful for future maintenance / changes.

src/utils/Draft.js
ruler501: Description gets added by the user, we're just modifying it so the name includes our estimated colors of the deck like it does for the bot.

src/utils/Draft.js
ruler501: Colorless gets added to every combination so just reduces variance so they get skipped here.

PR #1254, opened 2020-04-04
Analytics Rework

dekkerglen:
Rework of analytics. Brings all the code into react components.

Adds Pivot Table and Hypergeometric calculator


src/analytics/Asfans.js
```diff
@@ -0,0 +1,207 @@
...
+          return {
+            label: tuple[0],
+            asfan: (tuple[1].length / cube.cards.length) * 15, // 15 cards a pack
+          };
+        }),
+      },
+    ];
+  };
+
+  const calculateCustomAsfans = (draftFormat) => {
```
ruler501, line 92: I don't believe this will properly handle joins where a card falls into multiple different filters. Such as "rarity:Common" and "type:Land" in the case you have common lands in the cube. In that case the chance you pick it is `1/len(Commons) + (1 - 1/len(Commons))/len(Lands)` since it needs to have not been picked yet.

src/analytics/Averages.js
```diff
@@ -0,0 +1,199 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+import { Col, Row, Table, InputGroup, InputGroupAddon, InputGroupText, CustomInput, NavLink } from 'reactstrap';
+
+import { getCmc } from 'utils/Card';
+import { sortIntoGroups, getSorts } from 'utils/Sort';
+import ErrorBoundary from 'components/ErrorBoundary';
+
+const useSortableData = (items, config = null) => {
```
ruler501, line 10: This should be moved to common file it can be imported from instead of duplicated.

src/analytics/Averages.js
```diff
@@ -0,0 +1,199 @@
...
+    if (sortConfig && sortConfig.key === key && sortConfig.direction === 'descending') {
+      direction = 'ascending';
+    }
+    setSortConfig({ key, direction });
+  };
+
+  return { items: sortedItems, requestSort, sortConfig };
+};
+
+const HeaderCell = ({ label, fieldName, sortConfig, requestSort }) => {
```
ruler501, line 40: Same with this, imported instead of duplicated.

src/analytics/Averages.js
```diff
@@ -0,0 +1,199 @@
...
+};
+
+const Averages = ({ cards }) => {
+  const sorts = getSorts();
+  const characteristics = {
+    CMC: getCmc,
+    Power: (card) => parseInt(card.details.power, 10),
+    Toughness: (card) => parseInt(card.details.toughness, 10),
+    Elo: (card) => parseFloat(card.details.elo, 10),
+    Price: (card) => parseFloat(card.details.price),
```
ruler501, line 77: I'd prefer this be `parseFloat(card.finish === 'Foil' ? (card.details.price_foil ?? card.details.price) : (card.details.price : card.details.price_foil), 10)` with an additional item for `'Non-foil Price'`.

src/analytics/Averages.js
```diff
@@ -0,0 +1,199 @@
...
+    'Price Foil': (card) => parseFloat(card.details.price_foil),
```
ruler501, line 78: Stylistically I think `Foil Price` is the more common way to say it.

src/analytics/Averages.js
```diff
@@ -0,0 +1,199 @@
...
+  };
+
+  const [sort, setSort] = useState('Color');
+  const [Characteristic, setCharacteristic] = useState('CMC');
```
ruler501, line 82: lowercase for `Characteristic`

src/analytics/Averages.js
```diff
@@ -0,0 +1,199 @@
...
+
+  const groups = sortIntoGroups(cards, sort);
+
+  const average = (arr) => {
```
ruler501, line 86: I'd like these methods to be in a helper file since they can come up more places like the draft bot code. Also needs a version that works with weighted arrays becuase of asfan. I can give you the code for those if you'd like.

src/analytics/Asfans.js
```diff
@@ -0,0 +1,207 @@
...
+
+const Asfans = ({ cards, cube }) => {
+  const sorts = getSorts();
+
+  const [sort, setSort] = useState('Color');
+  const [formatId, setFormatId] = useState(-1);
+
+  const draftFormats = cube.draft_formats;
+
+  const calculateAsfans = () => {
```
ruler501, line 78: These should be in a shared file so they can be used more places.

src/analytics/PivotTable.js
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import PivotTableUI from 'react-pivottable/PivotTableUI';
+import 'react-pivottable/pivottable.css';
+
+import { Col } from 'reactstrap';
+
+const PivotTable = ({ cards }) => {
+  const data = cards.map((card) => ({
+    CMC: card.cmc || card.details.cmc,
```
ruler501, line 10: This needs to be `card.cmc ?? card.details.cmc` otherwise overrides of 0 don't work.

src/analytics/Tokens.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+  });
+  const data = sortTokens(resultingTokens).map(({ token, related }) => ({
+    card: token,
+    cardDescription: sortCards(dedupeCards(related))
+      .map(({ position }) => `[[${position}]]`)
+      .join('\n\n'),
+  }));
+
+  console.log(data);
```
ruler501, line 55: console statement

src/analytics/Cloud.js
```diff
@@ -0,0 +1,35 @@
...
+import { TagCloud } from 'react-tagcloud';
+import { Col } from 'reactstrap';
+import PropTypes from 'prop-types';
+
+const Cloud = ({ cards }) => {
+  const tags = {};
+  cards.forEach((card) =>
+    card.tags.forEach((tag) => {
+      if (tags[tag]) {
+        tags[tag] += card.asfan;
```
ruler501, line 11: Where does asfan get set here?

src/analytics/Asfans.js
dekkerglen: That's fair, I think the margin of error is incredibly small though, do you want to make a follow up PR that updates the Math here? I only reworked it because the old system made 0 sense to me.

src/analytics/Cloud.js
```diff
@@ -7,9 +7,9 @@ const Cloud = ({ cards }) => {
   cards.forEach((card) =>
     card.tags.forEach((tag) => {
       if (tags[tag]) {
-        tags[tag] += card.asfan;
+        tags[tag] += 1;
```
ruler501, line 10: We can address this in a future PR, but I'd very much like this to be able to be based off asfan or count as an option to the user. I can take that up though with some other improvements I wanted to do to it.

src/analytics/Asfans.js
ruler501: I'll work on it in a followup.

src/utils/draftutil.js
```diff
@@ -338,3 +323,76 @@ export function checkFormat(format, cards) {
   };
   return createPacks({}, format, 1, checkFn);
 }
+
+export function average(arr) {
```
ruler501, line 327: I can take up the making these work with weights in a followup.

src/analytics/PivotTable.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import PivotTableUI from 'react-pivottable/PivotTableUI';
+import 'react-pivottable/pivottable.css';
+
+const PivotTable = ({ cards }) => {
```
ruler501, line 6: Might be good to add a description for those of us who don't know how to use a pivot table.

src/analytics/Table.js
```diff
@@ -0,0 +1,122 @@
...
+            </tr>
+          </thead>
+          <tbody className="breakdown">
+            {secondaryLabels.map((label) => (
+              <tr key={label}>
+                <th scope="col">{label}</th>
+                {primaryLabels.map((primaryLabel) => (
+                  <td>
+                    {groups[primaryLabel][label] || 0}
+                    <span className="percent">{groups[primaryLabel][label] || 0 / cards.length}%</span>
```
ruler501, line 106: I think this should be 
```
                    {label !== 'Total' && (
                      <span className="percent">
                        {((100 * (groups[primaryLabel][label] || 0)) / groups[primaryLabel].Total).toFixed(0)}%
                      </span>
                    )}
```

ruler501:
Found some things testing.

src/analytics/PivotTable.js
dekkerglen: It's a somewhat advanced analytic tool, I'll link to a video describing how to use pivottables in general but I don't think it's something we could feasibly explain here

src/analytics/Asfans.js
dekkerglen: Actually the solution was super easy. Instead of doing a sum with total, I did this:
` total += (1-total) * sum / slot.length;`

ruler501: That's closer but still fails the case where another card is in multiple since length depends on if that was taken already.

src/analytics/Table.js
```diff
@@ -103,7 +103,9 @@ const AnalyticTable = ({ cards }) => {
                 {primaryLabels.map((primaryLabel) => (
                   <td>
                     {groups[primaryLabel][label] || 0}
-                    <span className="percent">{groups[primaryLabel][label] || 0 / cards.length}%</span>
+                    {label !== 'Total' && (
+                      <span className="percent">{(groups[primaryLabel][label] ?? 0) / cards.length}%</span>
```
ruler501: I still think this should divide by `groups][primaryLabel].Total` not `cards.length`. It also needs a to Fixed call.

dekkerglen: I think it makes more sense to show percentage from the total pool of cards

dekkerglen: You can always get the numbers you want by setting one sort to 'unsorted'

PR #1259, opened 2020-04-05
Fix mtgo export for sideboards.

ruler501:
Was only fixing names for mainboard.


routes/cube_routes.js
```diff
@@ -1836,6 +1836,7 @@ router.get('/deck/download/mtgo/:id/:seat', async (req, res) => {
       }
     }
     for (const [key, value] of Object.entries(side)) {
+      const name = key.replace(' // ', '/');
       res.write(`${value} ${key}\r\n`);
```
RKelln: `res.write(`${value} ${name}\r\n`);`

PR #1261, opened 2020-04-08
Add Suggestions Analytic

### LINKED ISSUE #6: Suggested Cuts
As a user, I would like cuts to be suggested for me when I attempt to add a new card.

Mimic CT functionality, where I type in a card, and after it autocompletes, a region is shown with a list of cards I can click to generate the swap in the changelog.

dekkerglen:
Fixes #6 
Adds in recommended cuts and additions. To work, this requires the flask server running from this repo: https://github.com/RyanSaxe/CubeCobraRecommender

If the server is not running, the page will only display 'loading' for both tables. If you want to create some debug data, there are some commented out lines of code you can swap in.

You can see it in action on the dev server: https://codecobradev.com/cube/analysis/15225


routes/cube_routes.js
```diff
@@ -3365,6 +3365,85 @@ router.post(
   }),
 );
 
+router.post(
+  '/api/adds/:id',
+  util.wrapAsyncApi(async (req, res) => {
+    const response = await fetch(
+      `http://127.0.0.1:8000/?cube_name=${req.params.id}&num_recs=${1000}&root=${encodeURIComponent(
```
ruler501, line 3372: Please make the host/port a config param.

Issue #1011, opened 2020-02-04
Add a Note field to cards


ruler501:
**What would you like to be able to do?**
Have a large freeform text field to make notes about a card in my cube.

**Describe how you would use the feature**
Note reasons for inclusion, indicators of past performance, general thoughts, and credits to the artists who did alters.

**Additional context**
Suggested by mauriciopasquier on discord.

mauriciopasquier:
It would be awesome to have these notes as a tooltip on each card, while drafting. That way I can tell new drafters about possible archetypes for a card, or things like packages ("This comes with 3 more Squadron Hawks"), what is the drafted booster for Booster Tutor, etc.

dekkerglen:
Not sure I like having these as tooltips during draft. That feels like a separate feature.

ThijsHoenjet:
I would like to propose another way of doing this: Add multiple key value fields to each card.
How it works:
As default their is a key field filled with "Description" and an empty field next to it to enter text for the description.
Then there is also a + button below that, that will add another key and value field to the form.
This way you can organize any data you want per card in a structured way, while not complicating things for people that just want a description field as that is there by default.

PR #1262, opened 2020-04-08
Add a Note field to cards

ThijsHoenjet:
Issue #1011 Add notes to cards in cube
Also fixed an issue with an uncontrolled react variable imgUrl that was undefined.


ThijsHoenjet:
I didn't add Notes to Cube.PREVIEW_FIELDS in models/cube.js. Maybe it should, let me know.

dekkerglen:
No need, those fields are what's needed to populate the cube preview, this won't be needed

PR #1266, opened 2020-04-10
Add Cube Resizing

dekkerglen:
This PR adds cube resizing. You can find the button here:

image

To view this modal:
image

The change also generates a changelog.

### DIFF
--- a/src/components/ResizeModal.js
+++ b/src/components/ResizeModal.js
@@ -1,4 +1,5 @@
 import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
 import useToggle from 'hooks/UseToggle';
 import CSRFForm from 'components/CSRFForm';
 import FilterCollapse from 'components/FilterCollapse';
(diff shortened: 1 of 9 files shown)


src/components/ResizeModal.js
```diff
@@ -0,0 +1,70 @@
...
+              valid={size.length > 0 && valid}
+              invalid={size.length > 0 && !valid}
+              onChange={(event) => setSize(event.target.value)}
+            />
+            <Input type="hidden" name="filter" value={JSON.stringify(filter)} />
+            <Card className="p-3">
+              <h5>Filter for restrictions:</h5>
+              <p>
+                If you include a filter, this will only add or remove cards that match the filter. If there are not
+                enough cards found to add or remove, your target size may not be reached.
```
ruler501, line 47: Can we make it fail nicer? Give an error saying could only add or remove x cards and not make any changes. If they really want the lower number of changes they can put in that amount.

dekkerglen: It's not really failling - it's like if I say add 100 hound creatures, I won't be able to get there and it will instead add all the hounds

ruler501: I still like the idea of letting the user know they can't grow their 360 cube to 800 with only RTR cards before making changes that are hard to undo.

Issue #684, opened 2019-11-14
"Get Image" of sample pack doesn't use custom images


GooberRF:
If you generate a sample pack that contains cards for which you've configured custom images, then click "Get Image", the generated image doesn't use the custom images, it uses the originals from Scryfall.

ThijsHoenjet:
Fixed in #1272

PR #1269, opened 2020-04-10
Add custom images to sample pack

ThijsHoenjet:
Issue #684 "Get Image" of sample pack doesn't use custom images

Custom images can have different sizes, so the merge images function needs to set the size of each image. The merge-images package doesn't support that yet. However, someone made a pull request with this functionality. For now I have made a copy of it. When the pull request is merged, we can go back to the reference the package.


dekkerglen:
It looks like this change has broken two unit tests - check out the Travis CI output for details. Reach out on discord if you have any issues!

PR #1271, opened 2020-04-11
Aws migration

dekkerglen:
This PR makes the changes necessary to host CubeCobra on Elastic Beanstalk.

There are a few changes for local development. `cubecobrasecrets` is no more! We are now using `dotenv` to handle environment variables. There is a new file `.env-SAMPLE` with a template. More detailed instructions will be included in the readme. 

This also creates a docker container for the purposes of running populate_analytics.js as a cron job on ECS


phulin:
Finally cubecobrasecrets is gone

Issue #509, opened 2019-09-26
User defined sort


dekkerglen:
As a user, I would like to be able to define my own sort, where I label each category and use a filter string to define what gets filtered into each category.

This will allow even greater flexibility and ability to customize how cube lists are displayed.

ruler501:
I can take this up.

dekkerglen:
added

PR #1273, opened 2020-04-14
Hybrid Inclusive Sort [WIP]

lane-eden:
Adds a new Primary Sort, "Color Category (Hybrid Inclusive)," which puts hybrid cards both in the multicolor column and in the mono columns that can cast those cards. For example, Rakdos Cackler will appear under: Black, Red, and Multicolor.


ruler501:
This may run into problems when sorting gets worked on for #509

PR #1285, opened 2020-04-17
Winston cloudwatch

dekkerglen:
sets up cloudwatch for winston


dekkerglen:
Closed - EB is already reporting to cloudwatch. As long as the logger is printing to stdout I'll have it

PR #1287, opened 2020-04-17
Add getters for all card properties

ruler501:
Also adds a schema for card details to updatecards.js

Does not use any of these getters yet, that'll be left to follow up PRs.

### DIFF
--- a/src/utils/Card.js
+++ b/src/utils/Card.js
@@ -77,8 +77,145 @@ export function cardsAreEquivalent(a, b) {
   );
 }
 
-export function getCmc(card) {
-  return card.cmc !== undefined ? card.cmc : card.details.cmc;
-}
+export const getTags = (card) => card.tags;
 
-export default { COLOR_COMBINATIONS, normalizeName, encodeName, decodeName, cardsAreEquivalent, getCmc };
+export const getFinish = (card) => card.finish;
+
+export const getStatus = (card) => card.status;
+
+export const getColorIdentity = (card) => card.colors ?? card.details.color_identity;
+
+export const getCmc = (card) => card.cmc ?? card.details.cmc;
+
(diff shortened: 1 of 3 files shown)


src/utils/Card.js
```diff
@@ -77,8 +77,145 @@ export function cardsAreEquivalent(a, b) {
...
+    ? card.details.price_foil ?? card.details.price
+    : card.details.price ?? card.details.price_foil;
+
+export const cardNormalPrice = (card) => card.details.price;
+
+export const cardFoilPrice = (card) => card.details.price_foil;
+
+export const cardIsFullArt = (card) => card.details.full_art;
+
+export const cardCost = (card) => card.details.parsed_cost;
```
phulin, line 115:
```suggestion
export const cardParsedCost = (card) => card.details.parsed_cost;
```

src/utils/Card.js
```diff
@@ -77,8 +77,145 @@ export function cardsAreEquivalent(a, b) {
...
+
+export const cardId = (card) => card.cardID ?? card.details._id;
+
+export const cardType = (card) => card.type_line ?? card.details.type;
+
+export const cardRarity = (card) => card.rarity ?? card.details.rarity;
+
+export const cardAddedTime = (card) => card.addedTmsp;
+
+export const cardImg = (card) => card.imgUrl ?? card.details.image_normal ?? card.details.image_small;
```
phulin, line 98:
```suggestion
export const cardImageUrl = (card) => card.imgUrl ?? card.details.image_normal ?? card.details.image_small;
```

src/utils/Card.js
ruler501: If we don't have an unparsed cost I don't think adding in parsed adds that much useful information and cardCost is so much shorter. Can still make the change though if you feel strongly about it.

PR #1293, opened 2020-04-18
Don't ignores CMC 0 Cards in Analytics Averages

### LINKED ISSUE #1292: Analytics Averages Tab Ignores CMC 0 Cards
Ignoring lands is correct, but ignoring cards with cmc 0 is wrong.

lane-eden:
Fixes #1292

### DIFF
--- a/src/analytics/Averages.js
+++ b/src/analytics/Averages.js
@@ -18,7 +18,21 @@ const Averages = ({ cards, characteristics }) => {
 
   const counts = Object.entries(groups)
     .map((tuple) => {
-      const vals = tuple[1].map((card) => characteristics[characteristic](card)).filter((x) => x);
+      const vals = tuple[1]
+        .filter((card) => {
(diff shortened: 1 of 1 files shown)


src/analytics/Averages.js
```diff
@@ -18,7 +18,20 @@ const Averages = ({ cards, characteristics }) => {
 
   const counts = Object.entries(groups)
     .map((tuple) => {
-      const vals = tuple[1].map((card) => characteristics[characteristic](card)).filter((x) => x);
+      const vals = tuple[1]
+        .filter((card) => {
+          if (characteristic === 'CMC') {
```
dekkerglen, line 23: can you add a comment describing why we're accommodating this edge case? I don't want people to look at this in the future and get confused why we're doing this

lane-eden: Sure

Issue #1279, opened 2020-04-16
Historic Cube Category


dekkerglen:
As a user, I'd like to be able to categorize my cube as 'historic'

PR #1300, opened 2020-04-19
Adds Historic cube category override

limanuljaska:
Users can override the cube category as Historic. Does not implement any legality or analytics related functionality for Historic.


limanuljaska:
Resolves #1279

dekkerglen:
This will also need a change in `src/components/CubeSearchNavbar.js`, around line 21

PR #1311, opened 2020-04-21
Delete decks

lane-eden:
- Allows users to delete decks they draft on any cube
- Allows the cube owner to delete any draft on cubes they own.


src/components/DeckDeleteModal.js
```diff
@@ -0,0 +1,73 @@
...
+    this.focusAcceptButton = this.focusAcceptButton.bind(this);
+  }
+
+  focusAcceptButton() {
+    if (this.acceptButton.current) {
+      this.acceptButton.current.focus();
+    }
+  }
+
+  confirm() {
```
ruler501, line 24: Couldn't this be async so it doesn't need a then callback?

lane-eden: I had copied that code from the BlogPostDelete modal, so I refactored the two modals into a ConfirmDeleteModal component and made both confirm functions use async.

PR #1329, opened 2020-04-28
Fix category cube search.

ruler501:


dekkerglen:
Need to update tests to reflect changes

nearley/cubeFilters.ne
```diff
@@ -49,10 +50,39 @@ categoryCondition -> ("category"i) (prefixCategoryOpValue | overrideCategoryOpVa
...
 
-prefixCategoryOpValue -> ":" prefixCategoryValue {% ([, prefix]) => genericCondition('categoryPrefixes', { $regex: prefix, $options: 'i' }) %}
+@{%
+const categoryConditionFunc = (property, category, value) => ({ 
+  query: {
+    $or: [
+      {
+        $and: [
+          { overrideCategory:true },
+          category,
```
dekkerglen, line 60: shouldn't this be: `{category}` ?

ruler501: No because the query was already built in the caller.

PR #1337, opened 2020-05-02
Hotfix

dekkerglen:
Fixes an issue where cube drafts seeds are always the same.
Fixes an issue where recommender wouldn't render
Adds a script to update schema violating cubes

### DIFF
--- a/models/cube.js
+++ b/models/cube.js
@@ -19,7 +19,7 @@ const Card = {
   colors: [
     {
       type: String,
-      enum: ['W', 'U', 'B', 'R', 'G'],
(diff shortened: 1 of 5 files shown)


models/cube.js
```diff
@@ -19,7 +19,7 @@ const Card = {
   colors: [
     {
       type: String,
-      enum: ['W', 'U', 'B', 'R', 'G'],
+      enum: ['W', 'U', 'B', 'R', 'G', 'C'],
```
ruler501, line 22: What cards use C instead of an empty array?

PR #1342, opened 2020-05-04
Draft size saving

ruler501:
Make everything indices into a central array for cards to save space. Also ensures we don't save card details as much as possible. Has migrations now.

This does assume that if two cards have the same ID then they are the same card, that can be invalid with custom cards, but I'm not sure there's a better way to handle it.


dekkerglen:
Closing for now - we wil revisit later

PR #1349, opened 2020-05-06
Add full docker image building and upload to Github packages

ruler501:


dekkerglen:
Why do we want this

ruler501:
It makes it easier to spin up test instances and validate setup instructions. Mostly I maintain this separately so I can use it for my test server and I'd like to just merge it in so I don't have to maintain a separate branch for it.

PR #1353, opened 2020-05-07
Use synergy in draft bots

ruler501:
Uses synergy with picks and internal synergy to draft cards, choose colors, and build a deck.


dekkerglen:
Did you fix the issue with pick order?

ruler501:
It's fixed now

PR #1358, opened 2020-05-09
Dr4ft export

thyeggman:
Adding a [dr4ft](https://dr4ft.info/) export option to cube decks.

TODO: test, add screenshot

Also did some refactoring of the deck export code to reduce it by a good chunk.

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -1569,6 +1569,25 @@ const exportToMtgo = (res, fileName, mainCards, sideCards) => {
   return res.end();
 };
 
+const writeDeckPart = (deck, res, resultFormat, nameFormatter) => {
+  const deckPart = {};
+  for (const col of deck) {
+    for (const card of col) {
+      const details = carddb.cardFromId(card.cardID);
+      const name = nameFormatter(details);
+      if (deckPart[name]) {
+        deckPart[name] += 1;
+      } else {
+        deckPart[name] = 1;
+      }
+    }
+  }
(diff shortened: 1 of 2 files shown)


dekkerglen:
Isn't dr4ft format the same as cockatrioce?

thyeggman:
@dekkerglen Default is slightly different from what I can tell.  The cockatrice export uses `1 Lightning Bolt` but an export from the dr4ft site uses `1x Lightning Bolt`.  I still need to do testing to see if they're compatible.  It may well be that they are, but when implementing I trusted that the cockatrice format was correct, and the dr4ft format differed, so I wrote a new method.

As a side note, if they are compatible, would you prefer one menu item with `Cockatrice / dr4ft (.txt)`, or separate buttons?  I think they should still have their own routes even if the dr4ft one just calls the original cockatrice implementation

dekkerglen:
I think leaving it cockatrice is fine. What's the use case for exporting a deck to dr4ft? To use their deckbuilder? 

Dr4ft now has the ability to import from CC using their UI. Dr4ft and cockatrice are also in the same ecosystem, for what that's worth.

thyeggman:
Ah, it looks like I misunderstood #1358 ... The user was asking for dr4ft import, not export like the title says 😓

The dr4ft export format differs from what the CC importer expects, so I think that's just a matter of modifying the regex or whatever it uses (I haven't checked) to accept the `1x`

I can remove the dr4ft export if you think it's just bulk (or just update the UI with dr4ft / cockatrice if they're compatible), but I'll make this PR about the import.

PR #1362, opened 2020-05-14
Improvements to Analytics

ruler501:
Has asfan used on all tabs where it is relevant.
Adds formatId and activeTab to the query parameters so you can share precise links.


dekkerglen:
conflicts

dekkerglen:
Can you post screenshots of how what the asfan changes look like? I strongly disliked how asfans were handled previously and don't want to go back to that. I do really want to merge the url query parameters though.

dekkerglen:
image
I think this piece should be moved here (same with all instances)

image
The dropdown should follow the style of other dropdowns - probably like this:
image

If you are doing this because this is the only way to get the divider, that's fine. Regardless, you shouldn't display the divider if there are no custom draft formats.

src/analytics/Averages.js
```diff
@@ -1,51 +1,57 @@
-import React, { useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import PropTypes from 'prop-types';
 
 import { Col, Row, Table, InputGroup, InputGroupAddon, InputGroupText, CustomInput } from 'reactstrap';
 
-import { sortIntoGroups, getSorts } from 'utils/Sort';
-import { average, median, stddev } from 'utils/draftutil';
+import AsfanDropdown from 'components/AsfanDropdown';
```
dekkerglen, line 6: AsfanDropdown does not exist in this repo

dekkerglen:
Failing CI

dekkerglen:
Conflicts

PR #1370, opened 2020-05-22
Fix issue where empty maybeboard

### LINKED ISSUE #1334: Filtering a list sometimes hides results
In list view, when filtering for a very small set of cards, the result page index doesn't change, even if the filtered list is very small. If the filtered list is shorter than one page, this can lead to no cards being displayed and no way to view the result.

To reproduce:
1. View a cube using "List View". (The cube should contain more than one page worth of cards.)
2. Navigate to the last page of the list.
3. Filter the list to a set of results with less than one page of cards (e.g. `name:"Black Lotus"`).

After applying a filter, the site should ensure that the user is trying to view a page of the list that exists.

Before applying filter:
image
After applying filter:
image

### LINKED ISSUE #1345: undefined error when selecting analysis
**Describe the bug**
When I select Analysis for either of my cubes, I get an error:

Cannot read property 'forEach' of undefined

**To Reproduce**
1. Go to https://cubecobra.com/cube/overview/rckhans
2. Click on Analysis
3. See error

**Expected behavior**
Should be able to see the Analysis page for the cube.

**Screenshots**

image

 - OS: Win10
 - Browser: Chrome
 - Version 81.0.4044.129 (Official Build) (64-bit)

**Additional context**
Error appears whether logged in or out.

dekkerglen:
Fixes #1345

Also fixes the issue of hiding elements in list paginated elements (accidentally comitted to this branch)
Fixes #1334

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -911,8 +911,8 @@ router.get('/analysis/:id', async (req, res) => {
       });
       return cards;
     };
-    cube.cards = addDetails(cube.cards);
-    cube.maybe = addDetails(cube.maybe);
+    cube.cards = addDetails(cube.cards || []);
+    cube.maybe = addDetails(cube.maybe || []);
 
     const priceDictQ = GetPrices([...pids]);
     const eloDictQ = getElo([...cardNames], true);
--- a/src/components/PagedTable.js
+++ b/src/components/PagedTable.js
@@ -8,10 +8,9 @@ import { Table } from 'reactstrap';
 const PagedTable = ({ pageSize, rows, children, ...props }) => {
   const [page, setPage] = useState(0);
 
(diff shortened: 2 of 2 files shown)


routes/cube_routes.js
```diff
@@ -911,8 +911,8 @@ router.get('/analysis/:id', async (req, res) => {
       });
       return cards;
     };
-    cube.cards = addDetails(cube.cards);
-    cube.maybe = addDetails(cube.maybe);
+    cube.cards = addDetails(cube.cards || []);
```
RKelln, line 914: Is there an easy way to set (the default value of) `.cards` and `.maybe` to `[]` for all cubes retrieved from `fineOne()`? That would prevent this sort of problem from happening in the future too.

dekkerglen: We have a default value in the Schema but that only works for new cubes

Issue #1302, opened 2020-04-19
Add first cube button styling broken 


dekkerglen:
New users will be greeted to this dashboard page:

This button has incorrect styling 

![image](https://user-images.githubusercontent.com/28238025/79691493-de200100-822d-11ea-8c75-76374cc758f2.png)

PR #1373, opened 2020-05-23
Add first cube button styling broken #1302

jyandell83:
Removed an a tag which was setting text color to "inherit". Changed color to "success" which is green.

Loved the MTGO cube, not totally sure how this works, just learning, but would like to help.


dekkerglen:
Most of the dev discussion occurs here: https://discord.gg/Hn39bCU
@Dekkaru there and I'll give you access to the hidden channels

PR #1383, opened 2020-05-28
Enable searching by quantity of tags.

ruler501:


dekkerglen:
Can you add documentation for this filter in the filter page? I think it's especially important because this is not something scryfall supports

PR #1392, opened 2020-05-29
Fix elo and prices for analyics

dekkerglen:
Fixes an issue with populate analytics where it would not fetch elo and prices correctly

### DIFF
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -225,6 +225,16 @@ function maybeCards(cube, carddb) {
   return maybe.map((card) => ({ ...card, details: carddb.cardFromId(card.cardID) }));
 }
 
+async function getCardElo(cardname) {
+  const rating = await CardRating.findOne({ name: { $regex: new RegExp(cardname, 'i') } });
+
+  if (!rating || Number.isNaN(rating.elo)) {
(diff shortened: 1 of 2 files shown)


serverjs/cubefn.js
```diff
@@ -225,6 +225,16 @@ function maybeCards(cube, carddb) {
   return maybe.map((card) => ({ ...card, details: carddb.cardFromId(card.cardID) }));
 }
 
+async function getCardElo(cardname) {
+  const rating = await CardRating.findOne({ name: { $regex: new RegExp(cardname, 'i') } });
```
phulin, line 229: This won't use the Mongo index and will thus be slow. Probably fine for where the function is being used now, but we're going to want to fix this in the future.

phulin:
Also, CI failing.

serverjs/cubefn.js
dekkerglen: This had reasonable perf when I ran it, and the size of that table grows extremely slow so I think it's ok

phulin: Yep, I'm just worried we'll accidentally let usage of this utility function sneak in to actual routes that get called much more frequently than populate_analytics.

PR #1398, opened 2020-05-30
Add new Quick Filter function.

phulin:
This PR adds a set of Quick Filter boxes right below the existing filter input. This is just an easier way for users who don't want to use the search syntax to get access to simple filters. Screenshot:

image


dekkerglen:
I feel like the quick filter button should be `success`

phulin:
Fixed colors.
image

PR #1401, opened 2020-05-30
Add per cube analytics for cards.

ruler501:
Does not do anything with the data yet, just populates it.


dekkerglen:
I don't think this change warrants a one-shot script - instead we should just have null checks and handle it that way

ruler501:
I included both checks and the script. Can remove the script then.

dekkerglen:
Conflicts

PR #1405, opened 2020-06-01
Fix top cards bugs.

phulin:
This PR fixes several issues with the Top Cards page, including the bug where data is garbage after the first two pages or so. It should also make a dramatic improvement in the performance of that page, and it should begin the process of focusing our data infrastructure on oracle IDs rather than card names.


phulin:
Should be ready to go now.

Issue #1399, opened 2020-05-30
Compare Cubes - Show Totals at Top of Screen


Gabrosin:
I often use the compare cubes tool to examine cubes like my own.  Having the overall count of differences between the two cubes would be a big help, rather than having to scroll through the page and do the mental math myself.

Should be as simple as re-using the header component with a title like Totals and the sums of the various other sections on the page.  Thanks!

bramleyjl:
I'll take a look at this one.

Gabrosin:
Thanks for the quick turnaround, the header is just what I was hoping for!

PR #1434, opened 2020-06-10
Comparison header

bramleyjl:
Addresses issue #1399 by adding a header to the cube comparison display showing the total number of shared and different cards in each cube.


dekkerglen:
Need to run `npm run beautify`

PR #1435, opened 2020-06-11
Add spinner for bot only draft.

ruler501:


src/components/LoadingButton.js
```diff
@@ -25,5 +25,12 @@ const LoadingButton = ({ onClick, loading, ...props }) => {
     </div>
   );
 };
+LoadingButton.propTypes = {
+  onClick: PropTypes.func.isRequired,
+  loading: PropTypes.bool,
+};
+LoadingButton.defaultProps = {
+  loading: undefined,
```
phulin, line 33:
```suggestion
  loading: null,
```

phulin: I think this is a better pattern, requires a change above

ruler501: I agree, was just trying to avoid changing component code.

PR #1444, opened 2020-06-12
Fix hyper geom anaysis labels and add hint text

### LINKED ISSUE #1427: Hypergeometric Calculator labels are switched
**Describe the bug**
The labels for "Sample size" and "Number of successes in population" are switched.

**To Reproduce**
1. Visit the analysis page and use the hypergeo calculator.

**Screenshots**
strange result
Strange result when using the fields as labeled.

Bildschirmfoto von 2020-06-06 18-41-49
Correct result when using the fields with the wrong labels.

wirmar:
fixes #1427 
HyperGeom_correct_labels

I also added some hint text on how to use the hypergeom calculator. If this goes against the design philosophy I can remove that commit.
HyperGeom_hinttext

### DIFF
--- a/src/analytics/HyperGeom.js
+++ b/src/analytics/HyperGeom.js
@@ -2,7 +2,18 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import ChartComponent from 'react-chartjs-2';
 
-import { Col, Row, Table, InputGroup, InputGroupAddon, InputGroupText, CustomInput, Input, Button } from 'reactstrap';
+import {
+  Col,
+  Row,
+  Table,
+  InputGroup,
+  InputGroupAddon,
+  InputGroupText,
+  CustomInput,
+  Input,
+  Button,
+  Form,
+} from 'reactstrap';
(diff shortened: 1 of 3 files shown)


src/analytics/HyperGeom.js
```diff
@@ -95,9 +95,11 @@ const HyperGeom = () => {
     return Math.min(Math.max(val, min), max);
   };
 
-  const calculate = (N, S, n, s) => {
-    const keys = [...Array(parseInt(s, 10) + 1).keys()];
-    const values = keys.map((x) => hyp(parseInt(N, 10), parseInt(S, 10), parseInt(n, 10), x));
+  const calculate = (populationSize, sampleSize, popSuccesses, sampleSuccesses) => {
```
wirmar, line 98: I changed the names of these arguments to be readable, since that is where the error came from.

wirmar: Whoops that introduced a lint error due to duplicate variable names.
I moved the calculation code to a util file.

PR #1455, opened 2020-06-13
[WIP] Draftbot scoring tweaks.

phulin:


src/utils/draftbots.js
```diff
@@ -77,135 +79,156 @@ export const fetchLands = {
...
+  'Fabled Passage': [...'WUBRG'],
+  'Terramorphic Expanse': [...'WUBRG'],
+  'Evolving Wilds': [...'WUBRG'],
 };
 
+// What is the raw power level of this card?
+// Scale is roughly 0-1, linear. Black Lotus should be ~1.
 export const getRating = (combination, card) => {
-  return Math.log(COLOR_SCALING_FACTOR[combination.length] * 10 ** ((card?.rating ?? 0) / 400));
+  return (card?.rating ?? 1200) / 1200 - 1;
```
ruler501, line 91: This should still be exponential since that's how elo models value

src/utils/draftbots.js
```diff
@@ -77,135 +79,156 @@ export const fetchLands = {
...
+// Has this color combination been flowing openly?
+// Scale from 0-1.
 export const getOpenness = (combination, seen) => {
   if (seen.cards.length === 0) {
     return 0;
   }
 
-  const seenCount = COLOR_SCALING_FACTOR[combination.length] * seen[combination.join('')];
-  return Math.log(seenCount);
+  return seen[combination.join('')] / seen.cards.length;
```
ruler501, line 157: We don't want an average here, we want this to grow as we see more cards.

src/utils/draftbots.js
```diff
@@ -77,135 +79,156 @@ export const fetchLands = {
...
-    COLOR_SCALING_FACTOR[combination.length] * (picked[combination.join('')] + 10 ** ((card?.rating ?? 0) / 400)),
-  );
+// How good are the cards we've already picked in this color combo?
+// Scale from 0-1.
+export const getColor = (combination, picked) => {
+  if (picked.cards.length === 0) {
+    return 0;
+  }
+
+  return picked[combination.join('')] / picked.cards.length;
```
ruler501, line 167: This absolutely can't be an average, we need to commit more to a color the more we take of it.

phulin:
Isn't that accomplished by the weights?

ruler501:
> Isn't that accomplished by the weights?

Not exactly, we want it to grow in proportion to our commitment to it for color or the lack of commitment we see from others for openness. Averages just tell us if on average our cards are good, so it would say we should play black when we have 1 griselbrand vs green with 15 elves.

PR #1458, opened 2020-06-15
fixed totals row in table analysis

### LINKED ISSUE #1282: Table total column counts cards without the attribute
For example:
image

Bottom right corner shouldn't be the cube size

wirmar:
fixes #1282 

SeedCube02 ColorIdentity/Toughness table:
Table_ColorCat_Toughness_SeedCube02
The totals are calculated from the amount of cards that were "considered" for the table. The percentages in the last column are also in reference to the `Total.Total`.

### DIFF
--- a/src/analytics/Table.js
+++ b/src/analytics/Table.js
@@ -7,14 +7,14 @@ import AsfanDropdown from 'components/AsfanDropdown';
 import ErrorBoundary from 'components/ErrorBoundary';
 import { sortIntoGroups, getSorts, getLabels, cardCanBeSorted } from 'utils/Sort';
 
-const AnalyticTable = ({ cards, cube, defaultFormatId, setAsfans }) => {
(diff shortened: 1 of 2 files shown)


src/analytics/Table.js
```diff
@@ -5,14 +5,17 @@ import { Col, Row, Table, InputGroup, InputGroupAddon, InputGroupText, CustomInp
...
+import { sortIntoGroups, getSorts, getLabels, cardCanBeSorted } from 'utils/Sort';
 
 const AnalyticTable = ({ cards, cube, defaultFormatId, setAsfans }) => {
   const sorts = getSorts();
 
   const [primary, setPrimary] = useState('Color Identity');
   const [secondary, setSecondary] = useState('Type');
 
+  // some criteria cannot be applied to some cards
+  cards = cards.filter((card) => cardCanBeSorted(card, primary) && cardCanBeSorted(card, secondary));
```
dekkerglen, line 17: modifying props is bad practice. We should be storing this in a new variable

PR #1462, opened 2020-06-16
Top Cards Fix and Major perf updates

dekkerglen:


routes/tools_routes.js
```diff
@@ -122,16 +120,18 @@ function shuffle(a) {
 }
 
 router.get('/api/topcards', async (req, res) => {
+  console.log(`getting top cards with sort: ${req.params.sort}, page: ${req.params.page}`);
```
phulin: remove

routes/tools_routes.js
```diff
@@ -140,27 +140,16 @@ router.get('/api/topcards', async (req, res) => {
...
 
 router.get('/topcards', async (req, res) => {
   try {
-    const { err, filter } = makeFilter(req.query.f);
-
-    if (err) {
-      req.flash('Invalid filter.');
-    }
-
-    const { data, numResults } = await topCards(filter, res);
```
phulin, line 155: The reason it was set up this way before was to avoid having a loading spinner on initial load. I get that simplicity is good, but it doesn't seem like much code to keep that functionality in.

routes/cube_routes.js
```diff
@@ -826,10 +826,12 @@ router.get('/playtest/:id', async (req, res) => {
       },
       'date seats _id cube',
     )
+      .lean()
```
phulin: I don't think this does anything

routes/tools_routes.js
```diff
@@ -69,47 +66,48 @@ async function topCards(filter) {
   }
 
   const oracleIds = [...oracleIdMap.keys()];
-  const query = filter ? { oracleId: { $in: oracleIds } } : {};
+  const query = filter
+    ? { $and: [{ oracleId: { $in: oracleIds } }, { 'current.picks': { $gt: MIN_PICKS } }] }
```
phulin, line 70:
```suggestion
    ? { oracleId: { $in: oracleIds }, 'current.picks': { $gt: MIN_PICKS } }
```
I think this works

routes/tools_routes.js
```diff
@@ -69,47 +66,48 @@ async function topCards(filter) {
...
-            qualifies && Number.isFinite(elo) ? elo : null,
-            Number.isFinite(cubes) ? cubes : 0,
-          ];
-        });
-    }),
-  );
-  const numResultsQ = CardHistory.estimatedDocumentCount(query);
-  const [allData, numResults] = await Promise.all([dataQ, numResultsQ]);
-  const [dataByRating, dataByElo, dataByPicks, dataByCubes] = allData;
+  const dataQ = async () => {
```
phulin, line 79: I don't think this needs to be in an async function. Just move the [data, numResults] to right after CardHistory.find

PR #1467, opened 2020-06-18
Fix querying exact values of power/toughness.

ruler501:


ruler501:
> Is there a reason why this has parseFloat but price doesn't?

Price is guaranteed to be a float already due to how we populate it.

PR #1468, opened 2020-06-19
Allow overriding elo for a cube by cube basis.

ruler501:


dekkerglen:
I'm not sure we want this. What is the use case? What is the advantage? Is it worth the UI bloat?

PR #1471, opened 2020-06-21
Large tweaks to draftbots

ruler501:
Significant performance improvements as well as some tweaks to get more playables. Normalizes each of the factors to roughly the range [0, 10].

Includes changes from #1455


dekkerglen:
Conflcits

PR #1475, opened 2020-06-24
Allow Showing breakdown while drafting

ruler501:
Depends on #1471


dekkerglen:
conflicts

src/components/DraftbotBreakdown.js
```diff
@@ -118,51 +130,10 @@ const TRAITS = [
...
-    }
-    return res;
-  }, [deck, draft, index, seatIndex]);
-
-  // load the weights for the selected pack
+export const Internal = ({ cardsInPack, draft, pack, picks, picked, seen }) => {
   const weights = useMemo(() => {
     const res = [];
-    for (let i = 0; i < TRAITS.length - 2; i++) {
+    for (let i = 0; i < TRAITS.length - 3; i++) {
```
RKelln, line 136: add comment why you're using 3 as magic number?

src/components/DraftbotBreakdown.js
```diff
@@ -206,6 +178,12 @@ const DraftbotBreakdown = ({ draft, seatIndex, deck, defaultIndex }) => {
         card.scores[i] = `+${(card.scores[i] - minScore).toFixed(2)}`;
       }
     }
+  } else {
+    for (let i = 0; i < TRAITS.length - 2; i++) {
```
RKelln, line 182: similarly, add comment as to why using 2 as magic number here

Issue #1492, opened 2020-06-27
`npm run populate-analytics` errors to ReferenceError


Cadiac:
**Describe the bug**
Running npm run populate-analytics at master branch errors, as it tries to access `rating` at https://github.com/dekkerglen/CubeCobra/blob/master/jobs/populate_analytics.js#L179 which no longer exists after recent refactoring.

```
$ npm run populate-analytics
...
Finished: 21500 of 21634 correlations.
Finished: 21600 of 21634 correlations.
Finish init of correlation matrix.
Started: cubes
Finished: all cubes
Started: decks
Finished: all decks
(node:24347) UnhandledPromiseRejectionWarning: ReferenceError: rating is not defined
    at processCard (/project/path/here/CubeCobra/jobs/populate_analytics.js:179:29)
    at /project/path/here/CubeCobra/jobs/populate_analytics.js:317:13
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
(node:24347) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:24347) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

**To Reproduce**
1. Checkout master branch. Run `npm run populate-analytics`
2. Wait until the script errors

**Expected behavior**
No errors, script runs to end

**Desktop (please complete the following information):**
 - OS: Manjaro, node v14.3.0
 - `master branch`, commit 4418d213

Cadiac:
https://github.com/dekkerglen/CubeCobra/pull/1495#issuecomment-650613025

PR #1495, opened 2020-06-27
Add rating lookup that was removed earlier back #1492

Cadiac:
This PR adds an apparent fix to #1492 `npm run populate-analytics` script, which is currently not working at master as it tries to use variable `rating` which no longer exists.

I added back the CardRating database query and at least for me locally the script seems to work with it.

This seems to be have been caused by a possibly destructive merge, so someone who knows more about this script should probably take a look at what has happened here.


dekkerglen:
This script is broken, but we don't need to do this because the rating information is now in the local card files

Issue #1049, opened 2020-02-14
Dragging cards no longer works on mobile.


phulin:

Cadiac:
The bug was found and fix PR is up.

PR #1499, opened 2020-06-28
Fix touch device detection to work again #1049

Cadiac:
Touch device detection was broken (at least on the devices I tested this with and as per issue #1049), which broke dragging cards at deckbuilder on mobile.

This PR fixes the touch device detection, and uses the correct backend for react-dnd on touch devices.

Also change how often the detection is executed to only happen once per DndProvider initialization. As I was debugging this I noticed that the check was being made constantly while cards were being dragged, and as real devices should never change from being touch devices to not being them doing the check only once seems sane and performant.

Tested the change on both emulator & real android phone / desktop browsers.

### DIFF
--- a/src/utils/Draft.js
+++ b/src/utils/Draft.js
@@ -40,7 +40,7 @@ export const addSeen = (seen, cards, synergies) => {
         const combStr = comb.join('');
         if (COLOR_INCLUSION_MAP[combStr][colorsStr]) {
           for (const { index } of seen.cards[combStr]) {
-            if (synergyMatrix[index][card.index] === null) {
+            if (synergyMatrix[index][card.index] === null && synergies) {
               const similarityValue = similarity(synergies[card.index], synergies[index]);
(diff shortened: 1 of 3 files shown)


src/utils/Draft.js
```diff
@@ -40,7 +40,7 @@ export const addSeen = (seen, cards, synergies) => {
         const combStr = comb.join('');
         if (COLOR_INCLUSION_MAP[combStr][colorsStr]) {
           for (const { index } of seen.cards[combStr]) {
-            if (synergyMatrix[index][card.index] === null) {
+            if (synergyMatrix[index][card.index] === null && synergies) {
```
ruler501, line 43: Need to set the value to zero so we don't add null below

Cadiac: Seems like I accidentally committed this workaround to get the draft view working at master now. I'll just revert that as it was never supposed to end up in this PR anyways, good catch.

PR #1502, opened 2020-06-28
Fix interpolation and dealing with null synergy.

ruler501:


src/utils/draftbots.js
```diff
@@ -192,7 +194,7 @@ export const getColor = (combination, picked) => {
 
 const getCoordPairs = (pack, pick, initialState) => [
   [pack - 1, initialState[0].length],
-  [pick - 1, initialState[0][0].length],
+  [pick - 1, initialState[0][pack - 1].length],
```
RKelln, line 197: Do you need `Math.max(pack - 1, 0)` here?

ruler501: Probably to be fully correct, the input should never be less than 1 though

dekkerglen: lets do this just in case this accidentally gets called after the last pick

Issue #1500, opened 2020-06-28
Run eslint & prettier as pre-commit hook


Cadiac:
**What would you like to be able to do?**
It is easy to accidentally submit code to PRs which doesn't pass lint/prettier checks. I would like to have pre-commit hook run prettier & eslint and force the submitted code to pass these checks.

**Describe how you would use the feature**
While running eslint/prettier should probably happen at editor, as we have many people working on this project enforcing unified code styles at commit time would make sense and prevent unnecessary noise to PRs failing these checks. It is good that the CI server runs these checks now, but repeatedly doing "fix styles" type of commits is frankly quite embarassing :innocent: 

**Additional context**
Good tooling for this exists, and I'm happy to try on adding this setup to this repo. A way to bypass that hook should also be considered for quick "just commit this now" hotfix situations.

PR #1504, opened 2020-06-29
Pre-commit hook to run eslint/prettier

Cadiac:
After making the mistake of submitting unlinted code to PRs and having them fail at CI, I think (#1500) we should add pre-commit hooks to enforce that committed code adheres to linting rules. While properly configured IDE (like VSCode with ESLint plugins) does pretty much the same thing, as more and more people work on this projects mistakes will happen.

`lint-staged` setups a git pre-commit hook on npm install that runs eslint with --fix option on staged files when commit is made. If fix can't be applied automatically, the commit will fail and you have an option to fix the issue manually.

To make this change take effect, `npm install` must be run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,14 @@ ESLint Extension for VSCode: https://marketplace.visualstudio.com/items?itemName
 
 VSCode (with the ESLint extension) is the recommended environment. When using this setup, make sure that your selected workspace is the root folder that you have cloned, this will ensure that the ESLint plugin can work with our linting rules. Using this plugin will make adhering to the linting rules significantly easier.
 
+This project also contains pre-commit hook that is installed on `npm install` by using `lint-staged`. Linting is also checked at CI tests when PRs are submitted. These hooks enforce that only code adhering to our ESLint rules gets committed to this repository, and automatically fixes fixable issues.
+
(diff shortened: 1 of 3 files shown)


ruler501:
Don't run with --fix on git commit. Let it fail and require the user to run something like `npm run lint-fix` to fix the issues. We should not be mutating state in a precommit hook. Does this properly stash and unstash unstaged changes and if so how does it work with merge commits?

Cadiac:
This approach can also be used without the --fix flag, which will then prompt the user to fix the detected issues manually somehow. I understand your concerns about different code being actually committed than what was being edited, and I admit that using it is a little bit lazy.

A backup stash is always done before any linting is done, and in case of an error at the tasks being run any modifications are rolled back.

I'll change this PR to not use the --fix flag if we're not feeling confident in using that.

Cadiac:
Without the `--fix` this now only runs the linter and aborts the commit if something is wrong:

```

Cadiac:
I'll test this with some merge cases and submit the examples here later.

Cadiac:
Lets just forget about this, I didn't have the time to create proper examples of its usage and this is such a minor thing anyways.

Issue #1055, opened 2020-02-15
Not enough cards to start draft should not Console.error


dekkerglen:
This is a 400 error code, not a 500. We should not be logging them in the server logs.

PR #1506, opened 2020-06-29
Make starting sealed with too small card pool 4xx error

Cadiac:
Addresses #1055, which seems to be already fixed for drafts but starting sealed with too small pool still resulted in 5XX level error being logged to server logs.


Cadiac:
Once this is merged #1055 can probbably be closed.

dekkerglen:
@Cadiac if you include the string "Fixes #issue", GitHub will close that issue number when the PR is merged :)

PR #1515, opened 2020-07-14
Added api route to export cubes to JSON via API

### LINKED ISSUE #1514: JSON Export for cubes
Currently, a lot of information on cubes cannot be obtained through the API. This issue is for an API endpoint to get a JSON object back containing relevant cube info.

@dekkerglen raised the concern of throttling becoming necessary if this were to become implemented.

Likewise42:
Fixes #1514 

Simple PR to allow cube objects to be retrieved from an API call. @dekkerglen had concerns about throttling if this is added, so this PR can probably sit until that is implemented.

### DIFF
--- a/app.js
+++ b/app.js
@@ -239,8 +239,8 @@ app.post('*', (req, res, next) => {
 const apiLimiter = rateLimit({
   windowMs: 60000,
   max: 1,
-  message: '429: Too Many Requests'
-})
+  message: '429: Too Many Requests',
+});
 app.use('/cube/api/cubeJSON', apiLimiter);
 
 // Route files; they manage their own CSRF protection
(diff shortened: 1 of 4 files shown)


dekkerglen:
Use the annotation "Fixes #", so it will auto close that issue when merged

Likewise42:
Updated description as requested

joshfinnie:
I heard some rumblings about the worry of someone scraping this endpoint. Have we thought of something like https://www.npmjs.com/package/express-rate-limit to help with that?

dekkerglen:
@joshfinnie yes that's the way to go

Likewise42:
Should we make a separate issue for that?

dekkerglen:
I think adding that to this PR would be good. You can configure throttle rates on a per route basis, so just apply it to this one for now

PR #1517, opened 2020-07-16
fix: update password reset to use lowercase emails

### LINKED ISSUE #1516: Password reset email needs to be case insensitive
Password recovery is not case insensitive, although the email check is. The entire process needs to be lowercased at all steps.

malexanderboyd:
Fixes #1516 

- refactored all references to `req.body.email` to `recoveryEmail` and cast to lowercase.

issue-1516

### DIFF
--- a/routes/users_routes.js
+++ b/routes/users_routes.js
@@ -148,7 +148,7 @@ router.post('/lostpassword', [body('email', 'Email is required').isEmail()], fla
   if (!req.validated) {
     res.render('user/lostpassword');
   } else {
-    const recoveryEmail = req.body.email.toLowerCase()
+    const recoveryEmail = req.body.email.toLowerCase();
     PasswordReset.deleteOne(
       {
         email: recoveryEmail,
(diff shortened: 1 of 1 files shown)


routes/users_routes.js
```diff
@@ -235,7 +236,7 @@ router.post(
           } else {
             User.findOne(
               {
-                email: req.body.email,
+                email: recoveryEmail,
```
malexanderboyd, line 239: This is the only one I wasn't 100% with - should we use the email exactly how the user types it or downcast it? Not sure if `findOne` is case-insensitive under the covers.

dekkerglen: We should use findOne without the regex, as it is O(1) vs O(n). As long as we consistently normalize on write and on read, it won't be an issue

malexanderboyd: Makes sense! 

Not sure how long the writes have been normalized to lowercase (new contributor), but for any user that somehow registered without being normalized to lower case would be unable to recover their password following this?

dekkerglen: No, the login recovery documents are only valid for 15 min. We already normalize emails for the User table

PR #1519, opened 2020-07-17
Ads

dekkerglen:
Adds an advertisment banner


src/components/Advertisement.js
```diff
@@ -0,0 +1,64 @@
...
+        />
+      </div>
+    </Col>
+  );
+};
+
+const Advertisment = () => {
+  const [option] = useState(Math.floor(Math.random() * 4));
+
+  if (option === 0) {
```
malexanderboyd, line 26: Considered moving these options into their own object? Instead of duplicating the component code, we could simplify a bit:

```
  const ads = [
    {
      text: "Cube Cobra's hosting fees are covered completely by donations. If you enjoy the service Cube Cobe provides, please consider",
      link: {
        text: "donating",
        href: "https://www.patreon.com/cubecobra"
      }
    },
     ...
 ]
if (shownAdvertisement <= 1) {
    return (
      <Col x="12" className="py-2">
        <Card>
          <CardBody>
            {ads[shownAdvertisement].text}
            <strong>
              <a href={ads[shownAdvertisement].link.href}> {ads[shownAdvertisement].link.text}</a>
            </strong>
          </CardBody>
        </Card>
      </Col>
    )
  } else {
    return <TCG />;
  }
```

Issue #1520, opened 2020-07-17
Cube Creation flash errors not displaying


dekkerglen:
Cube names must have a min length of 5. When creating a cube with the name 'Cube' you should get a red flash error, this is currently not displaying.

PR #1537, opened 2020-07-26
🐛 Fixed flash errors on cube creation

GV14982:
Flash errors now display when you try to create a cube with an invalid name

Changed the pug file for user_view to use flash instead of dynamic_flash


GV14982:
#1520

PR #1544, opened 2020-08-01
Add card search, rework card pages

### LINKED ISSUE #5: Card Search
As a user, I would like to be able to search for cards, and quick add them to my cube.

I would like a separate view where I can search for cards with scryfall search syntax, and click on cards to add them to a cube. I would like to be able to click on cards from other user's cubes, and quick add them to one of my cubes as well.

dekkerglen:
Fixes #5

Also adds elo and embedding to local carddb, we will need to update the calls to the embeddings route to just use carddb embeddings instead

New card search:
image

New card page:
image
image
image
image
image
image


routes/root.js
```diff
@@ -316,6 +316,8 @@ router.get('/search/:query/:page', async (req, res) => {
 
     const count = await Cube.countDocuments(query);
 
+    console.log(JSON.stringify(query, null, 2));
```
ruler501, line 319: console statement.

routes/tools_routes.js
```diff
@@ -19,7 +19,7 @@ const router = express.Router();
 /* Minimum number of picks for data to show up in Top Cards list. */
 const MIN_PICKS = 100;
 /* Page size for results */
-const PAGE_SIZE = 100;
+const PAGE_SIZE = 96;
```
ruler501, line 22: This seems like a strange number for page size.

routes/tools_routes.js
```diff
@@ -112,6 +112,72 @@ async function topCards(filter, sortField = 'elo', page = 0, direction = 'descen
   };
 }
 
+// sorts: ['elo', 'date', 'price', 'alphabetical']
+// direction: ['ascending', 'descending']
+// distinct: ['names', 'printings']
+
+const sortFunctions = {
+  elo: (direction) => (a, b) => {
+    const factor = direction === 'ascending' ? 1 : -1;
```
ruler501, line 121: `factor * (a.elo - b.elo)`

same for the others below.

PR #1559, opened 2020-08-21
Some small tweaks to make bots more sensitive to land counts.

ruler501:


ruler501:
This will get updated after the draft overhaul is merged.

ruler501:
Rebuilding this in another PR with a more in-depth rework of the draftbot code.

PR #1568, opened 2020-08-25
autocomplete tag filter in advanced filter modal

### LINKED ISSUE #370: Advanced Filter Tags Autocomplete
As a user, I would like the tags to autocomplete based on my cubes tags, for the advanced filter modal.

wirmar:
Also removed tags filter from advanced filter modal where no tags are present (for example "Search Cards")
fixes #370

Tag autocomplete screenshot

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -27,6 +27,7 @@ const {
   replaceCardHtml,
   abbreviate,
   buildTagColors,
+  cubeCardTags,
   maybeCards,
   getElo,
   CSVtoCards,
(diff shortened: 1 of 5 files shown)


routes/cube_routes.js
```diff
@@ -2633,6 +2634,19 @@ router.get(
   }),
 );
 
+router.get(
+  '/api/cubecardtags/:id',
+  util.wrapAsyncApi(async (req, res) => {
+    const cube = await Cube.findOne(buildIdQuery(req.params.id)).lean();
+    const tags = cubeCardTags(cube);
```
RKelln, line 2641: @dekkerglen this seems a bit painful to do each request, would this make sense to do once and stored when a cube is saved/updated?
Maybe safest/simplest to do the slow way though.

RKelln:
Looks good to me, although I wonder if the tags should be something stored instead of recalculated (see comment).

dekkerglen:
How does this handle a non-cube context, like top cards?

routes/cube_routes.js
wirmar: I'll take a look of this route is called too often, but it probably shouldn't be right now and it probably is not too expensive I would think?
Caching the value would add a lot of complexity, where I would guess it is not worth for the potential benefit it might bring?

wirmar:
@dekkerglen In a non-cube context the `cubeID` from the `CubeProvider.Consumer` will be null, so the Tags input field will not be rendered. This makes sense anyway, since there are no tags on which to filter in non-cube contexts.

PR #1570, opened 2020-08-25
React full

dekkerglen:
This PR moves all pages to 100% react


serverjs/render.js
```diff
@@ -0,0 +1,79 @@
...
+  pages.LoginPage = require('../dist/pages/LoginPage').default;
+  pages.RegisterPage = require('../dist/pages/RegisterPage').default;
+  pages.LostPasswordPage = require('../dist/pages/LostPasswordPage').default;
+  pages.NotificationsPage = require('../dist/pages/NotificationsPage').default;
+  pages.PasswordResetPage = require('../dist/pages/PasswordResetPage').default;
+  pages.UserAccountPage = require('../dist/pages/UserAccountPage').default;
+  pages.UserBlogPage = require('../dist/pages/UserBlogPage').default;
+  pages.UserDecksPage = require('../dist/pages/UserDecksPage').default;
+  pages.UserSocialPage = require('../dist/pages/UserSocialPage').default;
+  pages.UserCubePage = require('../dist/pages/UserCubePage').default;
```
phulin, line 47: this could be cleaner but I'm not sure if we care

serverjs/render.js
```diff
@@ -0,0 +1,79 @@
...
+        image_name: req.user.image_name,
+        artist: req.user.artist,
+      }
+    : null;
+
+  return res.render('main', {
+    reactHTML:
+      NODE_ENV === 'production' ? ReactDOMServer.renderToString(React.createElement(getPage(page), reactProps)) : null,
+    reactProps: serialize(reactProps),
+    loginCallback: req.path,
```
phulin, line 74: Do we still need this? Might be a good time to remove.

serverjs/render.js
```diff
@@ -0,0 +1,79 @@
...
+        about: req.user.about,
+        image: req.user.image,
+        image_name: req.user.image_name,
+        artist: req.user.artist,
+      }
+    : null;
+
+  return res.render('main', {
+    reactHTML:
+      NODE_ENV === 'production' ? ReactDOMServer.renderToString(React.createElement(getPage(page), reactProps)) : null,
```
phulin, line 72: Moving every page to SSR will increase server load. It's probably fine, but we should just be aware. Now that we're doing full react, we might be able to switch to `renderToNodeStream` which will be much more performant. I can think about how to do that if you want. It is possible to move the HTML header into react too by using stuff like Helmet.

dekkerglen: Server CPU is basically never the bottleneck, so I think this will be alright.

joshfinnie:
this seems like a huge step in the right direction... I could see a few areas of improvement (maybe?), like using a global context for user data or small things like that, but nothing that would stop this PR. valiant effort!

dekkerglen:
> this seems like a huge step in the right direction... I could see a few areas of improvement (maybe?), like using a global context for user data or small things like that, but nothing that would stop this PR. valiant effort!

Adding a context for users would be a great improvement, I'll add a separate issue to address that

phulin:
the current nested context system sucks, and ideally we’d be using redux or
something.... but at this point we’re probably stuck with it.

PR #1574, opened 2020-08-27
Draft format overhaul

### LINKED ISSUE #838: Custom draft format - predetermined cards
This is probably pretty niche, but could we get an option on custom draft formats to start a players pool with a predetermined set of cards? In our Commander Cube, every player starts with a Command Tower and Arcane Signet in their pool

Posted in proxy for 'zachtib' on discord

### LINKED ISSUE #1494: Drafting is broken at master with slightly broken draft synergies support
**Describe the bug**
Starting a draft at current master branch fails as the code at https://github.com/dekkerglen/CubeCobra/blob/master/src/utils/Draft.js#L37 attempts to access `synergies` which is null, even though it is marked as a required PropType. This seems like a fault at current implementation which seems to be still partly in development.

**To Reproduce**
1. Checkout master 4418d213 commit
2. Go to your cube, start playtesting and begin draft
3. Draft errors at addSeen()

**Expected behavior**
Drafting works without errors at master.

### LINKED ISSUE #1547: Update embeddings to not call ML server
Card embeddings now exist in carddb, we shouldn't have to call the ML server every time we start a draft.

### LINKED ISSUE #1565: Add support for 'supreme draft'
Description of the format: https://magic.wizards.com/en/articles/archive/magic-online/introducing-supreme-draft-2020-08-24

ruler501:
Makes the draft format much more compact by replacing most cardSchemas with Numbers.
Enables adding wastes to decks.
Rework the custom draft format model.
Rework the pack model.
Allow specifying a number of cards to get trashed at the end of a pack.
Allow marking specific packs as sealed.
Allow specifying how many cards to pick at a time.
Migrate database objects on the fly so we don't have to deal with running a one_shot_script that would entail downtime.

Still to do:
Allow specifying how many cards to pick for your first pick from a pack.(2XM format)

Depends on #1789 since it wasn't possible to test the DeckPage without the fix from it.

Fixes #838 
Fixes #1494 
Fixes #1547 
Effectively fixes #1565


.eslintignore
```diff
@@ -1,6 +1,20 @@
...
+one_shot_scripts/add_blog_names.js
+one_shot_scripts/add_deck_names.js
+one_shot_scripts/generate_short_ids.js
+one_shot_scripts/add_deck_owners.js
+one_shot_scripts/fix_cube_deck_count.js
+one_shot_scripts/fix_deck_count.js
+one_shot_scripts/import_ratings.js
+one_shot_scripts/remove_invalid_tags.js
+one_shot_scripts/seed.js
+one_shot_scripts/validate_colors.js
```
dekkerglen, line 17: I'm not a fan of adding new files to ignore, we should only ever remove files from this if we can

models/draft.js
```diff
@@ -1,31 +1,55 @@
...
 
+// Details on each pack, how to draft and what's in it.
+const Pack = {
+  cards: [
+    {
+      type: Number,
+      min: 0,
+    },
+  ],
+  pickAtTime: {
```
dekkerglen, line 12: not a fan of this name. Maybe `cardsPerPick` ?

one_shot_scripts/update_deck_format.js
```diff
@@ -0,0 +1,74 @@
...
+    seat.sideboard = (seat.sideboard || []).map((pack) =>
+      pack.map((card) => {
+        cards.push(card);
+        return cards.length - 1;
+      }),
+    );
+    seat.pickorder = (seat.pickorder || []).map((card) => cards.findIndex((card2) => card.cardID === card2.cardID));
+  }
+  deck.cards = cards.concat(basicsList);
+  return deck;
```
dekkerglen, line 35: needs better error handling. There is so much errored data across the DB, check out the old deck migration script for reference. I need to check a lot of fields to see if they were objects, primitives, or arrays all the time

one_shot_scripts/update_deck_format.js
```diff
@@ -0,0 +1,74 @@
...
+        if (deck) {
+          decks.push(migrateDeck(deck));
+        }
+      }
+    }
+    // eslint-disable-next-line no-await-in-loop
+    const operations = (await Promise.all(decks))
+      .filter((deck) => deck)
+      .map((deck) => ({
+        replaceOne: {
```
dekkerglen, line 60: why replaceOne? Why not just use `.save()` ? Is this to b/c you think the lean queries are necessary? If that's the case I think the better practice would be to be better about how we are modifying the object such that we only have a couple assignments at the end of the function

one_shot_scripts/update_draft_format.js
```diff
@@ -0,0 +1,61 @@
...
+        // eslint-disable-next-line no-await-in-loop
+        await Draft.bulkWrite(operations);
+      }
+      console.log(`Finished: ${i + batchSize} of ${count} drafts`);
+    }
+
+    mongoose.disconnect();
+    console.log('done');
+  });
+})();
```
dekkerglen, line 61: same feedback as update deck format

src/components/CustomDraftFormatModal.js
```diff
@@ -30,6 +31,7 @@ const CustomDraftFormatModal = ({ isOpen, toggle, formatIndex, format, setFormat
   const { cubeID } = useContext(CubeContext);
 
   const formRef = useRef();
+  const defaultPack = { filters: [''], trash: 0, sealed: false, pickAtTime: 1 };
```
dekkerglen, line 34: Again, not a fan of `pickAtTime`. Maybe `cardsPerPick` ?

src/components/CustomDraftFormatModal.js
```diff
@@ -142,7 +174,43 @@ const CustomDraftFormatModal = ({ isOpen, toggle, formatIndex, format, setFormat
...
+                      className="mr-2 ml-2"
+                      value={pack.pickAtTime}
+                      min={1}
+                      max={pack.filters.length}
+                      onChange={handleChangePickAtTime}
+                      data-index={index}
+                    />
+                    cards at a time.
+                  </Label>
+                </Form>
```
dekkerglen, line 212: Can you provide screenshots of before and after for the UI changes?

models/draft.js
ruler501: how about cardsPerPass?

one_shot_scripts/update_deck_format.js
ruler501: I'm going to strip out the migrations and just do it on the fly.

models/cube.js
```diff
@@ -190,7 +200,7 @@ cubeSchema.index({
   numDecks: -1,
 });
 
-const Cube = mongoose.model('Cube', cubeSchema);
+const Cube = mongoose.model('Cube', withMigrations(cubeSchema, cubeMigrations));
```
dekkerglen, line 203: I don't like this pattern - I'd much rather have a one shot script to update the database

dekkerglen (changes requested):

models/cube.js
```diff
@@ -126,7 +128,14 @@ const cubeSchema = mongoose.Schema({
         multiples: Boolean,
         html: String,
         markdown: String,
-        packs: String,
+        packs: [
+          {
+            filters: [String],
+            trash: { type: Number, min: 0 },
+            sealed: Boolean,
```
dekkerglen, line 135: sealed doesn't make sense as a pack variable, should be as a draft format property.

Also having picksPerPass and trash feels like a very weak way to customize each pack. I think we should instead have a 'steps' definition that looks something like (example is supreme draft):
`
steps: [
   {
      type: pick,
      count: 2,
   },
   {
       type: trash,
       count: 13,
   }
]
`

This would be more flexible and allow things like only the first pick is double. We would need to add validation to make sure that the numbers match up, but that wouldn't be too difficult.

models/cube.js
```diff
@@ -1,5 +1,7 @@
 const mongoose = require('mongoose');
 
+const CURRENT_SCHEMA_VERSION = require('./migrations/deckMigrations').slice(-1)[0].version;
```
dekkerglen, line 3: importing and then slice, and then taking index? there is no good reason why we should be doing this. Migrations should be a one shot, disposable script. No need to make these patterns that are nonsensical if you aren't familiar with the process

models/draft.js
```diff
@@ -1,30 +1,66 @@
...
+      type: Number,
+      min: 0,
+    },
+  ],
+  picksPerPass: {
+    type: Number,
+    default: 1,
+    min: 1,
+  },
+  sealed: {
```
dekkerglen, line 19: sealed doesn't make sense here

models/draft.js
```diff
@@ -1,30 +1,66 @@
 const mongoose = require('mongoose');
+
 const cardSchema = require('./cardSchema');
+const CURRENT_SCHEMA_VERSION = require('./migrations/draftMigrations').slice(-1)[0].version;
+
+// Details on each pack, how to draft and what's in it.
+const Pack = {
```
dekkerglen, line 7: I'd like to see this reflect my earlier comment about making a more flexible custom draft schema

models/migrations/migrationMiddleware.js
```diff
@@ -0,0 +1,26 @@
+const applyPendingMigrationsPre = (migrations) => async (doc) => {
```
dekkerglen, line 1: I don't think we need this file if we're doing this with the one-shot style

one_shot_scripts/apply_migrations.js
```diff
@@ -0,0 +1,85 @@
...
+
+const MIGRATABLE = Object.freeze([
+  { name: 'Cube', model: Cube, migrate: applyPendingMigrationsPre(cubeMigrations) },
+  { name: 'Deck', model: Deck, migrate: applyPendingMigrationsPre(deckMigrations) },
+  { name: 'Draft', model: Draft, migrate: applyPendingMigrationsPre(draftMigrations) },
+  { name: 'GridDraft', model: GridDraft, migrate: applyPendingMigrationsPre(gridDraftMigrations) },
+]);
+
+const migratableDocsQuery = (currentSchemaVersion) => ({
+  $or: [{ schemaVersion: { $exists: false } }, { schemaVersion: { $lt: currentSchemaVersion } }],
```
dekkerglen, line 24: this will considerably slow down the query, and we should probably just do this check after the docs have been loaded

routes/cube_routes.js
```diff
@@ -179,24 +180,26 @@ router.post('/format/add/:id', ensureAuth, async (req, res) => {
...
       }
       cube.draft_formats.push({
-        title: req.body.title,
-        multiples: req.body.multiples === 'true',
-        markdown: req.body.markdown.substring(0, 5000),
-        packs: req.body.format,
+        title,
+        multiples: multiples === 'true',
+        markdown: markdown.substring(0, 5000),
+        packs: JSON.parse(format),
```
dekkerglen, line 194: we need validation on format

serverjs/cards.js
```diff
@@ -127,8 +127,11 @@ function initializeCardDb(dataRoot, skipWatchers) {
 }
 
 function unloadCardDb() {
-  for (const attribute of Object.values(fileToAttribute)) {
+  for (const [filename, attribute] of Object.entries(fileToAttribute)) {
     delete data[attribute];
+    if (skipWatchers !== true) {
```
dekkerglen, line 132: skip watchers shouldn't permanently unwatch the file?

src/components/DecksPickBreakdown.js
```diff
@@ -74,7 +74,17 @@ const DecksPickBreakdown = ({ draft, seatIndex, deck, defaultIndex }) => {
 
 DecksPickBreakdown.propTypes = {
   draft: PropTypes.shape({
```
dekkerglen, line 76: we should use a DraftPropType for this validation

dekkerglen (changes requested):

models/cube.js
RKelln: I think per pack sealed is fine, where the sealed format is just all packs sealed. Pack specificity allows for format flexibility (for example a commander draft where you open a sealed pack of commanders before the rest of the packs).

That being said, `{type: pick, count: N}` where N is the number cards in the pack is equivalent to sealed and even more generic. In either case we likely need special UI handling when sealed or picks == number of cards in pack (either an add all cards button or skip the step and add all cards to pool automatically).

In general I advocate for generic backend, but premade templates (e.g. supreme draft checkbox, etc) for frontend UI.

src/components/DecksPickBreakdown.js
ruler501: Done in #1798

serverjs/cards.js
ruler501: When unloading we need to stop watching because otherwise they sawn reload and the threads watching will never stop. We don't actually have access to skipWatchers here so in #1798 I just wrapped it in a try catch.

routes/cube_routes.js
ruler501: I'm working on adding that.

one_shot_scripts/apply_migrations.js
ruler501: Seems fine.

models/draft.js
ruler501: Done in #1798

models/cube.js
ruler501: I can make it more intuitive, but I'd like a system for tracking the schema versions of documents.

models/cube.js
ruler501: I started implementing steps off the discord discussion before seeing this so the schema is
```js
{
                  action: {
                    type: String,
                    enum: ['pass', 'pick', 'trash', 'pickrandom', 'trashrandom'],
                  },
                  amount: {
                    type: Number,
                    default: 1,
                  },
                },
```
Amount defaults to 1 and is ignored for pass (open to changing this to allow for really esoteric formats with positive and negative allowing for customizing direction. It would require very little change to the code). 
I intend to add a feature that gives you a pick/trash all button for when the trash or pick amount is equal to the rest of the pack.

I do think UI shortcuts to make defaults would be good. I'm thinking that per pack you could autofill: Supreme, Sealed, Pick N at a time, Trash last N, 2XM, and maybe more if we get requests.

ruler501:
This is superceded by #1798. I'll push up the changes unifying them soon.

Issue #1567, opened 2020-08-24
Card Pages for new cards don't load


dekkerglen:
When new cards are spoiled, the respective card pages should be accessible with empty data. Currently, they don't load

danielprue:
I'm going to work on this

PR #1580, opened 2020-09-02
Fixed issue that causes newly revealed cards to not load card page

danielprue:
Bug fix for issue #1567 

Previous behavior: Cards that had scryfall data but no cardHistory data would not generate a card page, but instead redirect to 404. This is primarily an issue with newly revealed cards.

New behavior: Cards that have scryfall data but no cardHistory are now given a default empty data object, which gives enough information to generate and view the scryfall data on the page before cardHistory is added.

I am pretty new to this type of work flow, and I've never worked on a project this large, so I'm pretty open to feedback. Particularly, the big data object in the middle of the route feels a bit out of place to me. Is there a better location/format for this object, or is this fine organization?

### DIFF
--- a/routes/tools_routes.js
+++ b/routes/tools_routes.js
@@ -298,15 +298,69 @@ router.get('/card/:id', async (req, res) => {
     }
 
     // otherwise just go to this ID.
-    const data = await CardHistory.findOne({ oracleId: card.oracle_id });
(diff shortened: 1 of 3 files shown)


danielprue:
I manually tested this against a few of the newly revealed Zendikar Rising cards, but I'm re-reading the PR requirements and realizing I haven't written/updated any tests for it. Will work on that later today.

Issue #1578, opened 2020-09-01
Card Pages doesn't filter out current version correctly


dekkerglen:
Both Jace TMS pages:
https://cubecobra.com/tool/card/ec386bc3-137b-49b5-8380-8daff470f0bc
https://cubecobra.com/tool/card/a5d1f66f-75b5-4a9b-8b98-f237f065de9c

In the 2nd one, in the 'versions' section, it's missing the 2xm version, and it also shows the  current version in the list

danielprue:
The card versions displayed in the 'versions' section here are pulled from the cardhistories table in the database, which is populated by mapping the card's oracle id against the oracleToId.json fixture. I cannot replicate the bug with the missing versions locally, which makes me think that there is a problem with how the production db is populated. 

I wrote a fix to the other half of the issue, which filter's out the version currently being viewed, and displays a message if there are no other versions of the card, so I will make a PR for that code, but I think the missing version will take some deeper investigation into the production environment.

PR #1583, opened 2020-09-09
Fixed issue that showed current card version in the other versions list on Card Page

danielprue:
Partial bug fix for issue #1578 

Previous behavior: The versions section of a card page would display the current version of the card.

New behavior: The versions section now only shows other versions of the card. If there are no other versions of the card, a message now appears, indicating that there are no other versions to view

I was unable to replicate the issue where versions were missing in my development environment. I left a note in the issue about my investigation on the issue.


dekkerglen:
There are some CI failures

danielprue:
CI now passing all checks

Issue #1598, opened 2020-09-12
Peasant Cube Analytics appear broken


dekkerglen:
Via discord:
> I'm happy to see the peasant category in the card analytics page, but every card I look at shows up in 0% of peasant cubes (even cards I've confirmed exist in listed peasant cubes, like lightning bolt).
(in the play rate tab)

danielprue:
I found the bug fix for this, and will add a pull request soon, but I also noticed that the category detector doesn't automatically assign the peasant category as it does with pauper. Would like to also fix that in the same PR.

PR #1609, opened 2020-09-15
Fixed issue which caused peasant cube count to not display in card play rate analytics

danielprue:
Bug fix for #1598 

Previous behavior: Under the Play Rate tab of a card page, the peasant cube count always displays 0, even for cards that are in peasant cubes.

New behavior: Peasant category added in the cardhistories model. On the next card history update, the database will now populate with information about how many peasant cubes that card is in.

I also added behavior that will autodetect peasant cubes as cubes that are not pauper legal, but only contain common or uncommon cards and adds peasant to the type line. This is similar to logic that already exists for pauper cubes.

### DIFF
--- a/models/cardHistory.js
+++ b/models/cardHistory.js
@@ -20,6 +20,7 @@ const Datapoint = {
   size540: [Number],
   size720: [Number],
   pauper: [Number],
+  peasant: [Number],
   legacy: [Number],
   modern: [Number],
   standard: [Number],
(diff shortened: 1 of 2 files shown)


dekkerglen:
BTW are you on discord? That's where most of the dev discussion occurs
https://discord.gg/36nfm8c

danielprue:
Yep, I am on the discord already! This is the largest project I've worked on before, so I mostly lurk and and try to pick up some of the easier looking bug fixes, but I'll try to comment more in there as I get more comfortable with the project.

dekkerglen:
Do you have the contributor role? That unlocks the dev channels

danielprue:
Yep! My discord nickname should be set to the same as my github

PR #1611, opened 2020-09-15
Add Metadata to content

dekkerglen:
also fixes an issue where blogposts can't be added to changelogs


serverjs/render.js
```diff
@@ -82,6 +82,16 @@ const render = (req, res, page, reactProps = {}, options = {}) => {
 
   reactProps.loginCallback = req.baseUrl + req.path;
 
+  if (!options.metadata) {
+    options.metadata = [];
+  }
+  if (options.metadata.filter((data) => data.property === 'og:image').length === 0) {
```
lunakv, line 88: `!options.metadata.some(...)` might be a more semantically correct check?

serverjs/render.js
```diff
@@ -82,6 +82,16 @@ const render = (req, res, page, reactProps = {}, options = {}) => {
...
+    options.metadata.push({
+      property: 'og:image',
+      content: '/content/sticker.png',
+    });
+  }
+
   return res.render('main', {
```
lunakv, line 95: Replacing a falsy `options.metadata` with `null` is redundant, because, since the template does a truthiness check anyway.

models/cube.js
```diff
@@ -189,6 +189,6 @@ const Cube = mongoose.model('Cube', cubeSchema);
 
 Cube.LAYOUT_FIELDS = '_id owner name type card_count overrideCategory categoryOverride categoryPrefixes';
 Cube.PREVIEW_FIELDS =
-  '_id shortId urlAlias name card_count type overrideCategory categoryOverride categoryPrefixes image_name image_artist image_uri owner owner_name';
+  '_id shortId urlAlias name card_count type overrideCategory categoryOverride categoryPrefixes image_name image_artist image_uri owner owner_name image_uri';
```
lunakv, line 192: `PREVIEW_FIELDS` was modified, but `LAYOUT_FIELDS` is actually used for database connections in `cube_routes`.

lunakv: `PREVIEW_FIELDS` also already contains 'image_uri'

serverjs/render.js
lunakv: (the actual check is actually below the snippet, because it's too far from the changes, but I figured it was worth mentioning)

src/components/EditCollapse.js
```diff
@@ -61,13 +76,17 @@ const EditCollapse = ({ ...props }) => {
...
-  const handleChange = useCallback((event) => {
-    return {
-      blog: setPostContent,
-      add: setAddValue,
-      remove: setRemoveValue,
-    }[event.target.name](event.target.value);
-  }, []);
+  const handleChange = useCallback(
+    (event) => {
+      console.log(event.target.name);
```
RKelln, line 81: You want the console.log to remain?

PR #1616, opened 2020-09-17
[WIP] Dark mode

### LINKED ISSUE #686: Dark or "nighttime" theme for the site
A lot of people find dark website themes easier to look at for long periods of time (such as drafting). A darker theme for the site would be a great addition.

omniczech:
Fixes #686

Tagged as WIP until folks have had time to give feedback on color choices and review both dark and light modes. Appreciate poking around weird corners of the site as I tried to find every single view but want to be sure.


dekkerglen:
image
The background of this image doesn't match the banner

image
I think there is too little contrast with this text. Also, the artist name is unreadable.. maybe keep that the way it was before?

image
The 'podcast' text is unreadable. The title of these preview objects need more contrast

image
This line looks out of place

image
I think that the card names should be a little darker, less contrast

image
This lights up when being used

image
This lights up when being used

image
Uh oh!

image
React pivot table styles need overwriting

image
Results is not readable

image
I think this would look better with white text

image
This vertical line looks wrong

image
Dark mode scroll bars would be nice 

image
This lights up when being used

image
image
Markdown comment blocks color is incorrect

image
Background color of card names in draftbot analysis is wrong

image
`code` highlighting doesn't have enough contrast with background

omniczech:
> image
> The background of this image doesn't match the banner

This is an image where the background is the only area with non-transparent pixels, does this exist in a vector file format to allow us to color the fill on a theme needed basis

> image
> I think there is too little contrast with this text. Also, the artist name is unreadable.. maybe keep that the way it was before?

Reverted
 
> image
> The 'podcast' text is unreadable. The title of these preview objects need more contrast

Is there a convenient way to get some demo-data for testing this locally?

> image
> This line looks out of place

fixed

> image
> I think that the card names should be a little darker, less contrast

Brought this down a bit

> image
> This lights up when being used

The select element has been darkened when focused and on linux firefox this took care of the actual options, but I'd love verification that this isn't overridden in other browsers

> image
> This lights up when being used

fixed

> image
> Uh oh!

not sure how my branch was missing the width css, fixed

> image
> React pivot table styles need overwriting

cleaned this up, please check. also removed weird serif font

> image
> Results is not readable

upped contrast

> image
> I think this would look better with white text

hardcoded to always be white

> image
> This vertical line looks wrong

fixed 

> image
> Dark mode scroll bars would be nice

Is this browser defined? I'm getting dark in my local.

> image
> This lights up when being used

fixed

> image
> image
> Markdown comment blocks color is incorrect

I'm currently getting an error re: the markdown object being undefined after several attempts to `npm i` or even `npm ci`

> image
> Background color of card names in draftbot analysis is wrong

Where is this page?

> image
> `code` highlighting doesn't have enough contrast with background

I'm not stoked on the solution I've got in there, but take a look.

Issue #1601, opened 2020-09-12
Ongoing: Remove files from eslint ignore


dekkerglen:
We have a lot of files in eslint ignore, for no reason other than convenience. We should remove as many files as possible from there. For this story, feel free to make a PR that just addresses one of the files.

PR #1618, opened 2020-09-21
Style Improvements for Sort.js

jbts:
Working on #1601 for `src/utils/Sort.js`.

Most of the changes I've made are small, like replacing `==` with `===` or `var` with `let`.
Some changes are a little more involved, like renaming a variable to use camelCase or removing an unused parameter from a function.

Ideally these changes just affect the style of the code and not the code execution/logic.

There are a couple tricky style issues left that I'm not sure how to handle. I'll see if I can highlight them with some comments.

### DIFF
--- a/src/utils/Sort.js
+++ b/src/utils/Sort.js
@@ -716,8 +716,8 @@ export function cardGetLabels(card, sort) {
   }
   if (sort === 'Legality') {
     return Object.entries(card.details.legalities)
-      .filter(([k, v]) => ['legal', 'banned'].includes(v))
-      .map(([k, v]) => k);
+      .filter(([k, v]) => ['legal', 'banned'].includes(v)) // eslint-disable-line no-unused-vars
+      .map(([k, v]) => k); // eslint-disable-line no-unused-vars
   }
   if (sort === 'Power') {
     if (card.details.power) {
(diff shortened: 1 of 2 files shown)


src/utils/Sort.js
```diff
@@ -645,158 +679,179 @@ export function cardGetLabels(card, sort) {
...
+    ];
+  }
+  if (sort === 'Artist') {
     return [card.details.artist];
-  } else if (sort == 'Legality') {
+  }
+  if (sort === 'Legality') {
     return Object.entries(card.details.legalities)
       .filter(([k, v]) => ['legal', 'banned'].includes(v))
       .map(([k, v]) => k);
```
jbts, line 700: Here one of the functions doesn't use the key `k` and the other doesn't use the value `v`.

jbts:
I can't seem to comment in the same way for the other style issue, but it's on line 906. The first argument `label` isn't used by the function.

src/utils/Sort.js
jbts: Alright. I'll plan to add a comment to tell eslint to ignore those errors then.

dekkerglen: You can replace the variable name with `_` and it will suppress this error iirc

jbts: That hasn't been working for me, still getting the same error with `_` and no comment.

Issue #1612, opened 2020-09-15
Add centering to markdown


dekkerglen:
As a user, I would like to be able to center content using markdown

PR #1625, opened 2020-10-02
Add centering to markdown

xorg:
Implements #1612 

This PR adds the possibility to center elements using markdown.

To center a text, just wrap it in angular brackets >>>  <<<

The markdown wraps the centered text in <span> tags with the "centered" class.

### DIFF
--- a/src/components/MagicMarkdown.js
+++ b/src/components/MagicMarkdown.js
@@ -296,9 +296,7 @@ const OuterMarkdown = ({ markdown, limited }) => {
           );
         }
         if (section.startsWith('> ')) {
-          console.log(section);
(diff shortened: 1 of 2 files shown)


src/components/MagicMarkdown.js
```diff
@@ -309,6 +309,17 @@ const OuterMarkdown = ({ markdown, limited }) => {
             </Card>
           );
         }
+        if (section.startsWith('>>>')) {
+          console.log(section);
```
SevenIndirecto, line 313: Sorry to jump in here as I'm not a reviewer, but should probably remove the console output.

xorg: You're right, I just removed it.

dekkerglen:
Looks great! Thanks for taking the time to add to the markdown guide as well!

PR #1644, opened 2020-10-14
Markdown centering fix

### LINKED ISSUE #1612: Add centering to markdown

### LINKED ISSUE #1638: Tweaks needed for markdown
`>>> **Esper**- Flying **Grixis**- Direct Damage <<<`

This should leave white space between flying and grixis, but doesn't

lunakv:
Fixes #1612 
Fixes #1638 
Addresses issues introduced in #1625 .

*Warning*: This is a long comment. It may not need to be, but I just spent four hours on what amounts to changing six lines, so I'm going to write about this damn it. If you aren't interested and just want to see how it works, there are pictures at the end.

The current way centering is written in the markdown parser is flawed in several ways. Firstly, there are some code issues this PR fixes:
- The '>>>' parser includes the `lines = section.split().filter()` line followed by `lines.map` in the actual results. I assume this was copied from the '> ' parser. This isn't necessarily a bug, because the regex on which we're splitting cannot ever match anything, meaning `lines` will always contain just one element with the whole section, but it doesn't really make any sense (also, the centered section doesn't need to be split into lines at all).
- The centered section is wrapped in a `span` element. I believe that the centered section is a block-level element, not an inline one (supported by the fact that it's parsed in OuterMarkdown), which makes `div` a more appropriate choice.

Secondly, and rather more importantly, however, the current solution has some functional issues, mostly on the CSS side.

## How it works currently.
The `span` is displayed as horizontal flexbox container, which centers all the containers horizontally and justifies them to the center vertically.

### The problems with this approach:
1. **It doesn't work with other markdown.** Any InnerMarkdown element (such as piece of bold text, an image, or a link) will create its own flexbox container, splitting the line into multiple separated fragments. That means this source:
bad_boxes_src  
Turns into this result:
bad_boxes_res 
which I'd argue isn't the desired outcome.
2. **It doesn't work with other markdown.** The behaviour explained in the previous point is also the root cause of #1638 , because the bounding boxes by themselves have no padding/margin and whitespace is trimmed by default.
3. **It doesn't actually centre text.** This issue is best seen in a multiline centred block. Here's the source:
not_centred_src
and here's the result:

This is clearly not a centre-justified paragraph of text. The reason for this is simple - the flexbox layout doesn't actually centre text. Instead, it centres the containers inside it across the row (but doesn't affect the layout inside those containers themselves). This works when all we have is a single line of simple text - a single container is created in the middle with the width of the text inside it. But it breaks down when introducing multiple lines (or indeed multiple children, as seen above).

## What this PR does
We can realize that centring content isn't as difficult as it may seem. With the exception of card images and block LaTeX, everything is displayed inline by default, so just setting `text-align: center;` on the surrounding div gets us most of the way there. Block LaTeX also centers itself by default, so we don't need to do anything about it either. The only issue left to resolve is centering card images.

Those also turned out surprisingly simple to solve. Since card images are generated as a link inside their own div, we just have to set `margin: auto;` on the div encompassing that link and it will happily move its contents to the center for us. Problem solved.

Oh, and as a side note, the PR also fixes an issue I found where lines containing only ">\n" without a trailing space broke quotation blocks.

#### Result:

(You can even center lists if you're into that, it works fine, they just didn't fit on the screen.)
I confirmed this to be working locally in cube overviews and blog posts. I didn't look at content articles, and it doesn't work on comments, which don't allow any formatting with `>` or `<`.

### DIFF
--- a/public/css/stylesheet.css
+++ b/public/css/stylesheet.css
@@ -451,9 +451,12 @@ body.busy-cursor {
 }
 
 .centered {
-	display: flex;
-	justify-content: center;
-	align-items: center;
+	width: 100%;
+	text-align: center;
+}
+
+.centered > div {
+	margin: auto;
 }
 
 .centered-max400 {
--- a/src/components/BlogPost.js
+++ b/src/components/BlogPost.js
@@ -84,7 +84,7 @@ class BlogPost extends React.Component {
         <div style={{ overflow: 'auto', maxHeight: '50vh' }}>
           {post.changelist && (post.html || post.markdown) ? (
             <Row className="no-gutters">
-              <Col className="col-12 col-l-5 col-md-4 col-sm-12 blog-post-border" >
+              <Col className="col-12 col-l-5 col-md-4 col-sm-12 blog-post-border">
                 <CardBody className="py-2">
                   <CardText dangerouslySetInnerHTML={{ __html: post.changelist }} />
                 </CardBody>
--- a/src/components/MagicMarkdown.js
+++ b/src/components/MagicMarkdown.js
@@ -15,11 +15,15 @@ const Link = withModal('a', LinkModal);
 const InnerMarkdown = ({ markdown }) => {
   const markdownStr = markdown.toString();
   const split = markdownStr.split(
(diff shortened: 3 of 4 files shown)


lunakv:
Adding a preview of how it looks on mobile:

dekkerglen:
This is fantastic. Loved the detailed breakdown, makes the review process much easier for me!

PR #1654, opened 2020-10-20
Fixed cube type determination

### LINKED ISSUE #1653: Cube type doesn't update correctly
Create a new cube. Add a standard legal card to it, e.g. Island. The cube type is updated from `Standard Pauper Cube` to `Vintage (Pauper) Cube`, despite all cards in it being standard legal. The type returns to `Standard` when all cards from the cube are removed.

The pauper/peasant classifications seem to be working correctly.

lunakv:
Fixes #1653 

## Cause of issue
The problem, unsurprisingly, was in the following expression: `!cardIsLegal(carddb.cardFromId(card.cardID), intToLegality(type).toLowerCase())`. The `legalities` object checked by `cardIsLegal` has the first letter of each key capitalized,  but the parameters we passed to it were converted to lowercase. That meant we were checking the value of e.g. `legalities['standard']`, which doesn't exist (only `legalities['Standard']` does), which returned `undefined` every time and the legality check always failed on the first card until we got stuck with Vintage as the fallback format.

## Fix
Removed the `.toUpperCase()` call from the parameter. Since `cardIsLegal` is a private function used only here and nowhere else, this won't break anything. I also made the surrounding code a little prettier.


lunakv:
Looks like the issue stems from the fact that while Scryfall card objects have the legality keys all lowercase, we capitalize the first letters when converting them (see `convertLegalities` from `serverjs/updatecards.js`). The tests didn't catch this, because the mock objects were also using all lowercase keys everywhere. This PR now also fixes those objects.

PR #1665, opened 2020-10-23
Single Definitions for Complex Prop Types

### LINKED ISSUE #1575: Create complex object proptypes for consistent validation
We currently define many complex objects, like cubes, blogposts, user, in proptypes on every file. Lots of duplicated code, we should create a proptypes directory and be able to include any object we use that way.

jbts:
Fixes #1575 

Moving complex prop type definitions from files in `src/components/` to a new `proptypes/` directory.

There are often several definitions of the same prop type (like an article) in the `src/components/` directory. These don't always agree exactly, which can make creating a single definition more complicated.


jbts:
Last comment is no good.

The `id` attribute is the session id derived from the `_id` attribute of the user database object, if I understand correctly. I will need to revert the changes I have made to the `id` attributes of user prop types.

`UserPropType` itself should have an `id` attribute instead of an `_id` attribute. I also need to be careful not to replace something like the type of a `followers` attribute with `UserPropType` since `UserPropType` represents a current user.

jbts:
I believe I've fixed the mistakes I made related to `user.id` vs. `user._id` confusion. This has also reduced the number of files modified a little bit and simplified the changes that I've made.

There are a few console warnings that I'm getting about failed prop type validations, where an attribute is required but not provided. The `user.about` and `user.users_following` attributes may need to be changed so that they are not required.

dekkerglen:
@jbts One conflict on CardPage, but looks good otherwise!

PR #1667, opened 2020-10-24
Add default Elo value to card objects

### LINKED ISSUE #1655: Null Elo should default to 1200
For reference: https://cubecobra.com/cube/list/5f8f1c86fb32654470897b3a?s1=Unsorted&s2=Elo

Not all cards are displayed here. We should have cards with null elo default to 1200, which is the starting value anyway.

lunakv:
Fixes #1655.

The card update script now adds a default elo of 1200 to cards that don't have any computed elo yet. Sort by elo now also provides a fallback value of 1200 for cards with undefined elo. (Which should only be needed until the next card update is run).

### DIFF
--- a/serverjs/updatecards.js
+++ b/serverjs/updatecards.js
@@ -606,7 +606,7 @@ function convertCard(card, isExtra) {
     eur: card.prices.eur ? parseFloat(card.prices.eur, 10) : null,
     tix: card.prices.tix ? parseFloat(card.prices.tix, 10) : null,
   };
-  newcard.elo = catalog.elodict[name];
+  newcard.elo = catalog.elodict[name] ? catalog.elodict[name] : 1200;
   newcard.embedding = catalog.embeddingdict[name];
   newcard.digital = card.digital;
   newcard.isToken = card.layout === 'token';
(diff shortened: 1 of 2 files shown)


serverjs/updatecards.js
```diff
@@ -606,7 +606,7 @@ function convertCard(card, isExtra) {
     eur: card.prices.eur ? parseFloat(card.prices.eur, 10) : null,
     tix: card.prices.tix ? parseFloat(card.prices.tix, 10) : null,
   };
-  newcard.elo = catalog.elodict[name];
+  newcard.elo = catalog.elodict[name] ? catalog.elodict[name] : 1200;
```
dekkerglen, line 609: this can be written more concisely as `catalog.elodict[name] || 1200`

lunakv: True. I tried using `??` and that wasn't compatible, so I gave up. That brings up another question though - is it possible for a card to have 0 elo? As small of an edge case as it may be, it might warrant a `typeof` check instead.

dekkerglen: It's unrealistic for a card to have 0 Elo, `||` should be fine

lunakv: Okay, but when the Ghost of Bugs Yet to Come goes to haunt you at night, don't run to me.

PR #1679, opened 2020-10-28
Fix console warnings

wirmar:
Fixed different console warnings and some typo bugs.
(For example in the advanced filter modal: Fixed, so values can be entered for `priceEur` and `priceTix`.)

### DIFF
--- a/src/components/MagicMarkdown.js
+++ b/src/components/MagicMarkdown.js
@@ -24,7 +24,11 @@ const InnerMarkdown = ({ markdown }) => {
           if (section.startsWith('$$$')) {
             const sub = section.substring(1, section.length - 1);
             return (
-              <Latex displayMode trust={false} key={`section-${position}`}>
+              <Latex
+                displayMode
+                trust={false}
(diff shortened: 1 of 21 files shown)


src/components/MagicMarkdown.js
```diff
@@ -24,51 +24,84 @@ const InnerMarkdown = ({ markdown }) => {
           if (section.startsWith('$$$')) {
             const sub = section.substring(1, section.length - 1);
             return (
-              <Latex displayMode trust={false}>
+              <Latex
+                displayMode
+                trust={false}
+                key={/* eslint-disable-line react/no-array-index-key */ `section-${position}`}
```
wirmar, line 30: Index can be used as a key if no better option is available, therefore I disabled the lint rule here (like it was already done in some places in the file).
If no key is provided, the index would be taken by default, so this is technically no change, but the warning in the console disappears.
Source: https://reactjs.org/docs/lists-and-keys.html#keys

I'm not sure what the best approach for such a Situation is, but I think it is better if most warnings are gone, so the ones that are left are not ignored, and new warnings that appear when developing will be more likely to get fixed.

PR #1680, opened 2020-10-29
use compression middleware in express

### LINKED ISSUE #1678: Use gzip compression
Currently, the traffic sent to the browser by the server is not compressed.
This leads to longer download times and therefore higher [First Contentful Paint (FCP)](https://web.dev/fcp/).

This can be done via an express middleware https://web.dev/reduce-network-payloads-using-text-compression/#dynamic-compression
Can be added here: https://github.com/dekkerglen/CubeCobra/blob/master/app.js#L112

Another option would be to let nginx handle the compression in the production environment.

##### With compression (already provided in the development environment by webpack-dev-server):
2020-10-28_11h50_23

##### Without compression:
2020-10-28_12h03_04

wirmar:
fixes #1678

### DIFF
--- a/app.js
+++ b/app.js
@@ -9,6 +9,7 @@ const session = require('express-session');
 const passport = require('passport');
 const http = require('http');
 const fileUpload = require('express-fileupload');
+const compression = require('compression');
 const MongoDBStore = require('connect-mongodb-session')(session);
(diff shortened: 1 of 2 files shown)


joshfinnie (approved):
yea, this looks good from here...

Issue #1681, opened 2020-10-29
NODE_ENV not set correctly in production


wirmar:
![2020-10-29_15h36_51](https://user-images.githubusercontent.com/9307361/97588289-a2e06500-19fc-11eb-9900-6639cf0ec14b.png)

The development builds of react and jquery are used in production.
This can be fixed by starting the node server with the environment variable `NODE_ENV` set to `production`.
This environment variable will get picked up here https://github.com/dekkerglen/CubeCobra/blob/master/app.js#L219 and then used here https://github.com/dekkerglen/CubeCobra/blob/master/views/main.pug#L44.

The production builds are smaller and do not include unnecessary checks that slow down performance.

dekkerglen:
Doing this in production currently causes the server to fail to start. Needs more investigation

wirmar:
I just tried to fix the issues why it doesn't start in production. But now the server side rendering seems to introduce a lot of subtle bugs in the code 🤔
I will look into it a bit more, but maybe disabling server side render is the saver option? (Which would still allow using the production builds of the dependencies.)

PR #1682, opened 2020-10-29
fixed being able to start node server in production mode

wirmar:
related to #1681 

I'm not sure if I have caught every bug related to browser APIs not being available in server side rendering mode (nodejs environment).
Options would be to more thoroughly test the PR in NODE_ENV=production mode or to disable server side render?


src/hooks/UseNewID.js
```diff
@@ -1,10 +1,9 @@
-import { useMemo } from 'react';
-
-let lastID = 1000;
+import { useMemo, useState } from 'react';
 
 const useNewID = () => {
+  const [lastID, setLastID] = useState(1000);
```
wirmar, line 4: My change here is wrong (as pointed out by the linter). 
The problem why I changed it was, that on the server side render and the client hydration differnt IDs where generated I think.

dekkerglen (approved):
Looks good, just need to fix the failing CI

PR #1686, opened 2020-10-30
Fixed 404 redirects

### LINKED ISSUE #1685: 404 User invalid reroute
When you get a 404 user, it redirects you infinitely instead of redirecting you to the correct `404` page.

lunakv:
Fixes #1685 .

## Cause of issue
When requesting `cubecobra.com/user/view/xxx`, where `xxx` is a nonexistent username, the server would try to redirect to `'404'`. Since this is a relative path, the redirect would resolve to `cubecobra.com/user/view/404`. However, since there is no user named 404, this would trigger the redirect again, causing an infinite loop. The same issue could also be found in some redirects from `cube_routes`.

## Fix
Change the redirect to `/404`, which is always an available page.

## Additional notes
A lot of the changes made in `cube_routes` aren't actually due to redirect loops. Some were even left as `'404'`, because a cube with that id exists. Other problems fixed in this PR are
- some routes expected a draft id or a deck id, so when using the `'404'` redirect, the page would still redirect to `/404`, except with a cryptic database error instead of a clearer specified message. Those redirects were changed to `/404`.
- the POST requests for editing cubes would redirect to the 404 cube, which would then result in a 'Not Authorized' error, since most people don't have write access to that cube. This should probably almost never happen, since these routes are pretty much exclusively accessed from links, but I felt that the extra error could be confusing, so those redirects were also changed to `/404`.

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -498,7 +498,7 @@ router.get('/blog/:id/:page', async (req, res) => {
 
     if (!cube) {
       req.flash('danger', 'Cube not found');
-      return res.redirect('404');
+      return res.redirect('/404');
     }
 
     const countQ = Blog.countDocuments({ cube: cube._id });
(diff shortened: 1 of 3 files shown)


dekkerglen:
You must have a CC issue feed that notifies you immediately, right? This fix was _shockingly_ fast

lunakv:
Nah :) I just checked the issue tracker and this seemed easy enough to fix. I also had nothing better to do.

Issue #1669, opened 2020-10-24
Video Embeds in Markdown


ruler501:
**What would you like to be able to do?**
Have a syntax for embedding a video player in the page as a separate block.

**Describe how you would use the feature**
Allow users to watch referenced videos in page instead of navigating away and hoping they come back. Also saves them from needing to click through the warning modal for external links.

**Additional context**
There is not a standard markdown syntax for this so we would need to decide on something.

Issue #1639, opened 2020-10-12
Recommender suggestions background not set correctly in dark mode


dekkerglen:
![image](https://user-images.githubusercontent.com/28238025/95795451-5a187480-0cb8-11eb-8587-4861d2a24121.png)

dekkerglen:
Also, this: https://cdn.discordapp.com/attachments/600537623450091550/765968022824288276/Screenshot_from_2020-10-14_19-02-25.png

ruler501:
Card names in the table of scores on the Draftbot Analysis component have a white background.

ruler501:
Latex blocks in block quotes are illegible in dark mode, it's all light gray. Looking really closely I can somewhat make out the text as a very, very slightly different gray.

ruler501:
Markdown quote blocks are white text on white background.

struesdell:
> Markdown quote blocks are white text on white background.

I ran into this adding an overview to my cube with an included quote.

jbts:
@dekkerglen I think this can be closed now with the Markdown parser replacement being merged in.

PR #1688, opened 2020-10-31
Markdown parser replacement

### LINKED ISSUE #1668: Code Blocks in Markdown
Be able to use standard Markdown codeblocks(inline and paragraph) in places that accept markdown on the site. The text should be monospaced and have a different background to the rest of the text, see Github or Discord for examples.

It can be used to show code or otherwise set things apart from the rest of the text like button labels if you are describing menu interactions.

**Additional context**
single back ticks surrounding text is inline.
Triple backticks surrounding text marks a paragraph, I believe any text on the same line as the opening backticks is used to specify syntax highlighting. Blank lines at the beginning and end of blocks should be stripped.

### LINKED ISSUE #1670: Use Dedicated Markdown Parser
I would like to reimplement the Markdown parser in a nearley grammar. This would hopefully be simpler to write, read, and less likely to contain error.s I would like to have it specified as a nearley grammar that parses it into a UI element(or syntax tree if we want to use this on the backend as well, say for preparsing, filtering, etc.

### LINKED ISSUE #1671: Support Horizontal Dividers in Markdown
I would like to be able to specify that there should be a horizontal bar between specific blocks when writing markdown.

Add a horizontal divider between sections of my article.

**Additional context**
This is in standard markdown as 3 or more - on a line by themselves.

lunakv:
Resolves #1668. Resolves #1670. Resolves #1671. Fixes part of #1639 (regarding quote rendering).

This PR tracks the progress of replacing our in-house Markdown parser with [react-markdown](https://github.com/remarkjs/react-markdown/).

For a quick look at what rendered Markdown will look like, see [this comment](https://github.com/dekkerglen/CubeCobra/pull/1688#issuecomment-726644115)

## TODO:
* [x] Replace usage of current parser with react-markdown
* [x] Extend parser with currently used non-standard tokens 
    * [x] Strikethrough text
    * [x] Card links
    * [x] Card images
    * [x] Row of card images
    * [x] Symbols
    * [x] User links
    * [x] LaTeX math
    * [x] LaTeX inline math
    * [x] Centered text
        * [x] Single line
        * [x] Block
* [x] Test new parser 
    * [x] Write tests for parser (*available [here](https://github.com/dekkerglen/CubeCobra/pull/1688#issuecomment-722270269)*)
    * [x] Resolve found issues
* [x] Resolve differences between current parser and new parser
    * [x] Find inputs where parsers differ ([here](https://github.com/dekkerglen/CubeCobra/pull/1688#issuecomment-720135974))
    * [x] Decide which parser produces the desired behaviour
    * [x] Reproduce desired behaviour in new parser
    * [x] Compile a document enumerating remaining changes to existing documents
* [x] Determine which new Markdown features we want to include
    * [x] Heading autolinks
    * [x] Full GFM
    * [x] Code highlighting
* [x] Rewrite Markdown page
* [x] Write a developer overview for creating new plugins

## Known Differences / Bugs
- LaTeX math rendering is handled a little differently (see [comment below](https://github.com/dekkerglen/CubeCobra/pull/1688#issuecomment-720135974))
- ~rows of card images behave a little weird in centered text~
- ~horizontal rules are nearly invisible in dark mode~
- ~`limited` option passed from comments is not respected~
- ~comments are all around handled weirdly right now (more on that soon™)~
- ~block code needs formatting work~
    - ~inline code and block code should match in theme~
    - ~code should change theme in dark mode(?)~
    - choose code themes if `a11y` isn't what we want ([list](https://github.com/react-syntax-highlighter/react-syntax-highlighter/blob/master/AVAILABLE_STYLES_HLJS.MD) of available themes, [demo](https://react-syntax-highlighter.github.io/react-syntax-highlighter/demo/))
- ~tables need formatting work~
- ~heading links have no contents~
- ~headling links have `target="_blank"`, which we don't want~
- ~quotes have white text on white background in dark mode~

## Visual updates
- removed inline code background
- changed inline code color in both modes
- fixed quote block background in dark mode.
- slightly adjusted quote block background in default mode
- added more horizontal whitespace on both sides of Markdown content


src/markdown/userlink/index.js
```diff
@@ -0,0 +1,3 @@
+import plugin from './remark-userlink';
```
ruler501, line 1: Absolute imports, unless the intention is to split these off into separate packages or repost.

src/markdown/userlink/mdast-extension-userlink.js
```diff
@@ -0,0 +1,27 @@
+export const fromMarkdown = {
+  enter: { userlink: enterUserlink, userlinkValue: enterUserlinkValue },
```
ruler501, line 2: Please move function definitions to before they are used.

src/markdown/userlink/micromark-extension-userlink.js
```diff
@@ -0,0 +1,45 @@
+import assert from 'assert';
+
+const userlink = {
+  tokenize: tokenizeUserlink,
```
ruler501, line 4: Please move function definitions to before they are used.

src/markdown/userlink/remark-userlink.js
```diff
@@ -0,0 +1,11 @@
+import syntax from './micromark-extension-userlink';
+import { fromMarkdown } from './mdast-extension-userlink';
```
ruler501, line 2: Same comment on absolute imports.

src/components/MarkdownRenderer.js
```diff
@@ -0,0 +1,61 @@
...
+    </a>
+  );
+}
+
+function renderSymbol(node) {
+  const symbol = node.value.replace('/', '-').toLowerCase();
+  return <img src={`/content/symbols/${symbol}.png`} alt={symbol} className="mana-symbol-sm" />;
+}
+
+const Markdown = ({ markdown, limited }) => {
```
ruler501, line 33: I'm not a fan of having the component name different from the filename.

src/markdown/strikethrough/index.js
```diff
@@ -0,0 +1,12 @@
+import syntax from 'micromark-extension-gfm';
+import fromMarkdown from 'mdast-util-gfm/from-markdown';
+import { add } from '../utils';
+
+function strikethrough(options) {
```
ruler501, line 5: We've tended to use arrow functions for everything, not sure that they should be required though.

ruler501 (changes requested):

src/components/MarkdownRenderer.js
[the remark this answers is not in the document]
lunakv: Nothing yet, but the old parser uses it, so I assume we'll need it at some point. I believe it's a parameter passed from comments where some markdown features aren't allowed.

src/analytics/Tokens.js
```diff
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 
 import { getTCGLink } from 'utils/Affiliate';
 
-import MagicMarkdown from 'components/MagicMarkdown';
+import Markdown from 'components/MarkdownRenderer';
```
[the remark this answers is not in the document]
lunakv: Fair point. I renamed the file.

src/markdown/symbols/index.js
```diff
@@ -0,0 +1,3 @@
+import plugin from './remark-symbols';
```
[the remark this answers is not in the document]
lunakv: It can be. Now it is 👍.

src/markdown/symbols/remark-symbols.js
```diff
@@ -0,0 +1,16 @@
+import syntax from './micromark-extension-symbols';
+import { fromMarkdown } from './mdast-extension-symbols';
+import { add } from '../utils';
+
+export default symbols;
+
+function symbols(options) {
+  if (!options.allowed) {
```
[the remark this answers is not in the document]
lunakv: Good point. Fixed.

src/markdown/strikethrough/index.js
lunakv: Glancing through the code base, it looks like a mixed bag. Components and pages are pretty consistently written as lambdas, but a lot of utility functions are written with the regular function syntax. I can change it if you really don't like it, but personally I don't mind it as is.

src/markdown/symbols/mdast-extension-symbols.js
```diff
@@ -0,0 +1,27 @@
+export const fromMarkdown = {
```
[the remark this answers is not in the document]
lunakv: I agree with this in general, and moved the functions accordingly in most places. The one exception where I feel this isn't a good idea is for the `tokenize` functions in micromark extensions. Since those functions describe transitions of a state machine, it makes sense to me that they're written in order of when those states are reached.

ruler501:
These are the plug-ins I think we might be interested in. Some of them I believe you're already using.

#### Reasonably sure these will work

https://github.com/remarkjs/remark-math render math. Uses 1 and 2 dollar signs vs we currently use 2 and 3.
https://github.com/remarkjs/remark-slug slugs(ways to jump to specific headings) for headings
https://github.com/remarkjs/remark-toc generate a table of contents based on headings with links to those headings.
https://github.com/elviswolcott/remark-admonitions highlited box in different colors 
https://github.com/remarkjs/remark-github auto support recognizing github issues and issue numbers like github does
https://github.com/remarkjs/remark-footnotes footnotes that work as links to the footnote and back to where it was referenced.
https://github.com/remarkjs/remark-gfm support strike through, tables, automatic link recognition, and task lists(static)
https://github.com/react-syntax-highlighter/react-syntax-highlighter component to hook into for syntax highlighting example in the react-markdown Readme. 

#### I can't tell if these will work

https://github.com/remarkjs/remark-autolink-headings along with slugs gives you a little icon next to headings that is a link to the url that navigates directly to that heading. Unclear if works with react-markdown. 
https://github.com/Darkhax/remark-spoiler discord spoiler syntax, might not work with the new parser.
https://github.com/agentofuser/remark-oembed embed videos probable issues with new parser. This would fix #1669 
https://github.com/mottox2/remark-code-titles enables adding a title to code blocks

Also can you add to the description that this fixes #1668 #1670 and #1671

lunakv:
Thanks for the list! I am currently using remark-math, remark-gfm (more specifically, the strikethrough part of it), and remark-breaks for compatibility. I'll look at the rest more closely once we get to parity with the current parser. I also wanted to open a discussion before adding new capabilities.

One note about remark-math: despite what its documentation might imply, you don't actually *need* to have exactly one/two `$`s for inline/block math. Those are just the lower limits. Meaning any number of `$` can be parsed as an inline math label and two or more `$` can be parsed as a block math label. If you're using more than one `$`, the differentiation between them is determined by whether the closing label is on the same line as the opening label or not.

The result of this is that current math written for our parser will mostly work unchanged, with the exception of  lines written as `text $$$ math $$$ text`, which are seen as block math under our parser, but as inline math with remark-math. Those would need to be turned into
```
text
$$$
math
$$$
text
```
to be correctly displayed as block math.

src/markdown/strikethrough/index.js
ruler501: I personally like lambdas more, but let's see what @dekkerglen has to say.

dekkerglen: +1 for using lambdas

lunakv: I replaced the renderer functions in `Markdown.js` with lambdas. The markdown extensions themselves have to stay as regular functions, because the parser supplies custom `this` values to them, which wouldn't be possible with lambdas. I could probably comb through them and replace whichever functions won't break, but at that point I think it's better to stay consistent at least within the file.

lunakv:
I created two test files that should help with debugging.
- [common_mark.txt](https://github.com/dekkerglen/CubeCobra/files/5493406/common_mark.txt) uses features from CommonMark that weren't implemented in our current parser. This document is more useful for checking rendering, as we can assume the parser to work correctly on regular markdown input.
- [plugins.txt](https://github.com/dekkerglen/CubeCobra/files/5493397/plugins.txt) uses our custom plugins and tests how they integrate with other markdown elements.

src/markdown/centering/micromark-centering.js
```diff
@@ -0,0 +1,109 @@
+import assert from 'assert';
+import markdownSpace from 'micromark/dist/character/markdown-space';
+import markdownLineEnding from 'micromark/dist/character/markdown-line-ending';
+import spaceFactory from 'micromark/dist/tokenize/factory-space';
+import types from 'micromark/dist/constant/types';
+import codes from 'micromark/dist/character/codes';
+import { shallowEqual } from 'markdown/utils';
+
+function tokenizeCentering(effects, ok, nok) {
+  let size = 0;
```
ruler501: Should this get reset to 0 at some point?

src/markdown/userlink/micromark-userlink.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-use-before-define */
```
ruler501, line 1: You should add this to the elsint rules file instead of to all these files.

ruler501 (commented):
For plug-ins the ones I think are most important are slugs, autolink slugs(if it works), syntax highlighter, oembed(if it works, if not I think we should create our own, possibly with different syntax).

src/markdown/centering/micromark-centering.js
lunakv: It doesn't have to. The tokenize function is called again when a new block is attempted to be parsed.

src/components/Markdown.js
```diff
@@ -24,6 +27,24 @@ function renderBlockQuote(node) {
   );
 }
 
+function renderLink(node) {
+  const ref = encodeURI(node.node?.url ?? '');
+
+  const isInternalURL = (to) => {
```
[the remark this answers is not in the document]
lunakv: Good point. I just copied this verbatim from the original parser, but it probably deserves a place in `Utils`

lunakv:
Full GFM support was added, with a corresponding overview and text file update.

lunakv:
Slugs and autolink slugs were added. (Note: We may not want them enabled in comments)

lunakv:
Base support for syntax highlighting in code blocks was added. There are still some considerations to be made regarding styling (outlined in the overview).

lunakv:
Since all the most important features we wanted were implemented*, I'm now marking this PR as feature complete. All future efforts will be focused on debugging and documentation.

\* With the exception of oembed links. I don't want to use the existing plugin, because 1) it doesn't allow for any special syntax, it just converts urls directly to videos, which we might not want, b) I don't understand how iframes work nearly as well as I should, and 4) I'm not a fan of using a project that wasn't touched by its creator for more than a year. Because writing a new plugin for this would be a lot of work (both in research and in implementation), I think this feature merits a separate PR somewhere down the line.

lunakv:
I noticed that React throws a bunch of warnings about card images - since they are parsed as inline elements, they are rendered inside a paragraph, which leads to `<div>`s nested inside `<p>`s, which it apparently doesn't like. Since no browser actually cares and all of them render this just fine, I'm happy to propose this doesn't need fixing and let the DOM validator whine in the console.

lunakv:
Adding some screenshot previews of what a Markdown document can look like: (desktop Chromium)
dark_mode
light_mode

ruler501 (approved):

dekkerglen (approved):

PR #1721, opened 2020-11-17
Fix comment edit route

lunakv:
Editing comments caused them to be truncated to 500 characters instead of the correct 5000. This PR fixes that issue. I also removed the character limit on the comment parent type, since we already checked its value 10 lines earlier.

### DIFF
--- a/routes/comment_routes.js
+++ b/routes/comment_routes.js
@@ -83,7 +83,7 @@ router.post(
     const comment = new Comment();
 
     comment.parent = req.params.parent.substring(0, 500);
-    comment.parentType = req.params.type.substring(0, 500);
+    comment.parentType = req.params.type;
     comment.owner = poster._id;
     comment.ownerName = poster.username;
     comment.image = poster.image;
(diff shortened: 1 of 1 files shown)


routes/comment_routes.js
```diff
@@ -83,7 +83,7 @@ router.post(
     const comment = new Comment();
 
     comment.parent = req.params.parent.substring(0, 500);
-    comment.parentType = req.params.type.substring(0, 500);
+    comment.parentType = req.params.type;
```
dekkerglen, line 86: we should still substring this - as we don't want to create an opportunity for a 10GB database query, for example

dekkerglen (changes requested):

routes/comment_routes.js
lunakv: We really don't need to though. If you look 10 lines higher, we check the `type` against an array of values, the largest of which is 7 characters long, and if it matches none of them, we return a 400 and the code never gets to this point.

PR #1722, opened 2020-11-17
Fix some eslintignore exclusions

struesdell:
Related to #1601 

Added dependencies
- [classnames](https://www.npmjs.com/package/classnames) for terse className concatenation

ESLint fixes
- src/components/AspectRatioBox
- src/components/AutocardListItem

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "canvas": "^2.6.1",
     "chart.js": "^2.9.2",
     "cheerio": "^1.0.0-rc.3",
+    "classnames": "^2.2.6",
     "compression": "^1.7.4",
     "compute-cosine-similarity": "^1.0.0",
     "connect-flash": "^0.1.1",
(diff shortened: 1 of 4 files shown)


src/components/AutocardListItem.js
```diff
@@ -1,23 +1,47 @@
-import React, { useCallback, useContext } from 'react';
+import React, { useCallback, useContext, useMemo } from 'react';
+import cx from 'classnames';
+import propTypes from 'prop-types';
```
lunakv, line 3: `PropTypes` should be used instead of `propTypes`

dekkerglen:
@struesdell Is this still WIP?

struesdell:
Going alphabetically, I stepped next to AutocompleteInput and fell into a deep hole of "Maybe I can clean this up so that `//please don't try to understand why this works` is no longer necessary...

But I'll make that into another pull request.

package.json
```diff
@@ -48,6 +48,7 @@
     "canvas": "^2.6.1",
     "chart.js": "^2.9.2",
     "cheerio": "^1.0.0-rc.3",
+    "classnames": "^2.2.6",
```
lunakv, line 51: Since we're introducing `classnames`, at some point we probably want to go through the codebase and introduce it wherever it makes sense. That's a separate issue for another time though.

lunakv (approved):

PR #1725, opened 2020-11-19
Added null check

### LINKED ISSUE #1637: Cube deck history not loading when logged out
Visit this page when logged out:
https://cubecobra.com/cube/decks/5e0bcc3a584b3f55ad02b277/0

It errors. When logged in, it works fine.

lunakv:
Fixes #1637 
When not logged in, the decklist page would crash, because it tried to reference `user.id` with `user === null`. I added a null check to that comparison so that this doesn't happen.


jbts:
Believe this fixes #1637.

lunakv:
Correct, #1637 is this exact bug.

Issue #1691, opened 2020-11-02
Notifications for @'ing a user


dekkerglen:
Our markdown parser parses `@dekkaru` and creates a tag to that user. It would be nice to get notifications when someone tags you in a comment/blog post/cube description

lunakv:
As mentioned in #1727, the best way to do this (that I can see) is taking the new markdown parser (#1688) with our `userlink` extension and extracting all user mentions from there.
We should consider whether we want the notifications to show only from the original comment/post, or if we also want them to appear when a user mention is introduced in an edit. This feature is optional for comments, posts, and articles, but crucial to cube overviews, where editing is the only way to introduce a user mention. We'd most likely need some new database structures for this, since we wouldn't want a user to be spammed with notifications each time someone makes an edit.

PR #1727, opened 2020-11-20
Notify when mentioned in comments

joshfinnie:
This PR sets up the ability to notify people who were mentioned in comment. This is an idea that comes from #1691... seems like it should be good.

### DIFF
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -456,7 +456,7 @@ const generateSamplepackImage = (sources = [], options = {}) =>
 
 const getUsersMentioned = (content) => {
   return content.match(/@(\w+)/g);
-}
+};
 
 const methods = {
   getUsersMentioned,
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -454,7 +454,12 @@ const generateSamplepackImage = (sources = [], options = {}) =>
     );
   });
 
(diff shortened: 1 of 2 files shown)


serverjs/cubefn.js
```diff
@@ -454,7 +454,12 @@ const generateSamplepackImage = (sources = [], options = {}) =>
     );
   });
 
+const getUsersMentioned = (content) => {
+  return content.match(/@(\w+)/g);
```
lunakv, line 458: The capture group here is unnecessary and won't appear in the match results.

lunakv (commented):
This issue should probably be addressed only after we merge the new markdown parser from #1688. As it stands, this PR creates some inconsistencies with the new parser in what it considers a user mention.
- A match that's preceded or succeeded by a non-whitespace character will trigger this, but not the new parser (e.g. `mail@google.com` will notify user `google` despite not being an actual user mention).
- This (specifically `\w`) allows underscores in a username, but CubeCobra doesn't (so neither does the new parser)
- This doesn't respect escape characters (e.g. `\@dekkaru` will trigger this, but will be treated as plain text in the new renderer).

A more consistent solution would involve utilizing the new parser with our user mention extension and extracting the data from that, but that should be implemented after we finalize work on the parser itself.

PR #1734, opened 2020-11-22
Add API path to expose Cube.date_updated

Ragaxus:
I am writing a bot that needs to keep a local copy of a given cube in sync with Cube Cobra. Instead of polling for the current state of the Cube, I thought it would be preferable to create a new, light-weight path to indicate whether changes have been made to the cube since the last time the bot checked; that way, the bot can only pull down the contents of the cube when it needs to.

### DIFF
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -4185,7 +4185,7 @@ router.get('/api/date_updated/:id',
       });
     }
     return res.status(200).send({
-      date_updated: cube.date_updated,
+      date_updated: cube.date_updated.valueOf(),
     });
   }),
 );
--- a/routes/cube_routes.js
+++ b/routes/cube_routes.js
@@ -4289,15 +4289,16 @@ router.get(
   '/api/date_updated/:id',
   util.wrapAsyncApi(async (req, res) => {
     const { id } = req.params;
-    const cube = await Cube.findOne(buildIdQuery(id)).lean();
-    if (!cube) {
+    const result = await Cube.findOne(buildIdQuery(id), { date_updated: 1, _id: 0 }).lean();
+    if (!result) {
       return res.status(404).send({
         success: 'false',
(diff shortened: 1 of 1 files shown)


lunakv:
Please run `npm beautify` so that your code passes the style check.

routes/cube_routes.js
```diff
@@ -4174,4 +4174,20 @@ router.get(
   }),
 );
 
+router.get('/api/date_updated/:id',
+  util.wrapAsyncApi(async (req, res) => {
+    const { id } = req.params;
+    const cube = await Cube.findOne(buildIdQuery(id)).lean();
```
dekkerglen, line 4180: you can make this more performant by doing:
`await Cube.findOne(buildIdQuery(id), 'date_updated').lean();`

This will only fetch the only the `date_updated` field. If you need other fields, just add them to the string, separated with spaces

dekkerglen (changes requested):

routes/cube_routes.js
```diff
@@ -4174,4 +4285,22 @@ router.get(
   }),
 );
 
+router.get(
+  '/api/date_updated/:id',
+  util.wrapAsyncApi(async (req, res) => {
+    const { id } = req.params;
+    const result = await Cube.findOne(buildIdQuery(id), { date_updated: 1, _id: 0 }).lean();
```
dekkerglen, line 4292:
```suggestion
    const result = await Cube.findOne(buildIdQuery(id), 'date_updated').lean();
```
I'd prefer to keep projection consistent across code base

dekkerglen (changes requested):

routes/cube_routes.js
dekkerglen: Read more here: https://mongoosejs.com/docs/api.html#query_Query-select

PR #1740, opened 2020-11-25
Added '~' support in oracle filter

lunakv:
This PR introduces the option to use `~` as a placeholder for a card's name when filtering oracle text. 

Example usage: `oracle:"When ~ enters the battlefield"` matches cards with an etb trigger.


ruler501 (commented):
For legendary creatures and planeswalkers the name might be shortened. See Prossh for an example.

lunakv:
Good point. I extended the matcher so that it supports shortened names on legendaries. I also let `~` replace `"this spell"`, since that's what Scryfall does. Note: that made me change the direction of replacements - instead of replacing `~` with the card name in the filter text, now we squish the name into `~` in the oracle text.

PR #1747, opened 2020-12-03
Added 'collector number' filter

lunakv:
Scryfall supports `cn:` or `number:` as a filter that matches on collector numbers. This PR adds support for that filter.

The argument to this filter is a string and not a number, because collector number is stored as a string, because some collector numbers include non-numeric characters (e.g. '★')﻿

### DIFF
--- a/nearley/cardFilters.ne
+++ b/nearley/cardFilters.ne
@@ -24,6 +24,7 @@ import {
   cardType,
   cardOracleText,
   cardSet,
+  cardCollectorNumber,
   cardPower,
   cardToughness,
   cardTags,
@@ -91,6 +92,7 @@ condition -> (
   | cubesCondition
   | legalityCondition
   | layoutCondition
+  | collectorNumberCondition
 ) {% ([[condition]]) => condition %}
 
 @{%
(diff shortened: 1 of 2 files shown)


ruler501 (changes requested):
I'd really like to see some kind of ordering added to it. I believe it should always be the case that you can split a collector number into 2 parts, the number and the possibly empty suffix. For ordering do a numerical comparison of the first part and a lexicographical of the second.

lunakv:
I don't really see a reason to do that. None of the other filters care about ordering, so the only reason to add ordering would be for implementing Sort/View by collector number, which doesn't seem very useful. Only other upside is that different numerically equivalent inputs will be compared correctly, but I don't know why anyone would want to write `cn:075` rather than just `cn:75`. Neither can I see someone wanting to filter by `cn>8`. (As an aside, Scryfall also does string comparisons. Try searching for `cn:075` there and the results will be empty.) 

Another point to this is that making assumptions about the exact format of a Magic card field can lead to  problems down the line. As an example, while it's undoubtedly useful to have power/toughness represented as numbers, our current filter breaks when handling `*/*` creatures (or Infinity Elemental) because of this assumption.

dekkerglen:
I'll merge as is, but I would like to see a CIN sort as well - that adds the ordering the Ruler mentioned

PR #1750, opened 2020-12-09
Use embeddings in card objects for drafts

### LINKED ISSUE #1547: Update embeddings to not call ML server

dekkerglen:
Fixes #1547

### DIFF
--- a/src/utils/draftbots.js
+++ b/src/utils/draftbots.js
@@ -104,29 +104,47 @@ const scaleSimilarity = (value) =>
 const SYNERGY_SCALE = 0.2;
 const MAX_SYNERGY = 10;
 
-let synergyMatrix;
+const xorStrings = (a, b) => {
+  let s = '';
 
-export const setSynergyMatrix = (size) => {
-  synergyMatrix = [];
-  for (let i = 0; i <= size; i++) {
-    synergyMatrix.push(new Array(size + 1).fill(null));
+  // use the longer of the two words to calculate the length of the result
+  for (let i = 0; i < Math.max(a.length, b.length); i++) {
+    // append the result of the char from the code-point that results from
+    // XORing the char codes (or 0 if one string is too short)
+
+    // eslint-disable-next-line no-bitwise
+    s += String.fromCharCode((a.charCodeAt(i) || 0) ^ (b.charCodeAt(i) || 0));
   }
+
+  return s;
 };
 
-export const getSynergy = (index1, index2, synergies) => {
-  if (!synergies) return 0;
-  if (synergyMatrix[index1][index2] === null) {
-    const similarityValue = similarity(synergies[index1], synergies[index2]);
-    if (Number.isFinite(similarityValue)) {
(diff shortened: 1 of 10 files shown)


src/utils/draftbots.js
```diff
@@ -104,29 +104,47 @@ const scaleSimilarity = (value) =>
...
-  synergyMatrix = [];
-  for (let i = 0; i <= size; i++) {
-    synergyMatrix.push(new Array(size + 1).fill(null));
+  // use the longer of the two words to calculate the length of the result
+  for (let i = 0; i < Math.max(a.length, b.length); i++) {
+    // append the result of the char from the code-point that results from
+    // XORing the char codes (or 0 if one string is too short)
+
+    // eslint-disable-next-line no-bitwise
+    s += String.fromCharCode((a.charCodeAt(i) || 0) ^ (b.charCodeAt(i) || 0));
```
ruler501, line 116: This should use a better hash, this is relatively likely to collide. Honestly I'd prefer to just use the names as keys directly after some minor sanitizing such as replacing illegal characters with something like `\u<CODEPOINT>` and adding a `__`(double underscore) between the two names. Could also just go back to using indices as well which is probably the fastest way to do it, if you don't want to store the whole matrix then use `<index1>_<index2>` as your keys, but this matrix isn't that big (typical draft has at most 800 cards which is only 5mb when stored as a complete matrix of doubles).

src/utils/draftbots.js
```diff
@@ -104,29 +104,47 @@ const scaleSimilarity = (value) =>
...
 };
 
-export const getSynergy = (index1, index2, synergies) => {
-  if (!synergies) return 0;
-  if (synergyMatrix[index1][index2] === null) {
-    const similarityValue = similarity(synergies[index1], synergies[index2]);
+const synergyCache = {};
+
+export const getSynergy = (card1, card2) => {
+  if (
```
ruler501, line 125: This can be simplified to `card1?.details?.embedding?.length && card2?.details.embedding?.length` Same in the other places you use this.

src/utils/draftbots.js
```diff
@@ -104,29 +104,47 @@ const scaleSimilarity = (value) =>
...
+  }
+
+  const key = xorStrings(card1.details.name, card2.details.name);
+  if (!synergyCache[key]) {
+    const similarityValue = similarity(card1.details.embedding, card2.details.embedding);
+
     if (Number.isFinite(similarityValue)) {
-      synergyMatrix[index1][index2] = -Math.log(1 - scaleSimilarity(similarityValue)) / SYNERGY_SCALE;
-    }
-    // This happens when similarityValue is 1, usually because a card is a duplicate.
```
ruler501, line 123: Proper handling of the degenerate cases is very important here since the positive case can occur if the similarities are close enough and we do want to properly populate a 0 in the other cases.

ruler501 (changes requested):

PR #1753, opened 2020-12-10
Cloud tooltips

ruler501:
Depends on #1752 

Adds tooltips with count on hover(click for touch devices) to the tags in the tag cloud.

Also makes the tag-cloud use light colors in the dark theme instead of the dark colors it uses in the default theme.


src/analytics/Cloud.js
```diff
@@ -1,23 +1,66 @@
...
+    count: PropTypes.number.isRequired,
+  }).isRequired,
+  size: PropTypes.number.isRequired,
+  color: PropTypes.string.isRequired,
+};
+
+const COLOR_OPTIONS = {
+  dark: 'light',
+  default: 'dark',
+  light: 'dark',
```
lunakv, line 39: We don't have a `light` theme, only default and dark.

PR #1764, opened 2020-12-26
Add Color Identity Full and fix Color Category Full.

ruler501:


src/utils/Sort.js
```diff
@@ -1,5 +1,13 @@
-import { alphaCompare, fromEntries } from 'utils/Util';
-import { cardDevotion, cardPriceEur, cardTix, cardCmc } from 'utils/Card';
+import { alphaCompare, arrayIsSubset, COLORS, fromEntries } from 'utils/Util';
```
lunakv, line 1: `COLORS` isn't used anywhere and fails linter check

PR #1766, opened 2020-12-27
Fixed devotion of adventurer cards and parsed cost of DFCs

### LINKED ISSUE #1765: Devotion sort counts adventures improperly
“When sorting by color devotion, adventure cards sneak into the higher devotion categories.
For example [[Faerie Guidemother]] is in the white devotion 2 pool”

Devotion sort should only count the permanent side of the card and disregard the adventure side.

lunakv:
Fixes #1765 

This PR fixes the following issues:

### Adventurer card devotion
The devotion counter was previously counting all symbols in the parsed cost. With adventurer cards, we only want to count the symbols in the mana cost of the permanent side, so the devotion counter was adjusted to reflect this.

### DFC mana cost parsing
Due to a bug in `updatecards.js`,  double-faced cards didn't parse mana costs properly, but returned an empty array instead. This issue has been resolved so that the parsed cost now equals the mana cost of the front face (for example, [Archangel Avacyn](https://scryfall.com/card/soi/5/archangel-avacyn-avacyn-the-purifier) now has a parsed cost of `['w', 'w', '3']` instead of the previous `['']`)

### Updated tests and fixtures
Tests and test fixtures were updated to reflect changes to mana cost parsing, modifications to the card object layout, (such as the inclusion of `released_at`, `prices`, `layout` or `set_name` fields or removal of `Oldschool` and `Duel` from formats), and to replace incorrect values (e.g. wrong `Penny` legality or change to Scryfall's CDN for image URLs).


src/utils/Card.js
```diff
@@ -250,8 +250,13 @@ export const cardElo = (card) => card.details.elo;
 
 export const cardLayout = (card) => card.details.layout;
 
-export const cardDevotion = (card, color) =>
-  cardCost(card)?.reduce((count, symbol) => count + (symbol.includes(color.toLowerCase()) ? 1 : 0), 0) ?? 0;
+export const cardDevotion = (card, color) => {
+  console.log(card.details.name);
+  console.log(card);
+  let cost = cardCost(card);
+  if (cost && cardLayout(card) === 'adventure') cost = cost.slice(cost.findIndex((x) => x === 'split') + 1);
```
lunakv, line 257: Mana costs are reversed when parsing, so the cost of the permanent half is actually *after* the `'split'` symbol.

The layout check needs to be here because split cards also have a `'split'` mana symbol, and for them it makes more sense to count both halves.

PR #1768, opened 2020-12-27
Fixed typo

lunakv:
http://codecobradev.com/ doesn't actually exist (though it's a pretty cool domain)


dekkerglen:
Unfortunately, I do actually own the domain codecobradev.com (accidents happen)

PR #1769, opened 2020-12-28
Update health check policy

dekkerglen:
This change is described here: https://aws.amazon.com/premiumsupport/knowledge-center/elastic-beanstalk-instance-automation/

### Problem
The previous behavior is that unhealthy instances are removed from the load balancer group but not the autoscaling group. So if an instance becomes unhealthy and doesn't recover, it takes up autoscaling capacity. This causes transient errors to pile up and eventually take out the service when every instance in the has been taken out of the load balancing group, and so the service just 503s because the load balancing group is empty.

### Solution
Adjust autoscaling config so that unhealthy instances are removed from the autoscaling group, not the load balancing group.

### DIFF
--- a/.ebextensions/03_autoscaling.config
+++ b/.ebextensions/03_autoscaling.config
@@ -1,0 +1,6 @@
+Resources:
+  AWSEBAutoScalingGroup:
+    Type: AWS::AutoScaling::AutoScalingGroup
+    Properties:
+      HealthCheckType: ELB
+      HealthCheckGracePeriod: 900


.ebextensions/03_autoscaling.config
```diff
@@ -0,0 +1,6 @@
+Resources:
+  AWSEBAutoScalingGroup:
+    Type: AWS::AutoScaling::AutoScalingGroup
+    Properties:
+      HealthCheckType: ELB
+      HealthCheckGracePeriod: 900
```
joshfinnie, line 6: do we really need/want 15 minute lead time for the health check grace period? that seems like a lot and could lead to a buildup of dead ec2s still?

joshfinnie (approved):

joshfinnie (approved):

stephenmenton (approved):
considered 15m a little high but discussed on discord, cardbase update time. makes sense. approved.

dekkerglen:
> Note: HealthCheckGracePeriod refers to the amount of time, in seconds, that Amazon EC2 Auto Scaling waits before checking the health status of an EC2 instance that's come into service.

15 minutes is just to allow the card base update to run before service is health checked

PR #1792, opened 2021-01-15
Update react

ruler501:
Can make download sizes a little smaller for users as well as removing a bunch of unnecessary imports.


lunakv:
Looks like this branch is based on an older version of `master` that doesn't include #1790, meaning the `React` import wasn't removed from `LeaveWarningPage`

PR #1793, opened 2021-01-15
Fixed regression

lunakv:
Fixed issue introduced in #1786

When creating the `useQueryParam` hook, the parameters of this call weren't updated from the previous `useState` call, ultimately resulting in a crash whenever you try to open the Draftbot Analysis view (since the `index` was now being set to `null` instead of a number, which further code wasn't expecting).


ruler501 (approved):
huh I thought I'd caught all of these, thank you.

PR #1795, opened 2021-01-17
Refactoring card filter

lunakv:
Because "creature land" seems to be the one land category that steadily gets more cards with each set, this PR removes the previous manual enumeration of this category and replaces it with a general oracle-text-based filter condition (confirmed to be equivalent with Scryfall's definition for all current creature lands). It also adds "manland" as an alias to "creatureland" and does some other filtering-related refactoring.

#### Changes:
- reworked `is:creatureland` filter
- added `is:manland` as an alias
- moved `LandCategories` into a JSON file under `src/res`
- fixed `is:spell` and `is:permanent` to not include unusual cards, such as Planes and Vanguards
- split `isValue` grammar rule into multiple lines
- simplified `cardIsSpecialZoneType` to remove redundant planeswalker check


nearley/cardFilters.ne
```diff
@@ -170,4 +170,4 @@ notCondition -> "not"i isOpValue {% ([, valuePred]) => negated(genericCondition(
...
+isValue -> ("gold"i | "twobrid"i | "hybrid"i | "phyrexian"i | "promo"i | "digital"i | "reasonable"i | "dfc"i | "mdfc"i | "meld"i | "transform"i | "split"i | "flip"i | "leveler"i | "commander"i | "spell"i | "permanent"i | "historic"i | "vanilla"i | "bikeland"i | "cycleland"i | "bicycleland"i | "bounceland"i | "karoo"i | "canopyland"i | "canland"i | "checkland"i | "dual"i | "fastland"i | "filterland"i | "gainland"i | "painland"i | "scryland"i | "shadowland"i | "shockland"i | "storageland"i | "creatureland"i | "manland"i | "triland"i | "tangoland"i | "battleland"i | "modal"i) {% ([[category]]) => category.toLowerCase() %}
```
ruler501, line 173: Please split this onto multiple lines, it's very hard to read right now.
I think as a rough style guide we should try to avoid lines over 120 characters in nearley code.

src/utils/Card.js
```diff
@@ -74,6 +74,16 @@ export const CARD_CATEGORY_DETECTORS = {
     details.legalities.Commander === 'legal' &&
     ((details.type.includes('Legendary') && details.type.includes('Creature')) ||
       details.oracle_text.includes('can be your commander')),
+  spell: (details) => !details.type.includes('Land') && !details.type.includes('Conspiracy'),
```
ruler501, line 77: use the special card type filter we use for the draftbots so that it also filters out planes, vanguards, tokens, and other things that don't go into decks.

src/utils/Card.js
```diff
@@ -74,6 +74,16 @@ export const CARD_CATEGORY_DETECTORS = {
...
+  permanent: (details) =>
+    !details.type.includes('Instant') && !details.type.includes('Sorcery') && !details.type.includes('Conspiracy'),
```
ruler501, line 79: Same here.

src/utils/Card.js
```diff
@@ -74,6 +74,16 @@ export const CARD_CATEGORY_DETECTORS = {
...
+  historic: (details) =>
+    details.type.includes('Legendary') || details.type.includes('Artifact') || details.type.includes('Saga'),
+  vanilla: (details) => !details.oracle_text,
+  modal: (details) => details.oracle_text.includes('•'),
+  creatureland: (details) => details.type.includes('Land') && details.oracle_text.match(/\bbecomes? a .*\bcreature\b/),
```
ruler501, line 84: Please move this function out into a constant instead of duplicating it.

src/utils/LandCategories.js
```diff
@@ -1,34 +1,4 @@
 export default {
-  CREATURE: [
```
ruler501, line 2: Can we make this a json file and put it in res with the casting probability table?

ruler501 (changes requested):

PR #1801, opened 2021-01-24
Adding auto-update to copyright footer

joshfinnie:
Noticed the copyright in the footer was still on `2019` this just adds a simple script to auto increment the year using javascript.

### DIFF
--- a/src/layouts/Footer.js
+++ b/src/layouts/Footer.js
@@ -2,7 +2,7 @@ import React from 'react';
 
 import { Row, Col, Container } from 'reactstrap';
 
-import Copyright from '../components/Copyright';
+import Copyright from 'components/Copyright';
 
 const Footer = () => (
   <footer>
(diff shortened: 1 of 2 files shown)


src/layouts/Footer.js
```diff
@@ -161,7 +161,7 @@ const Footer = () => (
         </a>
         .
         <br />
-        All other content Copyright © 2019 Cube Cobra
+        All other content Copyright © 2019 <script>new Date().getFullYear()>2019&&document.write("- "+new Date().getFullYear());</script> Cube Cobra
```
ruler501, line 164:
```suggestion
       All other content Copyright © 2019 {new Date().getFullYear() > 2019 ? `- ${new Date().getFullYear()} ` : ' '} Cube Cobra.
```

ruler501 (changes requested):
Could we do this with a react component instead of a

src/layouts/Footer.js
joshfinnie: yea, that makes sense... i'll update tomorrow

ruler501 (dismissed):

ruler501 (approved):

PR #1808, opened 2021-02-12
Crash hotfix

lunakv:
In #1807, I forgot that route files aren't run through Babel. Because Node doesn't support optional chaining, the server would crash on start.

### DIFF
--- a/routes/tools_routes.js
+++ b/routes/tools_routes.js
@@ -45,8 +45,8 @@ async function matchingCards(filter) {
       const history = historyDict.get(card.oracle_id);
       return {
         ...card,
-        picks: history?.current.picks,
-        cubes: history?.current.cubes,
+        picks: history ? history.current.picks : undefined,
+        cubes: history ? history.current.cubes : undefined,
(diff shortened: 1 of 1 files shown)


ruler501:
I'm hoping we can fix this in the future since I think node 14 would allow us to not need babel to cross-compile for using on the server.

ruler501 (approved):

lunakv:
As far as I understand it, support for ES modules is still experimental in Node 14 and doesn't support all normal features, so it'll probably be a while yet before we can fully get rid of babel.

ruler501:
https://nodejs.org/docs/latest-v14.x/api/esm.html that says there's full support for modules in 14.

https://nodejs.medium.com/node-js-version-14-available-now-8170d384567e this lists optional chaining as having full support as well.

lunakv:
The same link (the first one) also says "Stability: 1 - Experimental" at the top.

ruler501:
I had missed that. Maybe node 16 then when it gets here.

PR #1819, opened 2021-02-21
Fix version selection

lunakv:
The route fetching card versions checks that each parameter passed to it is a valid UUID. "Extra" cards (such as back faces of TDFCs) have an extra "2" at the end of that ID. This PR fixes the ID validation so that it doesn't return errors when a cube includes these extra cards.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,5 @@
 webpack.*.js
 
-jobs/*.js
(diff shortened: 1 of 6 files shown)


lunakv:
(the checks are failing due to unused variables in the card fetch hotfix)

PR #1823, opened 2021-02-22
Don't open links to same page in new tab

### LINKED ISSUE #1821: Don't open links to same page in new tab
As a user, when adding a hyperlink to some markdown content, I'd like the link to not open a new tab if the destination is in the same URL as the origin. This can be useful for example for easy navigation within a markdown page when used in conjunction with heading autolinks.

lunakv:
Resolves #1821.

With this PR, markdown links that lead to the same page the document is rendered on (including query parameters) won't open in a new tab.


ruler501:
Does this work correctly with the link `#anchor` links which don't specify anything but the anchor?

lunakv:
It does now :)

PR #1827, opened 2021-02-27
Improve AnalyticTable percentages and freeze headers.

ruler501:


dekkerglen:
Can you please provide screenshots showing the different behavior

lunakv:
This looks like it works pretty well, but there are some visual quirks left to be ironed out:
### Chrome
- The border is thicker around the top row, but this doesn't extend to the corner cell, creating a weird inconsistency:
  Screenshot from 2021-03-04 17-30-54
- When scrolling up, you can see bits of the rows appear above the top row:
  Screenshot from 2021-03-04 17-31-13

### Firefox
A more problematic rendering. In FF, the sticky cells don't have a visible border between them, and the border between the top row and the first data row appears as two separate lines:
Screenshot from 2021-03-04 17-35-02

PR #1832, opened 2021-03-08
Changed "Random Cube" selection results

### LINKED ISSUE #1826: Improve random cube selection
As a user, when I hit random cube, I would like to see active and complete cubes more often

We can modify the query so it:
* excludes unlisted
* includes cubes updated within the last month
* includes cubes that have at least 360 cards

lunakv:
Fixes #1826.

Replaces the current random cube search with an aggregate pipeline that lets us easily add filter conditions to the search. The filter is currently set according to #1826, i.e. it returns only listed cubes of at least 360 cards updated in the last month. I'm not completely sure about the performance implications of this, but I'm guessing it won't be a performance bottleneck and we don't need to look at refactoring it until there's a problem. 

Note that this PR assumes we're running MongoDB 3.2 or newer, otherwise `sample` isn't available.

### DIFF
--- a/routes/root.js
+++ b/routes/root.js
@@ -102,10 +102,16 @@ router.get('/explore', async (req, res) => {
 });
 
 router.get('/random', async (req, res) => {
-  const count = await Cube.estimatedDocumentCount();
-  const random = Math.floor(Math.random() * count);
-  const cube = await Cube.findOne().skip(random).lean();
-  res.redirect(`/cube/overview/${encodeURIComponent(getCubeId(cube))}`);
+  const lastMonth = () => {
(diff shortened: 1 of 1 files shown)


dekkerglen:
We use DocumentDB in prod, I couldn't find an answer to whether or not that supports sample with an initial search

lunakv:
A search through Amazon's docs tells me they implement APIs from Mongo 3.6 and 4.0, and according to [this page](https://docs.aws.amazon.com/documentdb/latest/developerguide/mongo-apis.html#mongo-apis-aggregation-pipeline-stage) both `$match` and `$sample` are available in both.

Issue #310, opened 2019-09-09
Modules/Packages (like cycles but broader)


mvhsbball22:
As a user I would like to be able to create groups of cards that go together in some manner and share those groups with other people -- for example, a group of the "swords", a goblin sacrifice package, or an uncommon graveyard hate package.  

These groups would then show up on a page where users could browse submitted groups and rate them or vote on them or star them (whatever system makes sense site-wide).  These groups should be able to be added to a cube with a single click (like the land cycles). 

Additional features could also include sharing the groups.

The groups could be called packages, modules, groups, or something else.

mvhsbball22:
@purxiz This is the feature request that was mentioned in the #213 comments.

dekkerglen:
I like the name of packages. We could use this as a separate user generated content to supplement cube design. Adding pages to view, browse, and comment would be great.

SirManCub:
Agreed that Packages makes sense since many cube owners have packages that they use in some drafts, but remove or keep separate otherwise. Being able to easily incorporate it into a draft (without having to make a custom draft format) might be nice, too

purxiz:
Although I'm mentioned here, I don't think I"d really like to work on this, so if anyone else is interested feel free to hop in

Issue #213, opened 2019-08-31
Easy Add Cycles


dekkerglen:
As a user, I would like to easily be able to add common cycles, like lands.

Just, a button that allows one to add certain land cycles, or other commonly cubed cycles

purxiz:
Can you assign this to me?

emmett9001:
Is there a viable definition of a "cycle" that can be used for this purpose? It seems like kind of a nebulous concept since cycles don't necessarily have any metadata properties (CMC, name, power/toughness, etc) in common.

purxiz:
I was only going to add land cycles, which are pretty well defined. I don't
think other cycles are commonly cubed.

emmett9001:
Seems fine to start this with an option for "shocklands, fetchlands, abu duals, fastlands, etc". Some other common cycles in cubes are the Swords (Sword of War and Peace, Feast and Famine, etc), Forces (Force of Vigor, Force of Negation, Force of Valor), Titans (Frost Titan, Sun Titan, etc), Signets (Izzet Signet, Dimir Signet, etc), Moxen (Mox Emerald, Mox Jet, etc), and Eldrazi Titans (Emrakul, the Aeons Torn, Kozilek, Butcher of Truth, etc).

mvhsbball22:
Does it make sense to have this exist as a separate entity like a 'cube'
that people could add cards to and save? Then a user could visit the
'Cycles' page which could be sorted by things like "Frequency", "Cost",
etc.?

Doing it this way would fit into the social goals because people could
"star" cycles they liked.  It also avoids the problem of having card-like
metadata -- in fact, the metadata would be much closer to the cube object.

purxiz:
I'd rather they be two separate features. I'd like this feature request to
cover basically all the land cycles, and maybe some other generic cycles as
the commenter above mentioned, then we can maybe add a "modules" feature
for what you're implying. Personally, I'd rather keep any baked-in feature
to just lands, since basically every cube has lands, where as most cubes
(to the best of my knowledge) incomplete card cycles are more common than
complete ones.

I actually really like the modules idea though, and I think I'll work on it
next if you submit an issue

ruler501:
From discord discussion today on potential ideas for this:

> have it be packages not just cycles, things like power nine, rainbow lands, etc
> have a user voting system
> allow adding packages to vote on and they get added to the main list when they cross a certain vote threshold
> allow sorting by number of votes
[...]

PR #1837, opened 2021-03-10
Sort type constant

alexzhaohong:
This small PR aims to unify util/Sort.js to prepare it for custom user packages as mentioned in these issues:

User defined sort #509

Easy Add Cycles #213

Modules/Packages (like cycles but broader) #310

Summary:
* Does not change functionality
* Adds `CARD_TYPES` constant in `util/Sort.js`
* Modifies two sorts to use new constant

I ran `npm run test` but I am unfamiliar with how the tests are laid out. I visually inspected localhost:8080 and did not detect a difference in the error messages there.

My Next Steps:
* Define 'Land' category to have single color identity lands separate from colorless or multicolor gold lands

Future work:
* Add custom user labels and add admin-approved packages

### DIFF
--- a/src/utils/Sort.js
+++ b/src/utils/Sort.js
@@ -262,22 +262,7 @@ function getLabelsRaw(cube, sort) {
     return ['White', 'Blue', 'Black', 'Red', 'Green', 'Colorless'];
   }
   if (sort === 'Type') {
-    return [
-      'Creature',
-      'Planeswalker',
-      'Instant',
-      'Sorcery',
-      'Artifact',
-      'Enchantment',
-      'Conspiracy',
-      'Contraption',
-      'Phenomenon',
-      'Plane',
-      'Scheme',
-      'Vanguard',
-      'Land',
-      'Other',
-    ];
+    return CARD_TYPES.concat(['Other']);
   }
   if (sort === 'Supertype') {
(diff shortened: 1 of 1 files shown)


alexzhaohong:
Screen Shot 2021-03-09 at 5 35 36 PM
Screen Shot 2021-03-09 at 6 06 31 PM

These secondary sorts are "good enough" for me, but otherwise I do want to be able to sort lands. There seems to be conversations regarding how to implement this, so I created this first small change first just to get my feet wet, inline with the guidelines found at https://github.com/dekkerglen/CubeCobra/blob/master/CONTRIBUTING.md#your-first-contribution

ruler501 (approved):

dekkerglen:
I think this change is a good one, but I don't think that it helps move towards any of the issues that you linked.

For #509 - what we need to do is create a list of custom sorts saved on the cube object - the schema for this would just be essentially a dict where the key would be the category, and the value would be a piece of filter text specifying what cards get filtered in. We would just need to build out the UI to allow users to create (and preview) the custom sorts, and then adjust the cube sort so it brings in custom sorts from the cube context.

For the other two issues linked, these stories are distinct from how we display cubes. They would add a new collection in the DB where users can create and upload packages, then we would have some UI in Add/Remove where you can search for packages and add them to the cube. We would also want a way to browse packages and have a quick add to cube button

PR #1840, opened 2021-03-12
Add button to flip card on card pages

### LINKED ISSUE #1796: Card pages should handle DFCs better
Our current card page has problems displaying double-faced cards (picture below as an example). The page for a DFC's front face shows the name, cost, and type line of the front face only, then combines the oracle text of both faces, after which it displays the P/T or loyalty of the front face.

Aside from the aforementioned oracle text, the back face isn't indicated anywhere on the page except somewhat confusingly in the Versions list (and that's only for TDFCs) and in Autocard links.

As a user, I'd like to have a clearer separation between the front and back faces of a DFC and a clear link from one side of a DFC to the other. The issue of adequate separation also applies to split cards, which show both mana costs and names, but concatenate the oracle texts and show only the type line of the first part.

Screenshot from 2021-01-17 11-18-36

dekkerglen:
Fixes #1796 

image

image

Cards missing `image_back` attribute are unchanged:
image

### DIFF
--- a/src/pages/CardPage.js
+++ b/src/pages/CardPage.js
@@ -218,6 +218,7 @@ const CardPage = ({ user, card, data, versions, related, cubes, loginCallback })
   const [selectedTab, setSelectedTab] = useQueryParam('tab', '0');
   const [priceType, setPriceType] = useQueryParam('priceType', 'price');
   const [cubeType, setCubeType] = useQueryParam('cubeType', 'total');
+  const [imageUsed, setImageUsed] = useState(card.image_normal);
 
   const sortedVersions = versions.sort((a, b) => {
     const date1 = new Date(a.released_at);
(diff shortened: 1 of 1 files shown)


joshfinnie (approved):
looks good

lunakv:
I would very much prefer "Transform" over "Flip". Not only is it the correct term rules-wise, it just feels like generally a nicer word. We could even add some icon to it, Scryfall style (since we're already using octicons, [this one](https://octicons-primer.vercel.app/octicons/arrow-switch-16) might be an option)

lunakv:
Here's a mockup of about what I had in mind (I can make a PR with this change if that looks interesting) :
transform_mt1

lunakv:
This branch should be merged again, because the merge didn't include changes from #1841.

ruler501:
Can we add this to the card modal as well?

dekkerglen:
Good catch, I thought the other PR was merged into master, so this needed to be merged first

PR #1843, opened 2021-03-17
[styling] - Remove scrolling on direct blog posts

alexzhaohong:
Before: 
https://www.loom.com/share/e6b07059a92e44adb656838fac1edb50

After: 
https://www.loom.com/share/e6b07059a92e44adb656838fac1edb50

* Only prohibits scrolling on permalink pages. The cube blog index page is still set to 50vh max
* `noScroll` doesn't seem like a great flag name for me. Let me know if you have a better suggestion.
* the `div` style was set right on the page. I wanted to make minimal changes but I assume this is not best practice. Let me know if this is best practice. I am unsure where all the css files are.

### DIFF
--- a/src/components/BlogPost.js
+++ b/src/components/BlogPost.js
@@ -54,10 +54,12 @@ class BlogPost extends React.Component {
   }
 
   render() {
-    const { post, onEdit } = this.props;
+    const { post, onEdit, noScroll } = this.props;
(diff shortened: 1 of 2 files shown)


lunakv:
Just FYI, both links in your description lead to the same gif, so that doesn't really help show what's been done here :)

PR #1854, opened 2021-03-28
Add Job to automatically update podcasts

dekkerglen:
I spun up a new server to run periodic jobs, which currently is configured to update podcasts every ten minutes and populate analytics every 24 hours. That machine is currently running off of this branch.

Also moves winston configuration into separate file so the app config can be re-used for jobs.

### DIFF
--- a/jobs/populate_analytics.js
+++ b/jobs/populate_analytics.js
@@ -394,7 +394,7 @@ const run = async () => {
 
   winston.info('finished loading cards');
 
-  const ratings = await CardRating.find({}).lean();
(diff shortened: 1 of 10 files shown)


jobs/populate_analytics.js
```diff
@@ -460,5 +468,9 @@ mongoose
     useUnifiedTopology: true,
   })
   .then(() => {
-    run();
+    try {
+      run();
+    } catch (err) {
+      console.error(err);
```
lunakv, line 474: Shouldn't this be logged through winston as well?

PR #1864, opened 2021-04-05
Cache a promise for generating pack images

nex3:
This means that when N requests for the same image come in quick
succession (for example, when a pack image is displayed on Discord),
the server doesn't need to render the image N times. Instead, it
renders it once, then sends the resulting data out N times,
substantially decreasing the server load and increasing
responsiveness.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5027,6 +5027,11 @@
         "wordwrap": "0.0.2"
       }
     },
+    "clone": {
+      "version": "2.1.2",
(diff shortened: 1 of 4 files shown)


dekkerglen:
Need to run `npm run beautify` but otherwise looks good, this is actually our worst performing route, aside from expensive cube searches.

Couple caveats:
- We use a load balanced fleet so a cache on the node level will minimize the # of computations to the size of the fleet, not to a single call.
- A cubeid & seed tuple produces a unique pack only for that version of a cube - with a TTL of 5 min it shouldn't be a problem, but something to keep in mind

PR #1865, opened 2021-04-05
Add analytic for playtest data aggregation

### LINKED ISSUE #46: Stats on Decks drafted
As a user, I would like to be able to see statistics on: Mainboard rate, sideboard rate, and pick rate for each card in my cube.

Store this information on draft submit and on draft pick update. Add a tab under analysis that displays this information.

### LINKED ISSUE #1295: Table analytic .csv export
As a user, I would like to be able to export the table analytic as a .csv file

dekkerglen:
Fixes #46, Fixes #1295

Adds in passive tracking on a per-cube basis of picks, passes, mainboards, sideboards, and also adds cube specific elo (but doesn't use it anywhere other than analytics).

Also adds these fields for other analytics through the 'attributes' object

Also adds a little link to download csv for contents of any sortable table

New Analytic:
image
 
New fields in pivot table:
image

New field example in averages:
image

### DIFF
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -330,6 +330,79 @@ async function compareCubes(cardsA, cardsB) {
   };
 }
 
+const newCardAnalytics = (cardName, elo) => {
+  return {
+    cardName,
+    picks: 0,
+    passes: 0,
+    elo,
+    mainboards: 0,
+    sideboards: 0,
+  };
+};
+
+const removeDeckCardAnalytics = (cube, deck, carddb) => {
+  cube.numDecks -= 1;
+
+  for (const col of deck.seats[0].deck) {
+    for (const current of col) {
+      let pickIndex = cube.cardAnalytics.findIndex(
+        (card) => card.cardName.toLowerCase() === carddb.cardFromId(current.cardID).name.toLowerCase(),
+      );
+      if (pickIndex === -1) {
+        pickIndex =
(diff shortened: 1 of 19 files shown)


serverjs/cubefn.js
```diff
@@ -517,6 +590,16 @@ const methods = {
       pack: draft.initial_state[0][0],
     };
   },
+  newCardAnalytics,
+  getEloAdjustment: (winner, loser) => {
+    const diff = loser - winner;
+    // Expected performance for pick.
+    const expectedA = 1 / (1 + 10 ** (diff / 400));
+    const expectedB = 1 - expectedA;
+    const adjustmentA = 2 * (1 - expectedA);
```
ruler501, line 599: This scaling factor needs to be larger for cubes. If card a starts equal elo to card b it takes roughly(approximated with a continuous estimate) 965/scaling comparisons between the two before a's elo has grown by 100. Obviously our implementation would likely do this faster since it compares against multiple cards at once but it still would probably take at least 32/scaling first picks to have it grow by 100. In general to go up x points it takes an exponential number of picks and is inversely related to the scaling factor. It would be a good idea to differentiate cards much quicker since cubes are large enough you won't be able to impact card ratings that much otherwise.

src/pages/CubeAnalysisPage.js
```diff
@@ -126,6 +154,10 @@ const CubeAnalysisPage = ({
         />
       ),
     },
+    {
+      name: 'Playtest',
```
ruler501, line 158: I do not like this name. It is a verb and means something different from the other places we use it.

ruler501 (changes requested):

PR #1866, opened 2021-04-05
Issue 196

### LINKED ISSUE #196: "Remove" preview card image does not match included version
When hovering card names in the autocomplete list for the Remove text field in the Add/Remove view, the card preview image does not always match the version of the card included in the cube.

1. Navigate to List, then Add/Remove
2. Type any card name in the Card to Add field
3. Start typing the name of a card in the cube with at least two different artwork releases in the Card to Remove field
4. Mouseover the name of the desired card in the autocomplete list
5. Observe that the card preview image is the most recent printing as opposed to the image configured for the cube

In this situation I'd expect the preview image to match the image I've selected for the given card in my cube.

- OS: Ubuntu 16.04
- Browser: Chrome 70.0.3538.77

dekkerglen:
Fixes #196 

For a _long_ time, we actually had autocard not working on the drop down select here - this adds it back as well.

This also adds a checkbox to enable specifying card versions by set code and collector's id, turned off by default.

Showing correct version:
image

With checkbox off (default, same as old behavior, except with autocard now):
image

With checkbox on:
image

### DIFF
--- a/src/components/EditCollapse.js
+++ b/src/components/EditCollapse.js
@@ -14,6 +14,7 @@ import {
   Input,
   Card,
   FormText,
+  InputGroupText,
 } from 'reactstrap';
 
 import { encodeName } from 'utils/Card';
@@ -61,6 +62,7 @@ const EditCollapse = ({ ...props }) => {
   const [alerts, setAlerts] = useState([]);
   const [postContent, setPostContent] = useState('');
   const [mentions, setMentions] = useState('');
+  const [specifyEdition, setSpecifyEdition] = useState(false);
 
   const {
     changes,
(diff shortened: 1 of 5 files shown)


src/components/EditCollapse.js
```diff
@@ -216,6 +219,22 @@ const EditCollapse = ({ ...props }) => {
...
+                  <Input
+                    addon
+                    type="checkbox"
+                    aria-label="Checkbox for following text input"
+                    checked={specifyEdition}
+                    onChange={() => setSpecifyEdition(!specifyEdition)}
+                  />
+                </InputGroupText>
+              </InputGroupAddon>
+              <Input disabled value="Specify Versions" />
```
alexzhaohong, line 235: I’m not familiar with what this line, Input component with a disabled value string, does.

alexzhaohong (approved):
Huzzah!

src/components/EditCollapse.js
dekkerglen: Usually you use this to dynamically disable inputs but I just like the styling of this component so I'm using an input as a label. I won't defend it as best practice but I think it's fine here

PR #1869, opened 2021-04-08
CMC -> Mana Value

lunakv:
This PR changes the user-visible components with the text "CMC" or "Converted Mana Cost" to display the term "Mana Value" instead.

### What Wasn't Changed
An important distinction is that this PR only affects user-facing front-end values. Things like variable names, database fields, or route parameters were kept unchanged, as was the CSV header field. Since I'm not in the business of throwing away my time, I decided those things weren't worth changing.

### What Was Changed
- **The Card Modal:** the modal had a field called "CMC". That field was renamed to "Mana Value". Same for the corresponding field on the **Group Modal**.
- **The Filter**: A new `mv:` filter was added, and the text in the **Filter Modal** was adjusted to use "Mana Value". The documentation on the filter page was updated accordingly.
- **The Sorts**: All CMC-based sorts were renamed to their "Mana Value" equivalents, and in-code usage of sorting functions was appropriately adjusted.
- **List View**: The list view header was changed to say "MV" instead of "CMC". This is the only place that uses a shorthand form of "Mana Value" (aside from the filter), since the full phrase wouldn't fit on one line in the header and caused a jarring visual inconsistency.
- **unrelated**: fixed a bug where the server could crash if someone passed a non-numeric mana value from a card modal.

### Compatibility
To ensure that existing setups don't break too much, a few additional measures were taken to ensure some form of backwards compatibility.
- The `cmc:` filter is still available as an alternative to `mv:`.
- The `SortCollapse` and components using it will automatically translate CMC sorts to Mana Value sorts. This means that existing sort URL parameters will continue to work normally, as will the default values stored in database. Other components that require sorting, namely the analytics tabs, aren't backwards compatible. I'm of the opinion that the added complexity of trying to keep all of those backwards compatible would outweigh the benefits of that compatibility. Most players (I estimate) don't use analytics with a saved sort parameter. Those who do will not see a severe error, but simply an empty table, which can be easily corrected by changing/removing the sort value.
- As previously mentioned, all object fields, database fields, internal variables, HTML element arguments, and route parameters were left unchanged.

### DIFF
--- a/src/utils/Sort.js
+++ b/src/utils/Sort.js
@@ -145,8 +145,8 @@ export function GetColorCategory(type, colors) {
 
 export const SORTS = [
   'Artist',
-  'CMC',
-  'CMC2',
+  'Mana Value',
+  'Mana Value 2',
   'Color Category',
   'Color Category Full',
   'Color Count',
@@ -186,7 +186,7 @@ export const SORTS = [
   'Unsorted',
 ];
 
-export const ORDERED_SORTS = ['Alphabetical', 'CMC', 'Price'];
+export const ORDERED_SORTS = ['Alphabetical', 'Mana Value', 'Price'];
 
 const allDevotions = (cube, color) => {
   const counts = new Set();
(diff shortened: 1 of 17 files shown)


src/utils/Sort.js
```diff
@@ -145,8 +145,8 @@ export function GetColorCategory(type, colors) {
 
 export const SORTS = [
   'Artist',
-  'CMC',
-  'CMC2',
+  'Mana Value',
```
alexzhaohong, line 148: I'm glad you went with 'Mana Value' and not "MV" and "MV2"

alexzhaohong (approved):
I believe the breaking change for saved sorting is small enough that we can go ahead and commit to this new-age blasphemy

src/utils/Sort.js
lunakv: Yeah, I'm not a fan of using "MV" if it isn't necessary.

lunakv:
And just to clarify, the change is actually non-breaking for saved sorts on cubes. It's basically only breaking for those three people that have an Analytics bookmark with an explicit sort parameter (set to CMC).

PR #1870, opened 2021-04-09
Create process_decks.js

dekkerglen:
This PR adds a script that will set the initial state of each cube's respective cubeAnalytics document by processing all decks and drafts.

### DIFF
--- a/jobs/process_decks.js
+++ b/jobs/process_decks.js
@@ -86,7 +86,8 @@ const processDraft = (draft, deck, analytic) => {
 
 const processDeck = (deck, draft, analytic) => {
   try {
-    if (!deck.seats[0].bot) {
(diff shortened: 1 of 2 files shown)


jobs/process_decks.js
```diff
@@ -0,0 +1,159 @@
...
+      const [pick] = pack.splice(0, 1);
+      processPick(pick, pack, analytic);
+    }
+  }
+};
+
+const processDeck = (deck, draft, analytic) => {
+  try {
+    if (!deck.seats[0].bot) {
+      for (const col of deck.seats[0].deck) {
```
ruler501, line 90: You should exclude decks with an empty sideboard.

ruler501 (changes requested):

PR #1876, opened 2021-04-10
Remove space in csv export for tags

ruler501:


serverjs/cubefn.js
```diff
@@ -271,7 +271,7 @@ function CSVtoCards(csvString, carddb) {
         finish: finish || 'Non-foil',
         imgUrl: (imageUrl || null) && imageUrl !== 'undefined' ? imageUrl : null,
         imgBackUrl: (imageBackUrl || null) && imageBackUrl !== 'undefined' ? imageBackUrl : null,
-        tags: tags && tags.length > 0 ? tags.split(',') : [],
+        tags: tags && tags.length > 0 ? tags.split(';').map((t) => t.strip()) : [],
```
lunakv, line 274: Is `strip` some kind of undocumented string function I don't know, or is this supposed to be `trim`?

routes/cube_routes.js
```diff
@@ -1427,7 +1427,7 @@ function writeCard(res, card, maybe) {
   res.write(`${imgBackUrl},"`);
   card.tags.forEach((tag, tagIndex) => {
     if (tagIndex !== 0) {
-      res.write(', ');
+      res.write(';');
```
lunakv, line 1430: It might be a good idea to also prevent user-made tags from the UI from containing semicolons, otherwise exporting and re-importing a cube will mess things up.

ruler501:
I forgot about this pr. I'll fix it up later today.

routes/cube_routes.js
ruler501: I think that's good for a future PR since we don't prevent commas currently.

PR #1885, opened 2021-04-13
Adding workflow for pushing to dev server

dekkerglen:


.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
```
ruler501, line 8: I would prefer
```
  workflow_run:
      workflows: ["CI Tests"]
      branches: [master]
      types: 
        - completed
 ```
Then add this condition to the task
```
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
````
You'll still be able to do manual deploys but the automated ones will only run if CI passes.

File: .github/workflows/main.yml [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
```
ruler501, line 20: This is where you'd add the condition from above.

.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Node ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Installing NPM
+        run: npm install
+
+      - name: Building application
+        run: npm run build
```
ruler501, line 43: We should really setup incremental builds with caching, but that'll be a separate task, probably alongside upgrading to webpack 4.
https://webpack.js.org/guides/build-performance/

.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Generate deployment package
+        run: zip -r deploy.zip * -x node_modules/\* -x .\* -x private/\*
```
ruler501, line 46: Could we add this as a build artifact, might be useful if we need to debug something?

.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
```
ruler501, line 3: I like the name `Deploy Master` for this.

ruler501 (changes requested):

.github/workflows/main.yml
```diff
@@ -1,12 +1,12 @@
...
 
-name: CI
+name: Deploy master to dev server
 
-# Controls when the action will run. 
-on:
-  # Triggers the workflow on push or pull request events but only for the master branch
-  push:
-    branches: [ master ]
+workflow_run:
```
ruler501, line 5: You still need this to be in the `on:` block.

ruler501 (changes requested):
For uploading an artifact you'd add this as a step right before the deploy step.
```
- name: Upload Deployment Artifacts
  uses: actions/upload-artifact@v2
  with:
    name: deploy-zip
    path: deploy.zip
    if-no-files-found: error
```

ruler501 (approved):

PR #1914, opened 2021-05-12
Update the export scripts to store to a file system

ruler501:
so we can run them through zstd before uploading.

Also added a bunch of extra data to them for the cube map and other post-processing.


dekkerglen:
This feels like a net negative - I still like the ability for the script to directly upload to s3

ruler501:
This lets us compress all the exports into one archive though.

dekkerglen:
It lets us do that - but it doesn't do that. If this PR didn't remove existing functionality it wouldn't be a problem, but right now it adds more overhead to the existing scripts - instead it should tar the files and still upload those files

jobs/definition/weekly.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/sh
 node jobs/populate_embeddings.js
 node jobs/clean_cubes.js
-node jobs/download_cubes.js
+rm -r exports/
+node --max-old-space-size=8192 download_drafts.js
+node --max-old-space-size=8192 download_cubes.js
+node --max-old-space-size=8192 download_decks.js
+tar -cJf exports.tar.xz exports/
+aws s3 cp exports.tar.xz s3://cubecobra/exports.tar.xz
```
dekkerglen, line 9: I want to do this via node, not the awscli so it's less dependent on deployment details, such as how the aws profile is configured, instead of using the credentials in `.env`

ruler501: I can add a script that does that last step in node I guess, tar is much better done from the command line though I believe.

PR #1917, opened 2021-05-12
Default to pull carddb from s3 unless nos3 is passed.

ruler501:
If the correct env variables are defined for s3 access and the script wasn't passed the 'nos3' arg it will default to downloading from s3.

### DIFF
--- a/force_update.js
+++ b/force_update.js
@@ -1,7 +1,9 @@
 // Load Environment Variables
 require('dotenv').config();
-
+const AWS = require('aws-sdk');
+const fs = require('fs');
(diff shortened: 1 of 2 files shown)


lunakv:
This change by itself doesn't do much, because the `updateCardbase` function will still just download files from Scryfall. We also need to modify `serverjs/updatecards.js` to actually take advantage of this local storage.

dekkerglen:
Closing b/c finished implementation is here: https://github.com/dekkerglen/CubeCobra/pull/1926

PR #1919, opened 2021-05-14
Wiki Prototype

jbts:
This PR contains some work on a wiki prototype or proof of concept.

It adds some new routes, like `cubecobra.com/wiki/home` and `cubecobra.com/wiki/site_help/custom_draft_formats` that display wiki pages with information about different topics about the site.

The wiki pages are in a new `wiki/` directory in the GitHub repo and are written in markdown. The route `cubecobra.com/wiki/<path>` attempts to read in the file `wiki/<path>.md` and parse its contents with the `Markdown` component. `<path>` is constrained to match the regular expression `[A-Za-z/_]+` for security reasons (in particular, no dots `.` are allowed). The empty path, corresponding to `cubecobra.com/wiki`, is just redirected to `cubecobra.com/wiki/home`. There may still be some security issues with the way reading in files is done, I'm not sure exactly.

This PR isn't necessarily intended to be merged, but more for thinking about whether this might be a good way to implement a wiki and get some feedback. I'm not very experienced with web programming, so a lot of the route code is based on the other route code in the repository. There definitely might be some mistakes. Errors and bad routes aren't handled particularly well right now.

Some screenshots (note these wiki pages are just for fun and testing):

Home page
image

Another page (`wiki/create_cube`)
image

Route that doesn't exist
image

### DIFF
--- a/app.js
+++ b/app.js
@@ -178,6 +178,7 @@ app.use('/comment', require('./routes/comment_routes'));
 app.use('/admin', require('./routes/admin_routes'));
 app.use('/content', require('./routes/content_routes'));
 app.use('/packages', require('./routes/packages'));
+app.use('/wiki', require('./routes/wiki_routes'));
 
 app.use((req, res) => {
   return render(req, res, 'ErrorPage', {
--- a/routes/wiki_routes.js
+++ b/routes/wiki_routes.js
@@ -1,0 +1,31 @@
+const express = require('express');
+
+const fs = require('fs');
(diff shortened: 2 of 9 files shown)


ruler501:
This fails to solve the problem that a custom wiki was suggested for, namely not requiring a github account to make changes.

My suggestions still stay the same. We should just use github wiki instead of building our own solution. Alternatively we could use a pre-built open source solution, either way I do not think a full wiki is worth the dev time and maintenance burden.

PR #1920, opened 2021-05-14
Rework scheduled jobs

dekkerglen:
This PR adds four new npm aliases that run the respective `sh` scripts in `jobs/definition/`. The idea is we have different granularities of scheduled jobs, and contributors can add the jobs to the respective sh script. 

This PR also adds the job `clean_cubes` which cleans up some errored data automatically.

This PR also updates the export cubes job to fit the specifications for Lucky Paper's cube map.

This PR also archives some scripts that were used in the past that are not needed, but I think it's useful to include in the repo as references.


jobs/clean_cubes.js
```diff
@@ -0,0 +1,77 @@
...
+    cube.cards = cube.cards.filter((c) => c && !isInvalidCardId(c.cardID));
+
+    await cube.save();
+  }
+};
+
+try {
+  (async () => {
+    await carddb.initializeCardDb();
+    mongoose.connect(process.env.MONGODB_URL).then(async () => {
```
lunakv, line 46: Since this is already an `async` function, we could save ourselves a level of nesting by just `await`ing the `connect` call and getting rid of the `then` block

jobs/clean_cubes.js
```diff
@@ -0,0 +1,77 @@
...
+      // process all cube objects
+      console.log('Started');
+      const count = await Cube.countDocuments();
+      const cursor = Cube.find().lean().cursor();
+
+      // batch them in 100
```
lunakv, line 52: `batchSize` is 1000, not 100

jobs/definition/constantly.sh
```diff
@@ -0,0 +1,2 @@
+git pull
```
lunakv, line 1: I would put `#!/bin/sh` in all of the shell scripts. It's not strictly necessary, but it's best practice.

jobs/download_cubes.js
```diff
@@ -1,63 +1,103 @@
...
+try {
+  (async () => {
+    await carddb.initializeCardDb();
+    mongoose.connect(process.env.MONGODB_URL).then(async () => {
+      // process all cube objects
+      console.log('Started');
+      const count = await Cube.countDocuments();
+      const cursor = Cube.find().lean().cursor();
+
+      // batch them in 100
```
lunakv, line 66: `batchSize` is 1000

jobs/download_cubes.js
```diff
@@ -1,63 +1,103 @@
...
+    name,
+    owner,
+    owner_name,
+    tags,
+    image_uri,
+    image_artist,
+    image_name,
+    card_count,
+    users_following,
+    isListed,
```
lunakv, line 35: I'm not sure what else these exports are used for, but we should maybe consider just not dumping the cubes that are unlisted.

jobs/clean_cubes.js
lunakv: We should also pass `{ useNewUrlParser: true }` into the `connect` calls to get rid of mongoose warnings.

PR #1921, opened 2021-05-14
Create a new daily job to upload card files to s3

dekkerglen:
This PR adds a new job that will upload the contents of `private` to `S3://cubecobra/cards/` daily, so that when instances are cold-starting, they can pull those files and don't have to do any processing.


jobs/update_cards.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable no-await-in-loop */
+// run with: node --max-old-space-size=8192 populate_analytics.js
```
lunakv, line 2:
```suggestion
// run with: node --max-old-space-size=8192 update_cards.js
```

jobs/definition/daily.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 npm run cards
-node --max-old-space-size=32000 jobs/populate_analytics.js
\ No newline at end of file
+node --max-old-space-size=32000 jobs/populate_analytics.js
+node jobs/update_cards.js
```
lunakv, line 4: I'm a fan of ending files with a newline; it makes for nicer diffs.

jobs/update_cards.js
lunakv, line 2:
```suggestion
// run with: node --max-old-space-size=8192 update_cards.js
```

PR #1926, opened 2021-05-16
Have nodes use s3 by default for card definitions

dekkerglen:
This PR adds the env variable `USE_S3` which in the example env is set to true. If this value is present, cards will always be pulled from the s3 bucket instead of creating them ad hoc. This includes the daily update as well. If this field is not present, then the old process will occur.


jobs/populate_analytics.js
```diff
@@ -408,7 +408,6 @@ const run = async () => {
   winston.info('finished loading cards');
 
   const ratings = await CardRating.find({}).lean();
-  // await updatedb.updateCardbase(ratings);
```
ruler501, line 411: Remove the commented out code.

serverjs/updatecards.js
```diff
@@ -821,6 +817,42 @@ async function updateCardbase(ratings = [], basePath = 'private', defaultPath =
...
+  winston.info('Finished downloading files from S3...');
+};
+
+async function updateCardbase(ratings = [], basePath = 'private', defaultPath = null, allPath = null) {
+  if (!fs.existsSync(basePath)) {
+    fs.mkdirSync(basePath);
+  }
+  winston.info('Updating cardbase, this might take a little while...');
+
+  if (process.env.USE_S3) {
```
ruler501, line 851: Do a string comparison with `'true'` here so defining it as "false" wouldn't run with S3.

.env_EXAMPLE
```diff
@@ -11,3 +11,4 @@ FLASKROOT="http://167.172.233.110"
 PORT="5000"
 HOST="http://127.0.0.1:5000"
 CUBECOBRA_VERSION="3.4.1"
+USE_S3="true"
```
ruler501, line 14: Default should be false since there are no aws credentials in the default env file.

ruler501 (changes requested):

PR #1964, opened 2021-06-03
Update schemas to be strict again

dekkerglen:


models/cube.js
```diff
@@ -83,21 +81,7 @@ const cubeSchema = mongoose.Schema({
...
-        packs: [
-          {
-            slots: [String],
-            steps: stepsSchema,
-          },
-        ],
-      },
-      */
-    ],
+    type: [stepsSchema],
```
ruler501, line 84: This is wrong isn't it?

PR #1970, opened 2021-06-03
Show custom formats on the playtest page

lunakv:
Under the old model, the drafts from the database were transformed into the `draftFormats` property. Since there's now no need for that, the transformation was removed, however a reference to `draftFormats` was left in the playtest page, causing `formats` to always be empty.

### DIFF
--- a/src/pages/CubePlaytestPage.js
+++ b/src/pages/CubePlaytestPage.js
@@ -431,7 +431,7 @@ const DEFAULT_FORMAT = {
 };
 const CubePlaytestPage = ({ user, cube, decks, loginCallback }) => {
   const { alerts, addAlert } = useAlerts();
-  const [formats, setFormats] = useState(cube.draftFormats ?? []);
+  const [formats, setFormats] = useState(cube.draft_formats ?? []);
   const [editModalOpen, setEditModalOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


src/pages/CubePlaytestPage.js
```diff
@@ -585,12 +585,7 @@ CubePlaytestPage.propTypes = {
     defaultDraftFormat: PropTypes.number,
     _id: PropTypes.string.isRequired,
     owner: PropTypes.string.isRequired,
-    draftFormats: PropTypes.arrayOf(
-      PropTypes.shape({
-        title: PropTypes.string.isRequired,
-        markdown: PropTypes.string,
-      }),
-    ),
+    draft_formats: PropTypes.arrayOf(PropTypes.object),
```
ruler501, line 588: I'd like this to be more detailed.
```
PropTypes.shape({
  title: PropTypes.string,
  multiples: PropTypes.boolean,
  markdown: PropTypes.string.isRequired,
  packs: PropTypes.arrayOf(PropTypes.shape({
    slots: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
    steps: stepsSchema, // This should be in one of the proptypes already
  }).isRequired).isRequired,
}).isRequired).isRequired
```

ruler501 (changes requested):

PR #2004, opened 2021-06-05
Remove unnecesary card name to lowercase

### LINKED ISSUE #1984: Custom draft format not transitioning to deckbuilding
https://cubecobra.com/cube/playtest/wtwlf123

Glimpse Draft

dekkerglen:
Fixes #1984

Somehow cube analytics can have a null cardname, but all the card names are always inserted as .toLowerCase(), so this change requires no data cleaning and will fix the behavior

### DIFF
--- a/jobs/archived/process_decks.js
+++ b/jobs/archived/process_decks.js
@@ -91,7 +91,7 @@ const processDeck = (deck, draft, analytic) => {
       for (const col of deck.seats[0].deck) {
         for (const current of col) {
           let pickIndex = analytic.cards.findIndex(
-            (card) => card.cardName.toLowerCase() === carddb.cardFromId(current.cardID).name.toLowerCase(),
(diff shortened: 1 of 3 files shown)


lunakv:
I'm not convinced this actually fixes #1984. It might be a contributing factor, but I don't see it being the only one. We should spend some time actually root-causing this before the issue is closed.

lunakv:
We should also think about how the null card names got there and whether to clean them up.

dekkerglen:
This fixes the reports I've gotten, There is a separate issue with multiple form submissions, but #1984 is due to the errored state of cubeAnalytics

lunakv:
I've been able to replicate #1984 locally running from this branch, so I remain unconvinced that this is enough to fully close it.

lunakv:
(it's also entirely possible that my local db is just messed up though, which is why I'd love to see a full breakdown of this)

PR #2019, opened 2021-06-09
Rework Top Cards

dekkerglen:
The issue: Top Cards queries are incredibly slow, because they have to query the _entire_ card history table. 

This PR takes the fields that are used by top cards, and loads them into the cached carddb, removing the need for any DB queries. 

Three new filters have been added: `cubecount, pickcount, and popularity`

The logic of top cards is now merged with search cards, except when using top cards a `pickcount<${MIN_PICKS}` is added as a filter to replicate previous functionality.

It also reworks how we do ordered sorting, to keep that logic all in one place across top cards, search cards, and cube ordered sorting. I've added the ordered sorts `Pick Count, Cube Count, Popularity, Release Date, Elo`.

This PR will significantly increase availability and decrease latency, as this is currently our highest latency DB call, frequently taking over a minute.


alexzhaohong (approved):
What will be the latency after this change?

dekkerglen:
The latency I was referring to was the DB call latency, I'm removed the DB call completely.

dekkerglen:
Notably, this is the reason why top cards is "down" on production

PR #2061, opened 2021-06-17
Release 5.4

dekkerglen:
New Features:
- Added popularity sort and filter
- Fully implemented cube tags
- Added Patreon integration
- Added cube count and pick count sorts and filters

Tweaks:
- Changed advertisment banner background color

Bugfixes:
- Top cards reworked and fixed
- Cube querying improved
- Fixed anonymous draft notifications
- Fixed decklist uploading
- Fixed an issue where the last pick of drafts wouldn't be submitted
- Fixed an issue that prevented cube elo from being updated


jobs/clean_cubes.js
```diff
@@ -21,7 +21,11 @@ const DEFAULT_BASICS = [
 const BATCH_SIZE = 1024;
 
 const needsCleaning = (cube) =>
-  !cube.cards || !Array.isArray(cube.basics) || cardsNeedsCleaning(cube.cards) || cardsNeedsCleaning(cube.maybe);
+  !cube.cards ||
+  !Array.isArray(cube.basics) ||
+  cardsNeedsCleaning(cube.cards) ||
+  cardsNeedsCleaning(cube.maybe) ||
+  cube.tags.some((tag) => !tag || tag.toLowerCase() !== tag);
```
lunakv, line 28: Do we really want to disallow tags with uppercase characters?

lunakv: If anything, I'd argue that a better check here would be one that would clean up leading/trailing whitespace.

models/cardHistory.js
```diff
@@ -58,11 +57,6 @@ const cardHistorySchema = mongoose.Schema({
   },
 });
 
-cardHistorySchema.index({ oracleId: 1, 'current.rating': 1 });
-cardHistorySchema.index({ oracleId: 1, 'current.elo': -1 });
-cardHistorySchema.index({ oracleId: 1, 'current.picks': -1 });
-cardHistorySchema.index({ oracleId: 1, 'current.cubes': -1 });
```
lunakv, line 64: Just flagging this to mention that this change won't (I believe) remove the already built indexes from the database - those have to be deleted manually.

jobs/clean_cubes.js
lunakv: (I just realized this is for cube tags and not card tags, so it isn't that big of a deal. We should still remove the whitespace though.)

models/user.js
```diff
@@ -97,6 +97,10 @@ const UserSchema = mongoose.Schema({
     type: Boolean,
     default: false,
   },
+  patron: {
```
lunakv, line 100: What's the purpose of this field? It doesn't appear to be set anywhere in the patreon routes.

models/cardHistory.js
dekkerglen: This is correct, this is something I handle manually

jobs/clean_cubes.js
dekkerglen: Yes, because it makes querying by tags either much more expensive, or makes it case sensitive

models/user.js
dekkerglen: Created a PR:https://github.com/dekkerglen/CubeCobra/pull/2063

PR #2062, opened 2021-06-18
Disable ads for patrons

lunakv:
This PR makes sure that users that are linked as patrons on Patreon don't see ad messages on the site. It also adds an ad mentioning that fact.


dekkerglen:
Might be time to build out that `UserContext` we talked about

PR #2064, opened 2021-06-18
Add UserContext

### LINKED ISSUE #1573: Add UserContext
We use user info a lot, it would be great to be able to have a user context so we don't have to pass through all the user objects, and user ids all the way down.

dekkerglen:
Fixes #1573

Before having a `userContext`, we passed around `user` objects and `userid` into components, and it was quite messy and not necessary. Many components (such as all the have comments) required this field in order to function, and essentially just had a prop parameter solely just to pass through to it's respective comments section.

This PR userContext, which gets populated in `mainLayout`, and is across many components. Notably, this removes the `canEdit` pattern that we use in multiple places, and also cleans up a lot of ownership checks across the entire site.

### DIFF
--- a/src/components/DeckPreview.js
+++ b/src/components/DeckPreview.js
@@ -27,8 +27,6 @@ const truncateToLength = (len, s) => {
 };
 
 const DeckPreview = ({ deck, nextURL }) => {
-  console.log(deck);
-
   const user = useContext(UserContext);
   const canEdit = user.id === deck.owner || user.id === deck.cubeOwner;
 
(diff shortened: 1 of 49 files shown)


src/components/DeckPreview.js
```diff
@@ -23,7 +26,12 @@ const truncateToLength = (len, s) => {
   return s.length > len ? `${s.slice(0, len - 3)}...` : s;
 };
 
-const DeckPreview = ({ deck, canEdit, nextURL }) => {
+const DeckPreview = ({ deck, nextURL }) => {
+  console.log(deck);
```
lunakv, line 30: Errant `console.log` left.

src/contexts/UserContext.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const GroupModalContext = React.createContext();
```
lunakv, line 3: Constant should be renamed.

PR #2071, opened 2021-06-20
Separate analytics function into separate file

dekkerglen:
This fixes an issue with importing. For some reason having these functions in the same module breaks some of the imports. I don't know why this fixes it, but it does.


lunakv:
I think #2072 might be a cleaner solution to this

dekkerglen:
You're right, appreciate you actually finding the problem

PR #2075, opened 2021-06-20
Fix cube card count on bulk upload

### LINKED ISSUE #2033: Have CSV export enable "Show Unsorted Cards"
From discord:

> When I set my primary sort to "Tag", the export function only exports cards with tags and ignores untagged cards. This is true even if I have the "Show Unsorted Cards" feature enabled so that the display shows my untagged cards. Since my process for building this cube is very tag-centric I'd prefer not to have to change my sort each time I export+import (which is the quickest way for me to add tags) NOTE: This appears to happen in all 4 views (table/list/etc.) but I only tested CSV export

### LINKED ISSUE #2042: Have more exports use current sorting
Currently, .txt exports (and I assume other non-csv exports) don't use the current active sorts. CSV does, and we should preserve the sorting to the other exports as well.

### LINKED ISSUE #2074: Importing cards incorrectly setting cube card count
Create a new cube, go to import -> paste text, and put in `10x Island`. 10 islands will be added, but the cube card count will remain 0

dekkerglen:
Fixes #2074, Fixes #2042, Fixes #2033

In 5.4, we introduced a pre-save hook to the cube schema to update the cube's card_count, but it doesn't trigger off of update calls. This line adds ensures it will always be done.

Edit:
Accidentally pushed to this branch, this PR now also has all cube downloads respect selected sorts, and also respects the "showther" parameter

### DIFF
--- a/src/components/CubeListNavbar.js
+++ b/src/components/CubeListNavbar.js
@@ -288,7 +288,7 @@ const CubeListNavbar = ({
 
   const { canEdit, cubeID, hasCustomImages } = useContext(CubeContext);
   const { groupModalCards, openGroupModal } = useContext(GroupModalContext);
-  const { primary, secondary, tertiary, quaternary } = useContext(SortContext);
+  const { primary, secondary, tertiary, quaternary, showOther } = useContext(SortContext);
   const openCardModal = useContext(CardModalContext);
   const {
     showCustomImages,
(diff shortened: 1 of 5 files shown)


src/components/CubeListNavbar.js
```diff
@@ -440,13 +440,21 @@ const CubeListNavbar = ({
                   </>
                 )}
                 <DropdownItem href={`/cube/clone/${cubeID}`}>Clone Cube</DropdownItem>
-                <DropdownItem href={`/cube/download/plaintext/${cubeID}`}>Card Names (.txt)</DropdownItem>
+                <DropdownItem href={`/cube/download/plaintext/${cubeID}?${sortUrlSegment}${filterUrlSegment}`}>
```
lunakv, line 443: This is minor, but since we're now using it everywhere, it might be neater to define the whole parameter string into a variable so that the same three segments don't have to be repeated.

lunakv (approved):

PR #2082, opened 2021-06-21
Added "add new cube" option to your cubes dropdown

### LINKED ISSUE #1903: Add "Create A New Cube" to "Your Cubes" dropdown

It would be nice to add an option to create a cube to the "Your Cubes" dropdown, in addition to (or instead of?) the user profile dropdown. 

Creating cubes feels like it belongs alongside my existing cubes.

Not a major issue since creating an additional cube is pretty rare.

dekkerglen:
Fixes #1903 

image
image

### DIFF
--- a/src/layouts/MainLayout.js
+++ b/src/layouts/MainLayout.js
@@ -117,9 +117,7 @@ const MainLayout = ({ user, children, loginCallback }) => {
                               </DropdownItem>
                             ))}
                             <DropdownItem divider />
-                            <CreateCubeModalLink>
-                              <b>Create A New Cube</b>
-                            </CreateCubeModalLink>
+                            <CreateCubeModalLink>Create A New Cube</CreateCubeModalLink>
                           </>
                         </DropdownMenu>
                       </UncontrolledDropdown>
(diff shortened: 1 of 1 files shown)


lunakv:
Visually, I'm not a fan of the bold text. It stands out too much. The rule above the link is imo enough to separate it.

src/layouts/MainLayout.js
```diff
@@ -110,11 +110,17 @@ const MainLayout = ({ user, children, loginCallback }) => {
                           Your Cubes
                         </DropdownToggle>
                         <DropdownMenu right>
-                          {user.cubes.map((item) => (
-                            <DropdownItem key={`dropdown_cube_${item.name}`} href={`/cube/overview/${item._id}`}>
-                              {item.name}
-                            </DropdownItem>
-                          ))}
+                          <>
```
lunakv, line 113: Is there a reason this fragment needs to be here?

dekkerglen: It's need, a root element can only contain one jsx tag

dekkerglen: NVM: was looking at the wrong one, this can be removed

PR #2085, opened 2021-06-21
Added infinite scroll for dashboard feed

dekkerglen:
Previously, we used to have five pages of blogposts as the feed, and at some point I removed pagination. This PR adds back the ability to go further back, but this time with infinite scroll. There is a 2 second timeout before retrieving the next 5 blogposts, and there is also a safeguard to ensure no posts are doubled up, from a new posts being inserted during the scrolling. There is a loading wheel at the end of the feed that is displayed before more items are loaded.

### DIFF
--- a/src/components/Feed.js
+++ b/src/components/Feed.js
@@ -1,0 +1,71 @@
+import React, { useState } from 'react';
+
+import InfiniteScroll from 'react-infinite-scroll-component';
+
+import BlogPost from 'components/BlogPost';
+import { csrfFetch } from 'utils/CSRF';
+
+import { Spinner } from 'reactstrap';
+import BlogPostPropType from 'proptypes/BlogPostPropType';
+
+const wait = async (ms) => {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+};
+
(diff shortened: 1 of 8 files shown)


src/components/Feed.js
```diff
@@ -0,0 +1,71 @@
...
+    <InfiniteScroll dataLength={feedItems.length} next={fetchMoreData} hasMore loader={loader}>
+      {feedItems.map((item) => (
+        <BlogPost key={item._id} post={item} />
+      ))}
+    </InfiniteScroll>
+  );
+};
+
+Feed.propTypes = {
+  items: BlogPostPropType.isRequired,
```
lunakv, line 68: This should be an array.

src/pages/DashboardPage.js
```diff
@@ -147,7 +142,7 @@ const DashboardPage = ({ posts, cubes, decks, user, loginCallback, content, feat
 };
 
 DashboardPage.propTypes = {
-  posts: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+  posts: BlogPostPropType.isRequired,
```
lunakv, line 145: This should be an array.

PR #2116, opened 2021-06-25
Add QR code modal on cube overview

dekkerglen:
Adds a link to cube overview that opens up a modal with a QR code to your cube.

image
image

image
image

imageimage

imageimage

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2304,6 +2304,16 @@
         "lodash": "^4.17.11"
       }
     },
+    "@react-native-community/art": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@react-native-community/art/-/art-1.2.0.tgz",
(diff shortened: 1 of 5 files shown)


lunakv:
Few notes:
- The modal definitely needs some padding at the bottom. Right now it looks like it was cut off just below the code. Plus, given that basically all our modals have a footer with at least a `Close` button, it's visually inconsistent.
- If we're including a footer in the modal (as I think we should), then it might be a nice addition to add a `Download` button that would save the code as an image.
- The overview link seems to me to be a bit out of place. I feel like it's in too prominent of a spot for what it does and what it's surrounded by. I would probably put it below the image, grouped with all the other links.
- I'm not the fan of the all-white modal body in dark mode. Visually it would make more sense to me to just have a white border around the code itself and leave the rest of the modal to its usual color (see below for an example).
qr

PR #2117, opened 2021-06-25
Customize sort and filter usage in export

lunakv:
Since we now support sorting and filtering for all export types, this PR adds the option to disable either of those for your export. It also slightly refactors the `Tooltip` component to be more configurable.
sortfilter

### DIFF
--- a/src/components/Tooltip.js
+++ b/src/components/Tooltip.js
@@ -3,12 +3,15 @@ import PropTypes from 'prop-types';
 
 import { UncontrolledTooltip } from 'reactstrap';
 
-const Tooltip = ({ text, children, ...props }) => {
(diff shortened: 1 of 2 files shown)


src/components/Tooltip.js
```diff
@@ -3,12 +3,15 @@ import PropTypes from 'prop-types';
 
 import { UncontrolledTooltip } from 'reactstrap';
 
-const Tooltip = ({ text, children, ...props }) => {
+const Tooltip = ({ text, children, wrapperTag, tooltipProps, ...props }) => {
```
lunakv, line 6: No current usage of `Tooltip` sets any props besides `text`, so this is a non-breaking change.

PR #2132, opened 2021-06-29
QR code update

lunakv:
This PR implements changes to the QR code modal that I mentioned in #2116. In particular,
- the modal has a footer with a Close button and a Download button that lets you save the svg,
- the link was moved from the top to below the picture alongside the other links,
- modal background in dark mode was adjusted,
- the share icon was removed from the link, as it now clashed visually with the surrounding elements.

#### Desktop
qr-modal
qr-modal-light
qr-overview
#### Mobile
qr-modal-m
qr-over-m

### DIFF
--- a/src/components/QRCodeModal.js
+++ b/src/components/QRCodeModal.js
@@ -5,18 +5,31 @@ import { Button, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
 
 import QRCode from 'react-qr-code';
 
+const saveAsPng = (cubeName) => {
+  const svg = document.getElementById('qr-code');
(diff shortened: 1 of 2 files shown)


joshfinnie (approved):
this looks good to me. I am slightly concerned about how we handle long user names and both RSS and QRCode on the same line. especially if we want to start adding more ways to share the cube on these lines.

But overall, i enjoy these changes.

lunakv:
I agree that once we start adding more links to share the cube, we should revisit how they're displayed, but I think we can cross that bridge when we get to it.

dekkerglen (approved):

PR #2134, opened 2021-06-30
Ensure ads are only loaded once per page

dekkerglen:
Issue: Some component updates would cause the `useEffect` hook of the nitropay ads to re-initialize the ads. This is unintended, and can lead to laggy and buggy behavior.

Fix: Add a flag that ensure the ad component only gets initialized once

### DIFF
--- a/src/components/Advertisment.js
+++ b/src/components/Advertisment.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useContext } from 'react';
+import React, { useEffect, useContext, useState } from 'react';
 import PropTypes from 'prop-types';
 
 import AdsContext from 'contexts/AdsContext';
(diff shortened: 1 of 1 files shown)


src/components/Advertisment.js
```diff
@@ -109,7 +111,26 @@ const Advertisment = ({
         });
       }
     }
-  });
+  }, [
```
lunakv, line 114: I believe the canonical way to make sure an effect is only executed once is to give it an empty array of dependencies.

lunakv: Which would then make the state variable unnecessary

dekkerglen: I tested this, but the ads still re-initialized

lunakv: That's really weird. I'll have to do some testing tomorrow

lunakv: I don't know how you tested this, but I just used this method and in both Firefox and Chromium, the effect was run only once per page - I was able to write an article and go through a whole draft without the effect code rerunning again. Are you sure you weren't just seeing multiple ads being loaded once?

I submitted a cleaned-up version of my solution in #2135, you can try it out for yourself.

PR #2135, opened 2021-07-01
Created useMount hook

lunakv:
Replaces #2134. 

This PR creates a hook that serves as an equivalent to `componentDidMount` for purely functional components. It also applies the hook to the `Advertisment` component, making sure that the Nitro ad-loading code is only executed once per page.


dekkerglen:
I tested and this does work

dekkerglen (approved):

PR #2147, opened 2021-07-17
Fix +2 mace breaking filters

dekkerglen:
The problem: using filters when +2 mace is in the available pool causes the parser to crash.

The cause: when replacing the name in the oracle text with a placeholder, we treat the name as a regular expression, where `+` has special meaning

The solution: Treat any `+`'s in card names as string literals

### DIFF
--- a/src/filtering/FuncOperations.js
+++ b/src/filtering/FuncOperations.js
@@ -67,7 +67,7 @@ export const nameStringOperation = (op, value) => {
   const strOp = stringOperation(op, value);
   return (fieldValue, card) => {
     let expandedValue = fieldValue
-      .replace(new RegExp(card.details.name, 'g'), NAME_PLACEHOLDER)
+      .replace(new RegExp(card.details.name.replace('+', '\\+'), 'g'), NAME_PLACEHOLDER)
       .replace(NAME_ALIAS, NAME_PLACEHOLDER);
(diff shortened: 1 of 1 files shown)


ruler501 (changes requested):
This should be a global replacement in the string for future proofing.

dekkerglen:
No, there are many places the name is used not as a regex

ruler501:
You misunderstood. You should replace all instances of + not just the first.

src/filtering/FuncOperations.js
```diff
@@ -67,7 +67,7 @@ export const nameStringOperation = (op, value) => {
   const strOp = stringOperation(op, value);
   return (fieldValue, card) => {
     let expandedValue = fieldValue
-      .replace(new RegExp(card.details.name, 'g'), NAME_PLACEHOLDER)
+      .replace(new RegExp(card.details.name.replace('+', '\\+'), 'g'), NAME_PLACEHOLDER)
```
lunakv, line 70:
```suggestion
      .replace(new RegExp(card.details.name.replace(/+/g, '\\+'), 'g'), NAME_PLACEHOLDER)
```

lunakv: by the way, this all is the result of Node not having a proper `replaceAll` function until v15, which is just dumb.

PR #2149, opened 2021-07-26
Bulk Upload fix

### LINKED ISSUE #2145: +2 Mace causing issues
he new AFR card name "+2 Mace" is causing some problems in CubeCobra.  First, I notice that the "Paste" option for adding cards didn't like "2x +2 Mace", interpreting it as 4 "Mace" cards.

Next, when trying to quick filter this list (https://cubecobra.com/cube/list/afr-artisan-doubleton) on the word "token", I get the error message:

Invalid regular expression: /+2 Mace/: Nothing to repeat

lunakv:
Fixes #2145 (along with #2147).

The current logic for parsing bulk upload TXT files is some wild stuff. This PR rewrites it to simplify things and fix some issues. For instance, instead of pushing objects into an array we're currently iterating through (which is a no-no) and then doing some weird shenanigans to determine card counts and set codes, the code now does only one pass through the input list, splits it with a single regex, then chooses the appropriate version. Just be glad no one tried to bulk upload "Borrowing 100,000 Arrows" before now.

Aside from bugfixes, the only functional changes in this PR are that set codes now ignore case, and that missing cards are displayed as the whole line, instead of weirdly split. I also changed `missing` to be an array instead of a string, since we were stringifying elements into it only to split them again on render.

### DIFF
--- a/routes/cube/helper.js
+++ b/routes/cube/helper.js
@@ -75,7 +75,7 @@ async function updateCubeAndBlog(req, res, cube, changelog, added, missing) {
 
 async function bulkUpload(req, res, list, cube) {
   const lines = list.match(/[^\r\n]+/g);
-  let missing = '';
+  let missing = [];
   const added = [];
   let changelog = '';
   if (lines) {
@@ -125,7 +125,7 @@ async function bulkUpload(req, res, list, cube) {
             changelog += addCardHtml(details);
           }
         } else {
-          missing += `${item}\n`;
+          missing.push(item);
         }
       }
     }
--- a/routes/cube/helper.js
+++ b/routes/cube/helper.js
@@ -74,58 +74,58 @@ async function updateCubeAndBlog(req, res, cube, changelog, added, missing) {
 }
 
 async function bulkUpload(req, res, list, cube) {
-  const cards = list.match(/[^\r\n]+/g);
+  const lines = list.match(/[^\r\n]+/g);
   let missing = '';
   const added = [];
   let changelog = '';
-  if (cards) {
-    if ((cards[0].match(/,/g) || []).length > 3) {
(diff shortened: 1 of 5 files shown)


dekkerglen (dismissed):

routes/cube/helper.js
```diff
@@ -74,58 +74,58 @@ async function updateCubeAndBlog(req, res, cube, changelog, added, missing) {
...
-            count = 1;
-          }
-          for (let j = 0; j < count; j += 1) {
-            cards.push(numericMatch[2]);
+        // separate counts and sets from the name
+        //                              |    count?   |name|     (set)?      |
+        const splitLine = item.match(/^(?:([0-9]+)x? )?(.*?)(?: \(([^)(]+)\))?$/);
+        const name = splitLine[2];
+        const set = splitLine[3];
+        let count = parseInt(splitLine[1], 10);
```
ruler501, line 100: There's toNullableInt you could use here so it's just `toNullableInt(splitLine[1]) ?? 1`.

ruler501 (approved):

src/pages/BulkUploadPage.js
```diff
@@ -60,7 +60,7 @@ const BulkUploadPageRaw = ({ cubeID, missing, blogpost, cube }) => {
           </p>
           <Row>
             <Col>
-              {missing.split('\n').map((card, index) => (
+              {missing.map((card, index) => (
```
ruler501, line 63: Update propTypes with the new type.

ruler501 (changes requested):

routes/cube/helper.js
lunakv: can't use `??` yet.

Issue #613, opened 2019-10-17
Make Private Cubes more Private


dekkerglen:
Private cubes should only be accessible to it's owner, when the owner is logged in. It should not be displayed on the profile, or even be accessible via direct link.

PR #2152, opened 2021-07-31
Convert document references to ObjectId values

lunakv:
## ⚠️ This migration requires downtime
(See also **Running the migration** below.)

This PR changes our models so that values containing IDs of other documents are now of type `ObjectId` instead of just strings. In addition of being more semantically correct, this is a prerequisite to the upcoming implementation of #613.

### Document casting
When you set a value of a document's field, Mongoose automatically checks the schema type of that field and tries to convert the value appropriately before saving. The same happens to parameters of a query before it's executed. For example, statements such as `Cube.find({ owner: "610435dd6ca69a7343799e98" })` and `cube.owner = "610435dd6ca69a7343799e98"` will work under the new model exactly as under the current one and search/save cubes with the corresponding `ObjectId`. This is also the reason `blog.cube = cube._id` and the like work under the current model.

### Equality checking
`ObjectId` values are objects. That means they can't be compared using `===`, because that would compare reference equality, and while that actually works in some scenarios, we cannot rely on it in general. Therefore we must use the `.equals()` function that's defined on `ObjectId` to compare whether two values are equal. This function accepts either an `ObjectId` or a hex string, meaning it can be used even against string representations of IDs. Similarly, equality-based methods like `Array.includes` can't be used with `ObjectIds` and have to be replaced by an equivalent construct.

### Understanding `_id` vs `id`
By default, each document in the database has an `_id` property of type `ObjectId`. Because that type has some disadvantages (see above), we might sometimes prefer working with its string representation instead. Calling `toString()` on the IDs every time can be cumbersome though, which is why Mongoose provides a virtual property `id`, containing exactly that stringified `_id`. Currently, `_id` and `id` are used pretty interchangeably in the codebase, especially on `req.user`. With the exception of checking for equality, this is fine due to the implicit casting mentioned above. While not strictly necessary because of that, this PR changes most of those usages to `_id` for consistency and to emphasize we're working with object IDs.

### Old values
As mentioned, Mongoose will helpfully convert data to appropriate types. Unfortunately, this doesn't happen when those documents are read. If an old cube with a string `owner` is loaded, that property will still be a string. Any solution that would compel Mongoose to rectify this does not work on `lean` objects, which presents a challenge. Part of this PR is also a database migration that we want to run because of this (among other reasons).

That still leaves us with an inconsistent state while that migration is running. ~~Luckily, again thanks to the aforementioned type casting, the only time when these inconsistencies come up is when working with the values in our javascript, and the only thing we do that isn't supported there is equality checking through `.equals()`. To bridge the gap when we use the new models but still have documents stored under the old ones, this PR polyfills an `.equals()` method on `String.prototype` that can check for equality against strings and `ObjectId`s alike. Once the one-shot migration is completed, this extension won't be necessary and can be removed.~~ 
*The solution described above still leads to inconsistent behavior while the migration is running. To see why, consider the query `Cube.find({ owner: user._id })`. If the schema is changed so that `owner` is of type `ObjectId`, that query won't be able to find cubes that still store the value as a `string`*.

### Running the migration
1. Initiate downtime
2. Merge this PR, switching to the new models
3. Run the migration from `migrate_foreign_keys` file
4. Once migration finishes, bring site back online 

Steps 1 and 4 aren't strictly necessary - the app will still function while the migration is running - but leaving the site online during the migration will lead to an inconsistent state that we should avoid (see above).

### Front-end
You may notice that front end components are completely unaffected by this PR. This is because when serialized, an `ObjectId` is converted into its hexstring and stored in the resulting JSON as such. Therefore the `reactProps` received by the client will just see these converted values as strings the same way they did before. (There is one random `PropTypes` fix that I came upon, but that's unrelated.)

### Other things
A lot of our models had stupid comments, such as `// Blog schema` written above the schema for podcast episodes. These are clearly the result of copying. Since those comments are stupid, I removed them.


dekkerglen (approved):

routes/cube/index.js
```diff
@@ -244,7 +244,7 @@ router.post(
 
     const { user } = req;
     cube.users_following = cube.users_following.filter((id) => !req.user._id.equals(id));
-    user.followed_cubes = user.followed_cubes.filter((id) => id !== req.params.id);
+    user.followed_cubes = user.followed_cubes.filter((id) => !req.user._id.equals(id));
```
ruler501, line 247: Shouldn't this be `cube._id`?

routes/users_routes.js
```diff
@@ -562,7 +562,7 @@ router.get('/decks/:userid/:page', async (req, res) => {
     return render(req, res, 'UserDecksPage', {
       owner: user,
       followers,
-      following: req.user && req.user.followed_users.includes(user.id),
+      following: req.user && req.user.followed_users.includes(user._id),
```
ruler501, line 565: should this be `some((id) => id.equals(user_id))`?

ruler501 (commented):
Should probably finish removing the X schema comments from all the files since even the ones that are correct are just restating the variable name on the next line.

ruler501 (approved):

models/comment.js
```diff
@@ -1,13 +1,12 @@
 const mongoose = require('mongoose');
 
-// Deck schema
 const commentSchema = mongoose.Schema({
-  parent: String,
+  parent: mongoose.Schema.Types.ObjectId,
```
ruler501, line 4: We store Scyrfall Card-IDs here (or maybe oracle Ids I didn't check closely) so we can't make this an ObjectId field. See CardPage and as you drill down to CommentsSection you'll see how parent gets defined.

ruler501 (changes requested):

routes/cube/deck.js
```diff
@@ -586,7 +586,7 @@ router.post('/submitdeck/:id', body('skipDeckbuilder').toBoolean(), async (req,
     // req.body contains a draft
     const draftid = req.body.body;
     const draft = await Draft.findById(draftid).lean();
-    const cube = await Cube.findOne(buildIdQuery(draft.cube));
+    const cube = await Cube.findOne(draft.cube);
```
MBTree, line 589: should this be changed to 'findById' like the other changes?

routes/cube/index.js
```diff
@@ -1351,7 +1351,7 @@ router.get('/draft/:id', async (req, res) => {
       return res.redirect('/404');
     }
 
-    const cube = await Cube.findOne(buildIdQuery(draft.cube)).lean();
+    const cube = await Cube.findOne(draft.cube).lean();
```
MBTree, line 1354: should this be 'findById' too?

routes/cube/deck.js
lunakv: They can be, but this is fine as well. The two are equivalent in these cases.

lunakv:
The PR (and its description) has been updated to account for the fact that the migration isn't actually easily possible without downtime.

Issue #1706, opened 2020-11-08
Expand compatibility with Scryfall's 'IS:' filter


dekkerglen:
We are currently missing:
  // Others from Scryfall:
  //   commander, reserved, reprint, new, old, hires, foil,
  //   spotlight, unique, bikeland, cycleland, bicycleland,
  //   bounceland, Karoo, canopyland, canland, checkland,
  //   dual, fastland, fetchland, filterland, gainland,
  //   painland, scryland, shadowland, shockland, storageland,
  //   creatureland, triland, tangoland, battleland, masterpiece,
  //   spell, permanent, historic, modal, vanilla, funny,
  //   booster, datestamped, prerelease, planeswalker_deck,
  //   league, buyabox, giftbox, intro_pack, gameday, release,
  //   foil, nonfoil, full, split, flip,
  //   leveler,

dillonplunkett:
Just want to flag that `firstprinting`, which can be quite useful, is missing from the list of unsupported `is:` traits, probably because it's missing even from some of Scryfall's documentation.

MBTree:
Hi! I'd like to contribute to this issue. Are there any specific considerations of the card object I should take into account? Or is there no issue with adding the new fields Scryfall has, like "reprint" and "story_spotlight", to the card object for this?

dekkerglen:
Adding more fields to card objects is no problem - I would like to encourage you to join our discord though, as that's where most of the dev discussion takes place. Just ping me and I'll grant you access to the hidden dev channels: https://discord.gg/tZs6gfq

PR #2156, opened 2021-08-06
Adding firstprint option to "is:" filter

MBTree:
Small update to #1706, giving a different filter option that scryfall supports.


lunakv (approved):

lunakv:
Worth noting Scryfall also supports `is:firstprinting` as an alias.

ruler501 (changes requested):
I'd like to add `is:firstprinting` as an alias, but otherwise it looks great.

PR #2171, opened 2021-09-13
Default seat count

### LINKED ISSUE #685: Ability to configure explicit seat count for custom draft formats
I design my cube for 6, 5 pack drafts. When I configure a custom draft format, I can configure the 5 packs fine, however the default "Total Seat" count when starting a draft is still 8. I'd like the ability to default the seat count for my custom draft format to 6.

lunakv:
Resolves #685.

This PR adds the option of specifying the default number of seats for custom draft formats.

seatcount

### DIFF
--- a/src/pages/CubePlaytestPage.js
+++ b/src/pages/CubePlaytestPage.js
@@ -191,7 +191,7 @@ const CustomDraftCard = ({
           )}
 
           <LabelRow htmlFor={`seats-${index}`} label="Total Seats">
-            <Input type="select" name="seats" id={`seats-${index}`} defaultValue="8">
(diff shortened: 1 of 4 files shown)


src/pages/CubePlaytestPage.js
```diff
@@ -242,6 +242,7 @@ CustomDraftCard.propTypes = {
     title: PropTypes.string.isRequired,
     html: PropTypes.string,
     markdown: PropTypes.string,
+    defaultSeats: PropTypes.number,
```
ruler501, line 245: This needs to be updated both places that draft formats are used in the prop types.

ruler501 (changes requested):

PR #2188, opened 2021-10-03
Featured queue timeout hotfix

lunakv:
There are reports of users getting timeout errors when attempting to add cubes to the featured queue. However, adding cubes from the admin panel works fine. 

To help isolate the issue and pinpoint a potential source of timeouts, this PR adds a limit on how many times a queue modification can be attempted.

### DIFF
--- a/serverjs/featuredQueue.js
+++ b/serverjs/featuredQueue.js
@@ -2,11 +2,13 @@ const FeaturedCubes = require('../models/featuredCubes');
 const Cube = require('../models/cube');
 const Patron = require('../models/patron');
 
+const TRY_COUNT = 5;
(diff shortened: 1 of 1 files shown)


dekkerglen:
I don't think a retry policy is the right approach, as the success rate has been 0%

lunakv:
Maybe, but before I throw it away altogether, I'd like to confirm what the actual issue is and this helps us do that. Keep in mind that the admin dashboard uses the same system and has a 100% success rate.

lunakv:
Just to clarify, this is not intended to be a fix. It is a debugging PR that should help root cause the issue in production, since I can't replicate it localy.

Issue #1738, opened 2020-11-25
Carry over url params on cube edit


dekkerglen:
When I go to Visual Spoiler View, then make changes from there and commit them, the page reloads in list view. Also, when I return to spoiler view, if I had before switched to Large card size, it will have reset to Medium.

Solution: Append the url query params to the submit endpoint - that way after reloading the page, it will go back to the view it was at.

brandonwharton:
I'm working on this one.

PR #2191, opened 2021-10-07
Carry over url params on cube edit #1738

brandonwharton:
Full PR for the view selection and spoiler card sizing permanence issue outlined [here](https://github.com/dekkerglen/CubeCobra/issues/1738). I opted for sending any existing view selection as URL query params from the EditCollapse view when adding or removing cards. I then altered each of the redirect routes in the /cube/edit/:id endpoint to default to displaying the Table view, unless another view had been selected beforehand. 

For the issue with keeping card size selection permanent if going between Visual Spoiler and other views, I moved the scale state from VisualSpoiler up a level to it's parent component in CubeListPage, then passed that state back into Visual Spoiler as props. 

This is my first time doing an official PR for an open source project as a newer developer, so any feedback is more than welcome!

### DIFF
--- a/routes/cube/index.js
+++ b/routes/cube/index.js
@@ -1419,10 +1419,14 @@ router.post('/edit/:id', ensureAuth, async (req, res) => {
     let cube = await Cube.findOne(buildIdQuery(req.params.id));
 
     const viewQuery = req.query.view || 'table';
+    const scaleQuery = req.query.scale || 'medium';
 
     if (!req.user._id.equals(cube.owner)) {
       req.flash('danger', 'Only cube owner may edit.');
-      return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?view=${encodeURIComponent(viewQuery)}`);
+      return res.redirect(`
+      /cube/list/${encodeURIComponent(req.params.id)}?view=${encodeURIComponent(viewQuery)}&scale=${encodeURIComponent(
+        scaleQuery,
+      )}`);
     }
 
     cube.date_updated = Date.now();
@@ -1489,7 +1493,11 @@ router.post('/edit/:id', ensureAuth, async (req, res) => {
         }
       } else {
         req.flash('danger', 'Bad request format, all operations must be add, remove, or replace.');
-        return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?view=${encodeURIComponent(viewQuery)}`);
+        return res.redirect(
+          `/cube/list/${encodeURIComponent(req.params.id)}?view=${encodeURIComponent(
+            viewQuery,
+          )}&scale=${encodeURIComponent(scaleQuery)}`,
+        );
       }
     }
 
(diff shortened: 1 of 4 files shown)


routes/cube/index.js
```diff
@@ -1418,9 +1418,14 @@ router.post('/edit/:id', ensureAuth, async (req, res) => {
   try {
     let cube = await Cube.findOne(buildIdQuery(req.params.id));
 
+    let viewQuery = 'table';
```
lunakv, line 1421: These statements can be condensed a bit. The if statement isn't actually necessary
```suggestion
    const viewQuery = req.query.view || 'table';
```

routes/cube/index.js
```diff
@@ -1418,9 +1418,14 @@ router.post('/edit/:id', ensureAuth, async (req, res) => {
...
+    if (req.query.view) {
+      viewQuery = req.query.view;
+    }
+
     if (!req.user._id.equals(cube.owner)) {
       req.flash('danger', 'Only cube owner may edit.');
-      return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}`);
+      return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?view=${viewQuery}`);
```
lunakv, line 1428: Just for sake of correctness, these should probably be encoded.
```suggestion
      return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?view=${encodeURIComponent(viewQuery)}`);
```

routes/cube/index.js
```diff
@@ -1487,7 +1492,7 @@ router.post('/edit/:id', ensureAuth, async (req, res) => {
         }
       } else {
         req.flash('danger', 'Bad request format, all operations must be add, remove, or replace.');
-        return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}`);
+        return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?view=${viewQuery}`);
```
lunakv, line 1495:
```suggestion
        return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?view=${encodeURIComponent(viewQuery)}`);
```

routes/cube/index.js
```diff
@@ -1527,7 +1532,7 @@ router.post('/edit/:id', ensureAuth, async (req, res) => {
     }
 
     req.flash('success', 'Cube Updated');
-    return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?updated=true`);
+    return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?updated=true&view=${viewQuery}`);
```
lunakv, line 1535:
```suggestion
    return res.redirect(`/cube/list/${encodeURIComponent(req.params.id)}?updated=true&view=${encodeURIComponent(viewQuery)}`);
```

src/components/EditCollapse.js
```diff
@@ -82,6 +82,8 @@ const EditCollapse = ({ ...props }) => {
   const removals = changes.filter((change) => change.remove || change.replace).length;
   const newTotal = cube.cards.length + additions - removals;
 
+  const viewParams = new URLSearchParams(window.location.search).get('view');
```
lunakv, line 85: The `CubeListNavbar` component, from which this component is called, has access to a `cubeView` property which contains this data already. I would suggest instead of querying for it here again, edit `CubeListNavbar` so that it passes that property into `EditCollapse` and then just use the passed prop.

src/components/EditCollapse.js
```diff
@@ -245,7 +247,7 @@ const EditCollapse = ({ ...props }) => {
         </Button>
       </Row>
       <Collapse isOpen={changes.length > 0} className="pt-1">
-        <CSRFForm method="POST" action={`/cube/edit/${cubeID}`} onSubmit={handleMentions}>
+        <CSRFForm method="POST" action={`/cube/edit/${cubeID}?view=${viewParams}`} onSubmit={handleMentions}>
```
lunakv, line 250:
```suggestion
        <CSRFForm method="POST" action={`/cube/edit/${cubeID}?view=${encodeURIComponent(viewParams)}`} onSubmit={handleMentions}>
```

dekkerglen (approved):

PR #2199, opened 2021-10-27
feat(#613) Private cubes

### LINKED ISSUE #613: Make Private Cubes more Private

lunakv:
Resolves #613.

This PR changes the mechanism for visibility state on cubes. With this PR, a cube can be in one of three states - "public", "unlisted", or "private". The "public" state remains unchanged and the "unlisted" state corresponds to the old "private" state. What's new is the addition of the "private" state, which represents the highest level of restriction on a cube's visibility.

## Features of private cubes
When a cube is set to private, it basically appears as nonexistent to anyone besides its owner. In particular, on a private cube, no one besides its owner or an admin can:
- see the cube overview, list, or any of the other pages related to it,
- start a draft of the cube,
- generate sample packs of the cube,
- view blog posts attached to that cube, both directly and on their feed.

Since "private" is a superset of "unlisted", it follows that a private cube also won't appear in search results, user profiles or data exports. Some other features specific to private cubes are:
- a private cube can't be added to the Featured Cubes Queue, nor can it be directly featured,
- conversely, a cube in the Featured Cubes Queue cannot be set to private,
- a visual indicator appears on the overview page to clarify for the owner that their cube is set to private.

#### New cube settings modal
Screenshot from 2021-10-27 10-24-56
#### Icon appearing for private cubes (with a tooltip on hover)
Screenshot from 2021-10-27 17-00-45

### Data model
To ensure backwards compatibility with the current document structure, the visibility state is represented through two indicator variables: `isListed` (already present) and `isPrivate`. The states "public", "unlisted" and "private" are then represented by assignments of `true, false`, `false, false`, and `false, true` respectively. This way, all cube models that don't have the new `isPrivate` field will be correctly represented as either "public" or "unlisted", preserving their current functionality.

### Information leak
Most routes are guarded to ensure that they behave the same way for a private cube as for a nonexistent cube, even ones that would otherwise already reject anyone besides the owner. However, since the private cube actually does exist, you will be unable to change the custom ID to the same one that it has. While this can lead to detection of short IDs of private cubes, and could theoretically cause some confusion ("the settings tell me that the ID already exists, but the rest of the site says it doesn't?"), the former issue is unavoidable due to the necessity of preventing collisions, and after a discussion with other devs, the latter wasn't deemed problematic enough to warrant a remedy.

### Feed
When a cube is set to private, we want all blog posts related to it to become private as well, which means they shouldn't appear in the feeds of any followers for as long as the cube stays private. To accomplish this, I chose to create an aggregation pipeline that allows us to filters blog posts based on that criterion (`serverjs/blogpostFeed.js`). I acknowledge such a construct may not be the most intuitive at first sight, but several other approaches were considered and all of them had some drawback or obstacle that made me discard it in favor of this solution. According to the AWS documentation, all of the stages and operators used here should be supported on DocumentDB as well.


dekkerglen (approved):

dekkerglen (dismissed):

dekkerglen:
Merge conflicts

lunakv:
resolved

PR #2208, opened 2021-11-26
Fix/2207 csv import maybeboard parsing

### LINKED ISSUE #2207: CSV Import - Maybeboard parsing blocks upload
https://github.com/dekkerglen/CubeCobra/blob/ecc33661427bec5753d2f5f3b7af086325f2e508/serverjs/cubefn.js#L286

If an upload CSV file doesn't specify the "Maybeboard" column or that column is not included in the row for a given card, it will be undefined when `.toLowerCase()` is called, causing the upload to fail.

smartintruesdell:
Fixes #2207

### DIFF
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -243,7 +243,7 @@ function CSVtoCards(csvString, carddb) {
     'collector number': collectorNumber,
     status,
     finish,
-    maybeboard = 'false',
+    maybeboard,
     'image url': imageUrl,
     'image back url': imageBackUrl,
     tags,
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -283,7 +283,7 @@ function CSVtoCards(csvString, carddb) {
         const nonPromo = potentialIds.find(carddb.reasonableId);
         const first = potentialIds[0];
         card.cardID = matchingSetAndNumber || matchingSet || nonPromo || first;
-        if (maybeboard.toLowerCase() === 'true') {
+        if (typeof maybeboard === 'string' && maybeboard.toLowerCase() === 'true') {
           newMaybe.push(card);
         } else {
           newCards.push(card);
--- a/serverjs/cubefn.js
+++ b/serverjs/cubefn.js
@@ -243,7 +243,7 @@ function CSVtoCards(csvString, carddb) {
     'collector number': collectorNumber,
     status,
     finish,
-    maybeboard,
+    maybeboard = 'false',
     'image url': imageUrl,
     'image back url': imageBackUrl,
     tags,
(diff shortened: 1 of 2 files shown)


serverjs/cubefn.js
```diff
@@ -243,7 +243,7 @@ function CSVtoCards(csvString, carddb) {
     'collector number': collectorNumber,
     status,
     finish,
-    maybeboard,
+    maybeboard = 'false',
```
dekkerglen, line 246: I believe this should be `maybeboard: 'false'`

smartintruesdell: This form `const { variableNames } of data` is the ["for of" variation of Destructuring Assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#for_of_iteration_and_destructuring).

Using a `:` colon here would be the syntax for [assigning new variable names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assigning_to_new_variable_names), ie renaming a property.

`const { foo: bar } = { foo: 'baz' } // -> foo === undefined, bar === 'baz'`

Using an assignment operator `=` provides a [default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_values_2).

`const { foo = 'baz' } = { foo: 'bar' }; // foo === 'bar'`
`const { foo = 'baz' } = { bar: 4 }; // foo === 'baz'`

On the very next line, there's a colon being used to assign a new variable name.

https://github.com/dekkerglen/CubeCobra/blob/8840550168801628ed5204cdd2612edda389370b/serverjs/cubefn.js#L237-L253

smartintruesdell:
Commit de4a5b2cec96b3b92f0d2a071b64b14730fcdd7a added a test to an existing test file, and the failing checks [here](https://github.com/dekkerglen/CubeCobra/runs/4461352745) seem unrelated to the change.

On investigation it looks like there's an incompatibility between canvas@2.6.1 and node@16.13.0.

https://github.com/Automattic/node-canvas/issues/1511

They recommend upgrading to canvas@^2.7.x

lunakv:
yeah, Cube Cobra itself is running on Node 12 (soon to be 14 hopefully), so that's more of an issue with our CI implicitly using 16 apparently.

serverjs/cubefn.js
lunakv: I don't really have a problem with this solution, but maybe it would be more straightforward to just chuck `if (maybeboard && ...` into the check instead to avoid the possible confusion of mixing these two together?

lunakv: Or, as I just realized, leave your proposed check as is. If we're doing the type checking, we don't actually need a default, and vice versa.

PR #2242, opened 2022-03-07
fix(cards) Regex type hotfix

lunakv:
`RegExp.exec` returns `null` when a match isn't made, not `undefined`. Some recent card actually got to the point of failing the regex, which caused the import to crash on `result[4]` (since `result` is null).

### DIFF
--- a/serverjs/updatecards.js
+++ b/serverjs/updatecards.js
@@ -298,7 +298,7 @@ function getTokens(card, catalogCard) {
           const re = new RegExp(reString);
           const result = re.exec(ability);
           // eslint-disable-next-line no-continue
-          if (typeof result === 'undefined') continue;
+          if (!result) continue;
 
           let tokenPowerAndToughness = result[4];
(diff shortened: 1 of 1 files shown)


lunakv:
Root-causing the issue:
- the regex used for parsing token text was handled incorrectly - it checked whether the result was undefined, but .exec() actually returns null, so this was in essence an empty check.
- this didn't matter because the regex is only parsed on abilities including the word "token", and all abilities with that word matched that regex successfully. 
- On 2022-02-21, the Crimson Vow minigames cards were added into the exports, one of which is [[Monster of the Week]]. This card includes the word "token", but doesn't match the regex, so the result fell through the check as null and the import failed when result[4] was referenced a couple of lines later.
- Interestingly enough, this was a regression. In the past, the check used to be result == undefined, and because of the way == does type coercion, it would return true for both null and undefined. However, a refactoring patch changed that line to typeof result === 'undefined' in order to comply with ESLint, which resulted in the current breakage.

PR #2317, opened 2022-08-19
feat(filtering) Add filter for combined power and toughness

lunakv:
Implements Scryfall's `pt:` card filter (also aliased as `wildpair:`), filtering cards based on the sum of their power and toughness.

Also cleans up a typo in the `power` and `toughness` filters (`parseFloat` doesn't take a second argument)


lunakv:
Any chance of this getting merged?

PR #2391, opened 2023-05-07
Add button to apply search filter

### LINKED ISSUE #2387: Suggestion: add debounce to the card filter
Now that the card filter automatically applies filters, I'm getting ~500ms render times every time the input changes. This is fine for the final render but it reduces the responsiveness of the input while I'm typing my query. I recommend using a [de-bounce](https://medium.com/@jamischarles/what-is-debouncing-2505c0648ff1) to alleviate this issue

Danijj89:
Fixes #2387 

## Context
The typeahead in the filter component triggers multiple fetches and re-renders for each key input. We want to introduce a button to trigger the search on-demand instead.

## Changelog 
- Updates the search filter to include an apply button. Removes the typeahead functionality.

### DIFF
--- a/src/components/FilterCollapse.js
+++ b/src/components/FilterCollapse.js
@@ -48,10 +48,16 @@ const numFields = [
 const colorFields = ['color', 'identity'];
 
 const FilterCollapse = ({ isOpen, hideDescription }) => {
+  const { filterInput, setFilterInput, filterValid, filterResult } = useContext(CubeContext);
+
   const [advancedOpen, setAdvancedOpen] = useState(false);
   const [values, setValues] = useState({});
+  const [searchFilterInput, setSearchFilterInput] = useState(filterInput);
 
-  const { filterInput, setFilterInput, filterValid, filterResult } = useContext(CubeContext);
+  const applySearchFilter = useCallback(
+    () => {
+      setFilterInput(searchFilterInput)
+    }, [setFilterInput]);
 
   const applyAdvanced = useCallback(async () => {
     // Advanced Filter change. Render to filter input.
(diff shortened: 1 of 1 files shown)


src/components/FilterCollapse.js
```diff
@@ -126,9 +136,12 @@ const FilterCollapse = ({ isOpen, hideDescription }) => {
               placeholder={'name:"Ambush Viper"'}
               valid={filterInput.length > 0 && filterValid}
               invalid={filterInput.length > 0 && !filterValid}
-              value={filterInput}
-              onChange={(event) => setFilterInput(event.target.value)}
+              value={searchFilterInput}
+              onChange={(event) => setSearchFilterInput(event.target.value)}
             />
+            <Button color="success" onClick={applySearchFilter}>
+              Apply Filter
```
lunakv, line 143: As a historical note, this button used to just say "Apply" before it was removed. I don't known whether this is better, but the advanced filter modal also uses just "Apply", so the shorter variant at least scores a point for consistency.

src/components/FilterCollapse.js
```diff
@@ -48,10 +48,16 @@ const numFields = [
...
 
 const FilterCollapse = ({ isOpen, hideDescription }) => {
+  const { filterInput, setFilterInput, filterValid, filterResult } = useContext(CubeContext);
+
   const [advancedOpen, setAdvancedOpen] = useState(false);
   const [values, setValues] = useState({});
+  const [searchFilterInput, setSearchFilterInput] = useState(filterInput);
 
-  const { filterInput, setFilterInput, filterValid, filterResult } = useContext(CubeContext);
+  const applySearchFilter = useCallback(
```
lunakv, line 57: I don't think this needs to be a callback. It's a single-line function, so the cost of creating it is negligible, and it's only passed to a single button, so we're not really saving too much on child rerenders either.

lunakv:
By the way, if you use Github's builtin issue-tracking syntax in your description (e.g. `Fixes #2387`), the relevant issue will be automatically closed when the PR is merged.

src/components/FilterCollapse.js
Danijj89: Agreed. That is also what I thought but looking around I noticed that all functions seemed to use it so I just followed the pattern. I will remove it.

Danijj89:
Updated with fixes

lunakv (approved):

PR #2424, opened 2023-11-29
Use date passed in from FeaturedQueue.put document for updating featured cube object.

Eetai:
Cubes were getting this date reset with every put call, which meant it happened when they were pulled off the featured list (desired) and when they became featured (undesired). Resetting that date meant that they were never going to be old enough to reach featured=false.

### DIFF
--- a/dynamo/models/featuredQueue.js
+++ b/dynamo/models/featuredQueue.js
@@ -36,7 +36,7 @@ module.exports = {
   put: async (document) => {
     await client.put({
       [FIELDS.CUBE]: document[FIELDS.CUBE],
-      [FIELDS.DATE]: new Date().valueOf(),
+      [FIELDS.DATE]: document[FIELDS.DATE],
       [FIELDS.OWNER]: document[FIELDS.OWNER],
       [FIELDS.FEATURED_ON]: document[FIELDS.FEATURED_ON],
       [FIELDS.STATUS]: STATUS.ACTIVE,


lunakv:
The change looks good, well done! A few minor notes:
1. Could you add to the description some short summary of what issues this was causing and how this fix addresses them? I feel like I get it, but it's not necessarily obvious just from the change and the PR name.
2. This change assumes that the date field is always set on all passed document. Is that true? I'm fairly sure it is, just wanted to make sure you actually verified that. If it's not true (or it's conceivable it might not be true), we could even keep the "now" as a default if the field was not set.

Eetai:
> The change looks good, well done! A few minor notes:
> 
> 1. Could you add to the description some short summary of what issues this was causing and how this fix addresses them? I feel like I get it, but it's not necessarily obvious just from the change and the PR name.
> 2. This change assumes that the date field is always set on all passed document. Is that true? I'm fairly sure it is, just wanted to make sure you actually verified that. If it's not true (or it's conceivable it might not be true), we could even keep the "now" as a default if the field was not set.

So at the moment featuredQueue.js is the only place FeaturedQueue.put gets called, and that code deals with existing objects that have dates. The adding-to-queue code doesn't exist at the moment (fq.updateFeatured gets called places in the code but doesn't exist) and when that's added back it'll have to be with a date passed in.

lunakv (approved):

PR #2430, opened 2023-12-27
Fix issue caused by concurrent calls to trybotpicks

Eetai:
We believe concurrent trybotpicks calls incrementing the pack number at the same time are the cause of drafts ending early. This guards against that behavior by ensuring we don't make a new trybotpicks call until the previous one has returned.

### DIFF
--- a/src/components/CubeDraft.js
+++ b/src/components/CubeDraft.js
@@ -93,6 +93,25 @@ const CubeDraft = ({ draft, socket }) => {
     }
   };
 
+  const delayedTryBotPicksLoop = async () => {
+    // the trybotpicks response will eventually flip this status to 'done'
+    let status = 'inProgress';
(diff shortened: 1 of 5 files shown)


src/components/CubeDraft.js
```diff
@@ -124,22 +124,33 @@ const CubeDraft = ({ draft, socket }) => {
...
-            if (status === 'in_progress') {
+        // the trybotpicks response will eventually flip this to 'done'
+        let status = 'inProgress';
+        // let's ensure we only make one trybotpicks call at a time
+        let keepRunning = true
+        while (keepRunning) {
+            keepRunning = false
+            if (status === 'inProgress') {
+              try {
+              setInterval(async () => {
```
dekkerglen, line 135: this looks like it's going to keep setting up loops and making the problem worse

PR #2434, opened 2024-01-03
fix(patreon): fix level check on frontend and server

DianaLease:
Fixes these two issues posted by vaasa on Discord:

> - The front-end checks for levels 1 or 2 to determine if you're able to add to the featured queue, but the actual list of levels is:
const LEVELS = ['Patron', 'Cobra Hatchling', 'Coiling Oracle', 'Lotus Cobra'];
Meaning the check should actually be for levels 2+ and right now Lotus Cobras probably aren't getting the option to feature their cube. 

> - The level check was completely removed from the backend and all active patrons are allowed to feature a cube (see the canFeatureQueue function and how it changed in commit 7254fc9). This is either an undocumented change that didn't make its way to the frontend, or a refactoring mistake ( @DEKKARU  will know which one).

### DIFF
--- a/src/pages/UserAccountPage.js
+++ b/src/pages/UserAccountPage.js
@@ -41,6 +41,8 @@ import withModal from 'components/WithModal';
 import CubePreview from 'components/CubePreview';
 import CubePropType from 'proptypes/CubePropType';
 
+const LEVELS = ['Patron', 'Cobra Hatchling', 'Coiling Oracle', 'Lotus Cobra'];
+
 const AddFeaturedModal = ({ isOpen, toggle, cubes }) => {
   return (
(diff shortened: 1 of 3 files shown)


src/pages/UserAccountPage.js
```diff
@@ -41,6 +41,8 @@ import withModal from 'components/WithModal';
 import CubePreview from 'components/CubePreview';
 import CubePropType from 'proptypes/CubePropType';
 
+const LEVELS = ['Patron', 'Cobra Hatchling', 'Coiling Oracle', 'Lotus Cobra'];
```
lunakv, line 44: Just as an ex-post minor note, I would have liked some comment here saying this mirrors the definition from the model (and vice versa), but it's not really a big deal

PR #2437, opened 2024-01-07
style(lint): fix prettier and eslint errors

DianaLease:
This PR attempts to fix (or ignore, where appropriate) all linting errrors. I also removed the `src/components/ClientOnly.js` file because it wasn't being used anywhere (and it contained some strange code like passing props to a React Fragment).

I'm opening this as a draft PR as I still want to do more smoke testing and there are still two additional types of errors that I need to address (react/jsx-no-constructed-context-value and react/forbid-prop-types). But I wanted to get this open so people can start reviewing it since it quite a large PR. Also, if you'd rather me break this up into a few smaller PRs, let me know!


one_shot_scripts/fix_followers.js
```diff
@@ -3,6 +3,8 @@ require('dotenv').config();
 
 const mongoose = require('mongoose');
 
+// the below import does not exist and should be updated if we ever wish to run this script again
```
lunakv, line 6: @dekkerglen Should this script just be deleted at this point?

one_shot_scripts/migrateDecks.js
```diff
@@ -6,9 +9,11 @@ const fs = require('fs');
 const mongoose = require('mongoose');
 const carddb = require('../serverjs/carddb');
 
+// the below imports do not exist and should be updated if we ever wish to run this script again
```
lunakv, line 12: Same question

routes/job_routes.js
```diff
@@ -44,27 +43,30 @@ router.post('/featuredcubes/rotate', async (req, res) => {
...
   }
   await Promise.all(notifications);
-  
+
   return res.status(200).send('featured cube rotation check finished successfully');
 });
 
 const tryUpdate = async (podcast) => {
   if (podcast.inactive) {
+    // eslint-disable-next-line no-console
```
lunakv, line 52: It might be worth it to just globally disable this rule for the server code

src/analytics/Chart.js
```diff
@@ -12,31 +12,27 @@ import CubePropType from 'proptypes/CubePropType';
...
-  };
-  const colors = [...Object.values(colorMap), '#000000'];
-
-  const getColor = useMemo(
-    () => (label, index) => {
-      return colorMap[label] ?? colors[index % colors.length];
-    },
-    [colorMap, colors],
-  );
+  const getColor = useMemo(() => {
```
lunakv, line 23: This use of `useMemo` seems pretty weird.

src/components/CubeDraft.js
```diff
@@ -98,26 +98,31 @@ const CubeDraft = ({ draft, socket }) => {
...
 
           console.log(json);
 
-          if (json.picks == 0) {
-            await new Promise((resolve) => setTimeout(resolve, 5000));
+          if (json.picks === 0) {
+            await new Promise((resolve) => {
+              setTimeout(resolve, 5000);
+            });
+            // eslint-disable-next-line no-continue
```
lunakv, line 119: This feels like another rule that is overly opinionated and could be globally disabled.

src/components/CubeDraft.js
```diff
@@ -98,26 +98,31 @@ const CubeDraft = ({ draft, socket }) => {
...
             continue;
```
lunakv, line 120: But also, why is this `continue` even here? It's literally the last statement in the loop body (other than a `catch` block), so it does nothing?

src/components/DeckCard.js
```diff
@@ -114,17 +116,18 @@ const DeckCard = ({ seat, deck, seatIndex, view }) => {
       </CardHeader>
       {view === 'picks' && (
         <CardBody>
-          {deck.type === 'd' ? (
-            <>
-              {deck.seats[0].pickorder ? (
+          {
+            // eslint-disable-next-line no-nested-ternary
```
lunakv, line 120: `no-nested-ternary` is another rule that I feel like should be disabled at least in JSX code, since you don't have other options there.

src/components/FoilOverlay.js
```diff
@@ -1,9 +1,8 @@
 /* eslint-disable react/prop-types */
```
lunakv, line 1: These are the linter errors that should be fixed rather than disabled :slightly_smiling_face:

src/components/Markdown.js
```diff
@@ -77,20 +77,18 @@ const renderMath = (node) => <Latex trusted={false} displayMode>{`$$ ${node.valu
...
+  </a>
+);
 
 const renderSymbol = ({ value }) => {
   const symbol = value.replace('/', '-').toLowerCase();
   return <img src={`/content/symbols/${symbol}.png`} alt={symbol} className="mana-symbol-sm" />;
 };
 
-const renderCardlink = ({ name, id, dfc }) => {
+function renderCardlink({ name, id, dfc }) {
```
lunakv, line 91: Why was this one changed to a function when all the ones around it stayed as const lambdas?

src/pages/CardPage.js
```diff
@@ -408,6 +406,7 @@ const CardPage = ({ card, history, versions, draftedWith, cubedWith, synergistic
             {filteredVersions.length > 0 ? (
               <PagedList
                 pageSize={10}
+                // eslint-disable-next-line react/no-unstable-nested-components
```
lunakv, line 409: The `pageWrap` value could be extracted into a separate component to fix the warning.

one_shot_scripts/fix_followers.js
dekkerglen: it was just a maintenance script so we don't need it anymore

src/components/CubeDraft.js
dvndrsn: You're right, this `continue` statement can be removed.

src/components/CubeDraft.js
dvndrsn: Generally, this rule encourages you to refactor long bodies in loops into smaller functions. A return from the extracted function can serve the same purpose as a continue inside the loop.

I think this [stackoverflow post](https://softwareengineering.stackexchange.com/questions/434124/are-there-any-problems-with-using-continue-or-break) explains the reasoning well. A continue isn't always bad, but its good to consider other tools in the toolbox.

Are there are lot of other continue statements elsewhere causing lint noise too?

routes/job_routes.js
dvndrsn: Could also extract a wrapper module `logger` for server logs. Then the eslint disable would only need to be in the one module on its different log level methods that wrap the console log. That would also leave a path to later graduate to using a more fully featured logging dependency.

This [blog post describes some good practices for logging with express and nodejs](https://blog.appsignal.com/2021/09/01/best-practices-for-logging-in-nodejs.html). Winston seems like a popular choice and supports structured logging which would be nice for log aggregation and search.

dvndrsn: Poking around the codebase a bit more, we already have a wrapper than implements this pattern for [cloudwatch](https://github.com/dvndrsn/CubeCobra/blob/fbed9e7fd648a06d902e570129d6a63db8b208a6/serverjs/cloudwatch.js#L103). The methods `cloudwatch.info` and `cloudwatch.error` can be used instead of `console.info`/`console.error` as console log is the fallback when cloudwatch isn't configured.

There's [an example](https://github.com/dvndrsn/CubeCobra/blob/fbed9e7fd648a06d902e570129d6a63db8b208a6/app.js#L142) of using this log wrapper in app.js. Maybe consider replacing console log with a similar pattern?

PR #2439, opened 2024-01-15
[WIP] Add collaborators

### LINKED ISSUE #1029: Collaboration Support
Allow multiple people in a "group" to have read/write powers on a cube list

Similar to Google Doc, you can give people "edit" permissions so they can modify your cube lists

**Additional context**
Would need to have enhanced audit logs to know who modified a cube, when, and what was changed

jskrzypek:
Add support for cube collaborators.

### Open Questions/Items that need resolution

_This list comes from @lunakv on Discord_
- [ ] How (if at all) do we display the cubes someone collaborates on on their page? I think it would be odd not to have any links to them at all and show only owned cubes.
  - [ ] also extends to giving it as an option when adding a card through the card page or adding a package. 
- [ ] How do we handle removing collaborators? Both visually, and on the model side.
- [ ] How do we handle removing users that collaborate on something? If there are backlinks in the user objects (perhaps due to the first point), then they're easy to remove, but if there aren't, do we just leave them?
- [ ] Do we add some visual indicator to the History tab / blog posts that says which collaborator made a change?
- [x] Do we allow collaborators to create blogposts? If not, do we allow them to create the posts that are associated with an update?
- [ ] Users should probably get a notification when they are added or removed as a collaborator.

Closes #1029

### DIFF
--- a/dynamo/models/cube.js
+++ b/dynamo/models/cube.js
@@ -173,7 +173,7 @@ const batchHydrate = async (cubes) => {
   const owners = await User.batchGet(cubes.map((cube) => cube.owner).filter((owner) => owner));
 
   const collaboratorIds = cubes.reduce((set, cube) => {
-    return new Set([...(set || []), ...(cube.collaborators | []).filter((id) => id)]);
+    return new Set([...(set || []), ...(cube.collaborators || []).filter((id) => id)]);
   });
 
   const collaborators = await User.batchGet([...collaboratorIds]);
--- a/dynamo/models/cube.js
+++ b/dynamo/models/cube.js
@@ -180,7 +180,7 @@ const batchHydrate = async (cubes) => {
 
   let collaboratorsById = new Map();
   for (let collaborator of collaborators) {
-    if (collaborator && collaborator.id) {
+    if (collaborator?.id) {
       collaboratorsById.set(collaborator.id, collaborator);
     }
   }
--- a/dynamo/models/cube.js
+++ b/dynamo/models/cube.js
@@ -26,6 +26,7 @@ const FIELDS = {
   ID: 'id',
   SHORT_ID: 'shortId',
   OWNER: 'owner',
+  COLLABORATORS: 'collaborators',
   NAME: 'name',
   VISIBILITY: 'visibility',
   PRICE_VISIBLITY: 'priceVisibility',
@@ -162,16 +163,31 @@ const hydrate = async (cube) => {
   }
 
   cube.owner = await User.getById(cube.owner);
+  cube.collaborators = await User.batchGet((cube.collaborators || []).filter((id) => id));
   cube.image = getImageData(cube.imageName);
-
+  
   return cube;
 };
 
 const batchHydrate = async (cubes) => {
   const owners = await User.batchGet(cubes.map((cube) => cube.owner).filter((owner) => owner));
 
+  const collaboratorIds = cubes.reduce((set, cube) => {
+    return new Set([...(set || []), ...(cube.collaborators | []).filter((id) => id)]);
+  });
+
+  const collaborators = await User.batchGet([...collaboratorIds]);
+
+  let collaboratorsById = new Map();
+  for (let collaborator of collaborators) {
+    if (collaborator && collaborator.id) {
+      collaboratorsById.set(collaborator.id, collaborator);
+    }
+  }
+
   return cubes.map((cube) => {
     cube.owner = owners.find((owner) => owner.id === cube.owner);
+    cube.collaborators = cube.collaborators.map((id) => collaboratorIds.get(id)).filter((c) => c);
     cube.image = getImageData(cube.imageName);
     return cube;
   });
@@ -322,6 +338,9 @@ module.exports = {
     if (document.owner.id) {
       document.owner = document.owner.id;
     }
+    if (document.collaborators) {
+      document.collaborators = [...new Set(document.collaborators.filter(c => c).map(({id}) => id))];
+    }
 
     delete document.image;
 
@@ -340,6 +359,10 @@ module.exports = {
 
     if (document.owner.id) {
       document.owner = document.owner.id;
+    }
+
+    if (document.collaborators) {
+      document.collaborators = [...new Set(document.collaborators.filter(c => c).map(({id}) => id))];
     }
 
     return client.put({
@@ -376,6 +399,7 @@ module.exports = {
     [FIELDS.ID]: `${cube._id}`,
     [FIELDS.SHORT_ID]: cube.shortID,
     [FIELDS.OWNER]: `${cube.owner}`,
+    [FIELDS.COLLABORATORS]: cube.collaborators.map((id) => `${id}`),
     [FIELDS.VISIBILITY]: getVisibility(cube.isListed, cube.isPrivate),
     [FIELDS.PRICE_VISIBLITY]: cube.privatePrices ? PRICE_VISIBLITY.PRIVATE : PRICE_VISIBLITY.PUBLIC,
     [FIELDS.FEATURED]: cube.isFeatured,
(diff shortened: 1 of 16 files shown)


src/contexts/CubeContext.js
```diff
@@ -54,6 +54,18 @@ const getDetails = async (cardId) => {
   return json.card;
 };
 
+const canUserEdit = (cube, user) => {
```
lunakv, line 57: There are a lot of editable things in the cube, and if we're going to restrict this to only editing the list, I think it should be reflected in the name to prevent confusion.

src/contexts/CubeContext.js
```diff
@@ -805,7 +817,7 @@ export const CubeContextProvider = ({ initialCube, cards, children, loadVersionD
     [cube, changes, setChanges],
   );
 
-  const canEdit = user && cube.owner && cube.owner.id === user.id;
+  const canEdit = canUserEdit(cube, user);
```
lunakv, line 820: This field also controls other things, such as setting tag colors and editing draft formats. Whether this value was used over an explicit equality check seems to have been somewhat arbitrary in the past. Have you gone through its usages and checked which options it affects? 

The naming note from above also applies here.

src/contexts/CubeContext.js
jskrzypek: Yeah, that makes sense. I'd noticed that and was thinking I should add more nuances here. Perhaps I should add a `userCubeRole` field to the context that's one of `"owner"` `"collaborator"` or `"viewer"` and then update places that use canEdit to key off of that instead. Some would require "owner", others "owner" or "collaborator".

jskrzypek: Also sorry I missed the bit in CONTRIBUTING.md about the `[WIP]` prefix on the PR. I've added that now

lunakv: That sounds like a solid approach if you want to spend the time on it. Otherwise, just using two booleans instead of one should also work fine.

dynamo/models/cube.js
```diff
@@ -162,16 +163,31 @@ const hydrate = async (cube) => {
   }
 
   cube.owner = await User.getById(cube.owner);
+  cube.collaborators = await User.batchGet((cube.collaborators || []).filter((id) => id));
   cube.image = getImageData(cube.imageName);
-
+
```
lunakv, line 168: Try to make sure you run `npm run prettier` before each commit (or set up autoformatting in your IDE) so that we don't add any new formatting errors. (There are plenty of them as is, but that's being worked on)

dynamo/models/cube.js
```diff
@@ -162,16 +163,31 @@ const hydrate = async (cube) => {
...
-
+  
   return cube;
 };
 
 const batchHydrate = async (cubes) => {
   const owners = await User.batchGet(cubes.map((cube) => cube.owner).filter((owner) => owner));
 
+  const collaboratorIds = cubes.reduce((set, cube) => {
+    return new Set([...(set || []), ...(cube.collaborators | []).filter((id) => id)]);
```
lunakv, line 176:
```suggestion
    return new Set([...(set || []), ...(cube.collaborators || []).filter((id) => id)]);
```

dynamo/models/cube.js
```diff
@@ -162,16 +163,31 @@ const hydrate = async (cube) => {
...
+  });
+
+  const collaborators = await User.batchGet([...collaboratorIds]);
+
+  let collaboratorsById = new Map();
+  for (let collaborator of collaborators) {
+    if (collaborator && collaborator.id) {
```
lunakv, line 183:
```suggestion
    if (collaborator?.id) {
```

routes/cube/index.js
```diff
@@ -171,6 +174,29 @@ router.get('/view/:id', (req, res) => {
   return res.redirect(`/cube/overview/${req.params.id}`);
 });
 
+router.post('/collaborator/:id', ensureAuth, async (req, res) => {
```
lunakv, line 177: Maybe this route should be plural - `/collaborators/` - to reflect the change in functionality? That's just a nitpick though.

serverjs/cubefn.js
```diff
@@ -341,6 +341,34 @@ function cachePromise(key, callback) {
...
+
+  return cube.owner && user.id === cube.owner.id;
+}
+
+function isCubeCollaborator(cube, user) {
+  if (!user || !cube || !cube.collaborators) {
+    return false;
+  }
+
+  return cube.collaborators.some((c) => c && c.id === user.id);
```
lunakv, line 369:
```suggestion
  return cube.collaborators.some((c) => c?.id === user.id);
```

serverjs/cubefn.js
```diff
@@ -341,6 +341,34 @@ function cachePromise(key, callback) {
...
+    return true;
+  }
+
+  if (!blog.cube) {
+    return false;
+  }
+
+  const cube = typeof blog.cube === "string" ? await Cube.getById(blog.cube) : blog.cube;
+
+  return cube.owner && user.id === cube.owner.id;
```
lunakv, line 361:
```suggestion
  return user.id === cube.owner?.id;
```

src/components/BlogPost.js
```diff
@@ -14,12 +14,14 @@ import CommentsSection from 'components/CommentsSection';
...
+import { userIsDocumentOwner, userIsDocumentCollaborator } from 'utils/Util';
+import CubePropType from 'proptypes/CubePropType';
 
-const BlogPost = ({ post, noScroll }) => {
+const BlogPost = ({ post, cube, noScroll }) => {
   const user = useContext(UserContext);
   const [editOpen, setEditOpen] = useState(false);
   const scrollStyle = noScroll ? {} : { overflow: 'auto', maxHeight: '50vh' };
-  const canEdit = user && user.id === post.owner.id;
+  const canEdit = user && (userIsDocumentOwner(user, post) || (cube && userIsDocumentOwner(cube, user)));
```
lunakv, line 24: The order of the arguments seems flipped between the two calls here.

dynamo/models/cube.js
jskrzypek: Didn't know this was supported in node now! nice :)

src/utils/Util.js
```diff
@@ -184,6 +184,23 @@ export function getCubeDescription(cube, changedCards) {
...
+  
+export function userIsDocumentCollaborator(user, cube) {
+  if (user && cube && cube.collaborators) {
+    return cube.collaborators.some((c) => c.id === user.id);
+  }
+
+  return false;
+};
+
+
```
jskrzypek, line 203: Changing the variable name to doc here
```suggestion
export function userIsDocumentOwner(user, doc) {
  if (user && doc?.owner) {
    return doc.owner.id === user.id;
  }
  
  return false;
};
  
export function userIsDocumentCollaborator(user, cube) {
  if (user && doc?.collaborators) {
    return doc.collaborators.some((c) => c.id === user.id);
  }

  return false;
};

```

src/components/BlogPost.js
jskrzypek: 💯 I also don't need to the short-circuiting checks for `cube` or `user` either 
```suggestion
  const canEdit = userIsDocumentOwner(user, post) || userIsDocumentOwner(user, cube);
```

dynamo/models/cube.js
jskrzypek: I will definitely do that before making this no longer a draft

dekkerglen:
appears to be outdated and abandoned, closing

jskrzypek:
Yeah, sorry I've never managed to find the time to go back to this. I might
make a concerted effort to try again in the future , given the various dev
improvements you've made, but I'd probably start from scratch regardless.

Issue #2421, opened 2023-11-01
Pressing the Enter Key in the filter field should apply filter


WesleyM77:
**What would you like to be able to do?**
When typing a filter into the filter field on a cube's list filter tab, pressing "Enter/Return" should apply the filter.

**Describe how you would use the feature**
Saves a bit of time when quickly trying to locate many cards by name (like when I replace proxies with real cards).

**Additional context**
None.

Qashworth:
Hi @dekkerglen! I have a local branch that I believe covers this issue but I'm having some trouble opening a pull request or pushing the branch. When time permits any guidance you can offer would be much appreciated!

WesleyM77:
@Qashworth I believe you push your branch to a forked version of the repo and open the PR on this repo pointing to a branch from your fork.

Qashworth:
@WesleyM77 Thank you so much! That did the trick, I was able to submit.

PR #2440, opened 2024-01-18
Enable Enter/Return key to apply current Filter [Issue-2421]

Qashworth:
### Description
This PR wraps the button in a form to handle `Enter` / 'Return' keypress events within the .component instead of requiring users specifically click the '_Apply_' Button discussed in #2421 

### Background
Currently users need to explicitly press the '_Apply_' button to submit filtered searches. This is an attempt to handle `Enter`/`Return` Key presses within the `Filtercollapse.js` component instead of requiring users specifically click the '_Apply_' Button. I've wrapped the `InputGroup` in a `Form`, removed the '_Apply_' button's `onClick` event, added the `"submit" type` to the '_Apply_' button, and added a function to handle when the `onSubmit` event is fired.

This should allow the 'Apply' button to maintain it's ability to be clicked as well as fire when the `Enter`/`Return` Key is pressed.

### How to test
- Check out this branch locally & start app
- Open a Cube list
- Click the 'Filter' option
- Type desired search criteria into the Filter input
- Click the `Enter` / `Return` key on your keyboard
- The Cube list should be filtered accordingly

### Additional Notes:
My first PR here (of hopefully many more) please be gentle! 😂

### DIFF
--- a/src/components/FilterCollapse.js
+++ b/src/components/FilterCollapse.js
@@ -122,26 +122,33 @@ const FilterCollapse = ({ isOpen, hideDescription }) => {
     [values],
   );
 
+  const handleOnSubmit = (e) => {
+    e.preventDefault();
+    applySearchFilter();
+  };
+
   return (
     <Collapse className="px-3" isOpen={isOpen}>
       <Row>
         <Col>
-          <InputGroup>
-            <InputGroupText htmlFor="filterInput">Filter</InputGroupText>
-            <Input
-              type="text"
-              id="filterInput"
-              name="filterInput"
-              placeholder={'name:"Ambush Viper"'}
-              valid={filterInput.length > 0 && filterValid}
-              invalid={filterInput.length > 0 && !filterValid}
-              value={searchFilterInput}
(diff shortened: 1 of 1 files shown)


lunakv (approved):
Looks good :+1: 

I noticed that this is handled slightly differently in other places such as  `pages/PackagesPage`, and `components/AutocompleteInput`, but I don't think that's much of an issue.

Oh, and by the way, if you put something like `Fixes #2421` into your PR description, it will [automatically](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) link to that issue and close it when the PR is merged.

Qashworth:
Thanks @lunakv, I'll be sure to format future PR's accordingly!

I did just notice that I failed to add the `Form` component to the other imported components for `FilterCollapse.js` in this PR, which breaks List view entirely: 
Screenshot 2024-01-24 223039
Screenshot 2024-01-24 223035

What would be the best way to go about including that component? Reverting this PR, opening a new one, some other third thing?

PR #2441, opened 2024-01-20
Capitalize "Maybeboard"

lunakv:
There are several instances where the word "maybeboard" is improperly lower-cased, such as in the header of the list view or in the "Show maybeboard" button text. This PR properly capitalizes the word in those places.


src/pages/CubeListPage.js
```diff
@@ -53,7 +53,7 @@ const CubeListPageRaw = () => {
                     list: <ListView cards={boardcards} />,
                   }[cubeView]
                 }
-                {boardname === 'maybeboard' && <hr />}
+                {boardname !== 'mainboard' && <hr />}
```
lunakv, line 56: This is purely a stylistic change to mirror the check from line 41 and anticipate the existence of other boards.

Issue #2442, opened 2024-01-20
Cube Comparison Ignores Tertiary and Ordered Sort


lunakv:
When comparing two cubes on the comparison page, only the primary and secondary sorts are respected. The tertiary sort and the ordered sort are ignored.

**To Reproduce**
1. Go to the cube comparison page of two cubes.
3. Open the Sort options
4. Change either tertiary or ordered sort.
5. The lists stay unchanged.

**Expected behavior**
The view should be reordered based on the updated options.

**Screenshots**
![image](https://github.com/dekkerglen/CubeCobra/assets/38463785/e74c27a2-9be7-465a-bccb-8789b77da531)

**Additional context**
The cause can be seen in the `CompareView.js` file. The `CompareGroup` component has `'Mana Value'` hard-coded as a sort option.

dekkerglen:
Fixed

PR #2443, opened 2024-01-20
Fix secondary sort error in cube comparisons

### LINKED ISSUE #2436: Error when using cube comparison secondary sort
**Describe the bug**
When using "secondary sort", I sometimes get the error shown in the screenshot below.

**To Reproduce**
1. Go to 'Compare Cubes'
2. Choose an option for 'secondary sort'
3. Keep trying different options until you see error (happened for me with "Tags" as secondary option)

**Expected behavior**
No error and instead see comparison data

**Screenshots**
Error:
Screenshot 2024-01-07 at 3 37 36 PM
Error when reproducing locally:
Screenshot 2024-01-07 at 3 40 30 PM

lunakv:
Fixes #2436.

### Issue
The error was caused by the `getLabels` function being incorrectly passed a whole column instead of an array of cards. 

For example, if the primary sort was `Color Category` and the secondary sort was `Guilds`, the `columns` object would end up looking like:
```
{
    "White": { 
        "Azorius": [...],
        "Boros": [...],
        ...
    },
    "Blue": { ...}
    ...
}
```

And then the `column` variable takes on the individual values of that, each of which is an object with guild names as keys (e.g. `{"Azorius": [...], "Boros": [...], ...}`). Because the `getLabels` function expects an array of cards and instead got this object, it caused errors, most notably because objects are not iterable and some sort options require iterating over the passed cards.

This also explains why only some tags were affected. Options such as `Mana Value` only return a predetermined set of labels, so only those options that actually need the list of cards (such as `Tags`) cause a crash.

### Fix
Change the call so that the column is flattened into a simple array of all the cards within it.

### DIFF
--- a/src/components/CompareView.js
+++ b/src/components/CompareView.js
@@ -174,7 +174,7 @@ const CompareView = ({ cards, both, onlyA, onlyB }) => {
                     </Col>
                   </Row>
                 </div>
-                {getLabels(column, sortSecondary, cube.showUnsorted)
+                {getLabels(Object.values(column).flat(), sortSecondary, cube.showUnsorted)
                   .filter((label) => column[label])
                   .map((label) => {
                     const group = column[label];


lunakv:
Oh, and I found another bug while debugging this, which I wrote up into #2442

Issue #2451, opened 2024-05-16
Make cube history publicly accessible through API


dsoskey:
**What would you like to be able to do?**
Make the data behind `/cube/history/:id` a publicly accessible API route

**Describe how you would use the feature**
I want to programmatically ingest the history API to generate snapshots of a cube's contents over time

**Additional context**
- Alternatively I could write a web scraper, but direct API access would be simpler and more reliable.
- I plan to author a PR for this issue

PR #2456, opened 2024-06-05
[2451]: added /cube/api/history/:id route for programmatic consumption

dsoskey:
[related issue](https://github.com/dekkerglen/CubeCobra/issues/2451)

This route is a specialized version of `/cube/getmorechangelogs` that is:
- publicly accessible (and therefore rate limited)
- has a higher page size to reduce the number of calls needed to iterate through a cube's history. Open question: is 50 an appropriate page size?

## Testing
Since i couldn't find any automated testing for this repo, I manually tested this on my dev setup using a cube with 2 changelog entries with the following paths:
- `/cube/history/api/testid`: successfully returns a 200 with 2 entires
- `/cube/history/api/testid` with a body containing lastKey: successfully returns a 200 with 2nd entry only

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -556,5 +556,5 @@ MigrationBackup/
 # End of https://www.gitignore.io/api/node,komodoedit,webstorm+all,visualstudio,visualstudiocode
 push-docker.ps1
 
-# asdf
+# asdf version manager config (https://asdf-vm.com)
 .tool-versions
--- a/.gitignore
+++ b/.gitignore
@@ -555,3 +555,6 @@ MigrationBackup/
 
 # End of https://www.gitignore.io/api/node,komodoedit,webstorm+all,visualstudio,visualstudiocode
 push-docker.ps1
+
+# asdf
+.tool-versions
--- a/routes/cube/api.js
+++ b/routes/cube/api.js
@@ -431,7 +431,7 @@ router.get(
   }),
 );
 
-router.get(
+router.post(
   '/history/:id',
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -555,3 +555,6 @@ MigrationBackup/
 
 # End of https://www.gitignore.io/api/node,komodoedit,webstorm+all,visualstudio,visualstudiocode
 push-docker.ps1
+
+# asdf
```
dekkerglen, line 559: ?

routes/cube/api.js
```diff
@@ -431,6 +431,39 @@ router.get(
   }),
 );
 
+router.get(
```
dekkerglen, line 434: Prefer post for API endpoints

dekkerglen (changes requested):

.gitignore
dsoskey: i use [asdf](https://asdf-vm.com/) to manage my node versions during local dev, so i ignored the config file it generates per repo. I can add a link to this comment to make it more clear

routes/cube/api.js
dsoskey: Im happy to switch to POST, but I'm curious why CubeCobra prefers POST to GET in this situation

dsoskey: Also with the switch to POST, would you prefer lastKey use the request body instead of a search param?

dekkerglen (approved):

PR #2471, opened 2024-10-15
stream export for cube and cards

dvndrsn:
Streaming cube and cards directly to file will reduce memory footprint for export scripts. It _may_ improve performance, but I didn't have a large dataset to try to run it against. I didn't fully understand the export decks script, so I didn't try to optimize that one yet.

**Future work**
For the cubes export, if it is possible to do a snapshot on the first run (export all results, as currently), then do a delta from the modified date on subsequent runs, it would be possible to significantly reduce the time/cost of running this process regularly. That approach would add some complexity to the exports consumer to reconstruct the current cube state by reading applying deltas of cubes changed to the snapshot of initial cubes exported.

If there is another bottleneck (ex. decks export, etc.) can do another pass of optimization.


dekkerglen:
PR is outdated - closing for now

Issue #1770, opened 2020-12-28
Adjust scrollbar for dark mode


lunakv:
In dark mode, the site uses the default horizontal scrollbar, which is usually white. That can be visually jarring. We should adjust the scrollbar color in dark mode so it matches the rest of the theme.

nicolinilr:
Hi I'd like to work on this issue, but I ran into a (possibly simple) roadblock: in my local cubecobra, I can't login (as it's not connected to the account database), but as a result I can't access my account change to get to darkmode and see my changes. Is there a development user/pass for the local copy?

ruler501:
You can register like normal through the web interface.

PR #2475, opened 2024-10-31
Fix for #1770 with Tailwind

KaelenProctor:
The Text component changes are unrelated to #1770 however as I was testing with the tailwindcss branch the errors prevented me from using pages. src/pages/UserAccountPage.js had the Text component added in the migration from headings to Text, so the rest looked like a miss with assuming Text was already imported.

Concern: public/css/stylesheet.css is being committed while also built via tailwindcss might result in confusion, whether to edit src/css/stylesheet.css or public/css/stylesheet.css

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "node": ">=20.0.0 <21.0.0"
   },
   "scripts": {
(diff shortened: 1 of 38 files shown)


dekkerglen:
Concern is valid, we should add that to gitignore

Issue #1181, opened 2020-03-09
Arena Download to Clipboard


dekkerglen:
Arena imports decklists through the clipboard, so instead of downloading a .txt, the export to arena should copy to clipboard with a client side script.

![image](https://user-images.githubusercontent.com/28238025/76249440-b47cbe80-6219-11ea-8938-4b3bdf1d3177.png)

PR #2478, opened 2024-11-03
Add Arena export (via copy to clipboard)

KaelenProctor:
Implements #1181. For the mainboard of the cube it generates the text to copy/paste into Arena, and has a button to copy direct to browser clipboard.

As I would only classify myself as having a working man's knowledge of React/Hooks, happy to adjust in case there are more React-y ways of doing things.

#### General

1) Use Sort/filter buttons apply to the Arena export when the modal opens. Tested several sort changes of the primary and secondary sorts, assuming the rest would work the same
2) Add and remove cards from the cube, the Arena export list is updated when modal opens
3) Multiple of the same card are exported and in Arena are counted together

#### Importing to Arena:
1) For each set on Arena (including extras like Jumpstart, Anthologies, and remasters) exported the card names via Scryfall. This got me approximately 170 cards per set. Tested over 40 sets.
2) Imported those names into a Cube
3) Exported to Arena via copy to clipboard
4) Imported to Arena

Issues encountered:
1) The backside of meld cards (ex. Mishra, Lost to Phyrexia) are not found on import to Arena. It does not have the backsides as "cards"
2) Cards ending with exclamation mark (ex. Fear, Fire, Foes!) cannot be imported (even using Arena export/import). Known bug open with Arena

### Screenshots

Export modal
export

Modal after clicking copy to clipboard button
export-after-copy

Example of filter applying when "Use filter" is enabled
filter-applying

Example of filter ignored when "Use filter" is off
filter-not-applied

Sorting applied
sorting

### DIFF
--- a/src/components/cube/CubeListNavbar.tsx
+++ b/src/components/cube/CubeListNavbar.tsx
@@ -4,6 +4,7 @@ import { Flexbox, NumCols } from 'components/base/Layout';
 import CompareCollapse from 'components/cube/CompareCollapse';
 import EditCollapse from 'components/EditCollapse';
 import FilterCollapse from 'components/FilterCollapse';
+import ArenaExportModal from 'components/modals/ArenaExportModal';
 import PasteBulkModal from 'components/modals/PasteBulkModal';
 import UploadBulkModal from 'components/modals/UploadBulkModal';
 import UploadBulkReplaceModal from 'components/modals/UploadBulkReplaceModal';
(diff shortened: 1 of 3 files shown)


dekkerglen:
This is going to be obsoleted by tailwind css and ts migration, will hold off until that is done and then we can rebase

KaelenProctor:
@dekkerglen I've rebased the work against the master branch after the tailwindcss changes were merged in. Converted to Typescript while doing so, have to say the types are pretty nice to work with.

PR #2498, opened 2024-12-22
Fix header linking in Markdown. Fix nested lists styling so they indent

KaelenProctor:
# Bugs

## Causes
The custom component renders for H1-6 that generate `<Text>` instead of literal `<h#>` causes the rehype-slug plugin for ReactMarkdown to do nothing. Then the rehype-autolink-headings plugin also does nothing as it relies on rehype-slug adding id attributes to the headings in order to generate the anchors (which in retrospect I am unsure what value that gave, as the autolink headings are normally prepended before the heading text per https://github.com/rehypejs/rehype-autolink-headings?tab=readme-ov-file#use). As such the markdown anchors people have used in their table of contents had nothing to link to within the page as no "header" elements in the rendered markdown contain id attributes.

Secondly for the list styling the padding is set to zero with the new stylesheet for `<ul>` and `<ol>`.

## Solutions
In order to retain the new styling of H1-6 with `<Text>` then we have to manually generate the ids like rehype-slug (using github-slugger) and add them to the `<Text>` components containing the header text.

For the ul/ol, without changing the overall element styling the simplest solution was to add one of the padding classes (ps-2) to the markdown rendered ul/ol in our components. That provides non-zero padding-inline-start, resulting in some indentation.

# Concerns
1. NPM and/or yarn installs from the master branch result in modified lock files. So the changes with rehype-autolink-headings,  rehype-slug, and github-slugger are not as isolated as I'd like.

# Testing

## Primary bug fixes
The bulk of testing was done with the Cube Overview markdown. I started from https://cubecobra.com/cube/overview/data 's overview as it was the reported context and has fairly complex markdown including headers and links to headers.

Gif of the header linking working locally after the changes:
header-linking

How nested lists are styled on current code:
unpadded-lists-in-markdown

Nested list styling with the added padding:
padded-lists-in-markdown

## Secondary testing

1. Validated that if the header contains Latex that the generation of the id skips. The nested context is more than just text so unable to see how an ID could be defined
2. Card name links in the Tokens page renders as markdown the same as before
3. Blog Post Markdown working
4. Markdown entry with preview working
5. Blog Post comments markdown working
6. Markdown page working
7. User about me markdown working

I was unable to test some other Markdown usages locally such as in the Draft decks or Articles/Video posts, but they should not be unique compared to the tested usages.


src/components/Markdown.tsx
```diff
@@ -163,73 +182,91 @@ const renderCardrow: React.FC<RenderCardrowProps> = (node) => (
 interface RenderH1Props {
   children: ReactNode;
 }
-const renderH1: React.FC<RenderH1Props> = (node) => (
-  <Text xxxxl semibold className="mb-4">
-    {node.children}
-  </Text>
-);
+const renderH1: React.FC<RenderH1Props> = (node) => {
```
KaelenProctor, line 185: Feels like there should be some way to have a single version of render header that takes only the size (eg xxxxl) as a parameter, but I couldn't figure out the currying

src/components/base/Text.tsx
```diff
@@ -60,11 +62,17 @@ const Text: React.FC<TextProps> = ({
...
 
+  let optionalProps = {};
+  if (id !== '') {
+    //@ts-ignore
+    optionalProps['id'] = id;
+  }
+
   if (area) {
-    return <p className={`${classes} h-full`}>{children}</p>;
+    return <p {...optionalProps} className={`${classes} h-full`}>{children}</p>;
```
KaelenProctor, line 72: Please let me know if there is a better way with React for optional empty props. With `id={id}` I saw empty ids on elements which I do not want.

src/pages/MarkdownPage.tsx
```diff
@@ -29,7 +30,7 @@ const MarkdownPage: React.FC<MarkdownProps> = ({ loginCallback }) => (
         <CardBody>
           <p>
             CubeCobra supports regular Markdown as well as some extra features specific to our site. If you need any
-            help regarding how to use markdown, please <a href="/contact">contact us</a>.
+            help regarding how to use markdown, please <Link href="/contact">contact us</Link>.
```
KaelenProctor, line 33: I noticed links on the markdown explanation page weren't styled as links anymore, so converted to Link component.
Before:
Capture

After:
Capture2

yarn.lock
```diff
@@ -2777,13 +2772,6 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react-google-recaptcha@^2.1.9":
```
KaelenProctor, line 2780: @types/react-google-recaptcha@^2.1.9 is also directly above this with the same integrity sha and resolved path, so it is not removed. Unsure how duplication could have gotten in

PR #2513, opened 2025-01-02
Playwright PoC

jamesriley72281:
It's just a test in case anyone wants to try this out. Some setup is needed, as it's a basic setup as this point.


jamesriley72281:
The more I've thought about this, the more I see a few hurdles we will need to address based on how we want to use this. I will type up a document explaining our options in the future, including how this would be used and what needs to be done to make each happen.

jamesriley72281:
Document discussed in my previous comment https://docs.google.com/document/d/1-RgsrhMOtrjPGSYMhoy-_TfHmjZjZ0ZX7P2gFtjve2Y/edit?usp=sharing

jamesriley72281:
Closing request

PR #2515, opened 2025-01-04
Fix cards not being trashed

KaelenProctor:
# Problem
Reported in Discord. Clicking cards when the draft step is "trashing" instead added them to the mainboard. This also occurred if the step was "auto trash".
Dragging the card correctly trashed it.

A corollary to the issue is that the draft step "Pack X Pick Y: <action>" title got out of sync and would be ahead of the current draft position (and sometimes illogical such as the pick being larger than than number of cards per pack).

The source of the problem comes from the 1.0.0 refactor and the conversion to the Drag and Drop Context surrounding the pack and deck stacks (boards). Before then (see https://github.com/dekkerglen/CubeCobra/blob/dd3fde3cbdfff410fe9a85df477091d957f35c4c/src/components/CubeDraft.js#L157) the click and drag (on move) were two different handles, with a click transforming into a move action.

After the refactor there are onDragStart and onDragEnd handlers, with the bulk of the logic in the latter. When the drag is less than 200ms it is considered a click, but the click handler was rebuilt to only put cards into the mainboard. The rest of the onDragEnd handler would consider the step and trash appropriately.

# Solution
Refactored the handler logic to move the business logic of the step action away from the drag or click actions.

During testing I enabled the sideboard DeckStack and validated that cards can be dragged or clicked from the mainboard to the sideboard and vice versa. However left the sideboard disabled as per Discord discussion nothing tracks the state of the sideboard so if the page is reloaded everything would go back to the mainboard.

## Concerns
I haven't been able to reproduce it consistently nor understand why, but I've seen cases where the auto-draft useEffect timeout triggers multiple times for the same draft step. In theory that shouldn't happen because of useEffect and specifically the pack dependency (let along the loading state), as that is an array containing the indices of the cards in the pack (the index being the position in the randomized set of cards). Thus after each active draft step, the pack array would be a unique array because its either a new pack or when the pack wheels, a subset of a previous pack array. **So likely its best to chalk this up to in progress development rather than an actual issue**

The CubeDeckbuilderPage has very similar logic for handling clicks and drag/drops, so we should probably refactor for consistency.

# Testing
For all tests created a custom draft with 3 packs, 5 cards per pack. The first pack is non-standard with trashing and auto-trashing alongside picks, see:
trashing-draft-setup

## Before
Clicking on cards during the trashing steps, and in the auto trash step actually adds the cards to the mainboard. Also note that the pack X, pick Y gets out of sync after the first trash step.
trashing-clicking-adds-to-board

However if cards are dragged from the pack to the mainboard the trashing works. The auto-trash step still fails. The pack X, pick Y gets out of sync after the auto-trash step.
trashing-dragging-works2

## After
Clicking on cards during the trashing steps, and in the auto trash step does not change the mainboard. The pack X, pick Y stays accurate.
trashing-fixed-clicking

Dragging works still, and the auto-trash is working.
trashing-fixed-dragging

### DIFF
--- a/src/components/draft/CubeDraft.tsx
+++ b/src/components/draft/CubeDraft.tsx
@@ -7,8 +7,9 @@ import DeckStacks from 'components/DeckStacks';
 import Pack from 'components/Pack';
 import AutocardContext from 'contexts/AutocardContext';
 import { CSRFContext } from 'contexts/CSRFContext';
+import CardInterface from 'datatypes/Card';
 import Draft from 'datatypes/Draft';
-import DraftLocation, { addCard, locations, removeCard } from 'drafting/DraftLocation';
+import DraftLocation, { addCard, location, locations, removeCard } from 'drafting/DraftLocation';
 import { draftStateToTitle, getCardCol, setupPicks } from 'drafting/draftutil';
 import useMount from 'hooks/UseMount';
 import { cardCmc, cardType, makeSubtitle } from 'utils/Card';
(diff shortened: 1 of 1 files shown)


src/components/draft/CubeDraft.tsx
```diff
@@ -201,7 +258,7 @@ const CubeDraft: React.FC<CubeDraftProps> = ({ draft, socket }) => {
         const dragTime = Date.now() - (dragStartTime ?? 0);
 
         if (dragTime < 200) {
-          return onClickCard(source.index);
+          return selectCardByIndex(source.index);
```
KaelenProctor, line 261: selectCardByIndex determines the card in the pack from its index, then calls applyCardSelectionForStep. Here we do that because the drag and drop handler has detected a click.
Below when the handler determines it is a drag, we also call applyCardSelectionForStep if the source of the drag was the pack and the target the deck or sideboard (the latter cannot occur given the sideboard deck stack is still commented out).

On any other drag we do moveCardBetweenDeckStacks which handles moving cards within a deck stack or across them (which again cannot occur with the sideboard commented out). Drags with the pack as the target do nothing since that is cleraly illogical (unless conspiracy draft rules could be implemented lol)

KaelenProctor:
Pushed additional changes:
1) Fixed issue with dragging cards between deck and sideboard, which isn't available since the sideboard is off. Original code swapped boards instead of cards =(
2) Fixed dragging from pack into a stack of cards in the deck always adding to the end. Dragging to the column adds the card to the end, and dragging on top of existing card adds before the existing card.

On the latter note I've noticed the drag/drop behaviour within a row/column to be different depending on whether the card dragged is before or after the card dragged to. If the card grabbed is after the card dragged to, the dragged card is added **before**. If the card grabbed is before the card dragged to, the dragged card is added **after**. I think this occurs because the dragged card is removed from the board first and then added using the original target index but that has shifted.

I dunno if its a big deal to worry about though.

KaelenProctor:
Couple additional changes.
1) Again related to sideboard, which isn't used. Added support for clicking to move cards between deck/sideboard

2) Improved the behaviour when you reload the draft page and picks are fetched from the backend. Before all the picks were added to row zero of the deck, eg the creature row. Now it organizes them by creature vs non-creature rows, the same as if you clicked cards while drafting. Any re-organization in the deck done via drag/drop is lost the same as before.

# Testing
Reloading behaviour before, all cards appear on row zero:
reload-draft-all-same-row

Reloading behaviour after, creatures are on the first row and non-creatures on the second.
reload-draft-organize-creature-non-creature

PR #2518, opened 2025-01-07
Mobile drag drop fix

KaelenProctor:
# Problem
In mobile trying to drag sortable or drag/droppable items ends up scrolling the screen instead of the dragging the item.

# Solution
In the DndContext, both drag/drop and sortable, they recommend `touch-action: none` https://docs.dndkit.com/api-documentation/sensors/pointer#touch-action on the draggable elements in order to disable the default browser behaviour. Since touch is only relevant to mobile, we don't need to make the CSS rule apply only for mobile browsers, it causes no issues with desktop ones.

A small other fix I found with the sortable (draft seats, tag colors) items is an error could be triggered if the item is dragged off the sortable area. This was already handled in the drag/drop (drafting, deckbuilding) by checking that the over property of the DragEvent isn't empty, so applied the same logic with comments.

## Concerns
1) In Firefox (desktop and mobile) clicking on the dropdown in the Tag color modal starts the drag and it continues even when you clicked an option

# Testing
Tested in Chrome and Firefox

## Before
Dragging gets stuck as the browser starts to scroll:
before-mobile-tag-color-modal-stuck
before-mobile-drafting-stuck
before-mobile-draft-seats-stuck
before-mobile-deck-build-drag-stuck

JS error when the sortable item is dropped off the area:
before-draft-seats-drag-off-error
before-tag-color-drag-off-error

## After
Dragging does not get stuck:
fixed-mobile-tag-color-modal
fixed-mobile-drafting-drag-drop
fixed-mobile-draft-seats-drag-drop
fixed-mobile-deck-build-drag-drop

No JS error when the sortable item is dropped off the area:
fixed-tag-color-drag-off-no-error
fixed-draft-seats-drag-off-no-error

### DIFF
--- a/public/css/stylesheet.css
+++ b/public/css/stylesheet.css
@@ -1210,6 +1210,10 @@ video {
           user-select: none;
 }
 
+.resize {
+  resize: both;
+}
+
 .list-inside {
   list-style-position: inside;
 }


package.json
```diff
@@ -18,7 +18,7 @@
     "test:watch": "npm run webpack --watch & npm run test-loud --watch --verbose false",
     "nearley": "sh nearley/helper.sh",
     "ci-build": "npm run nearley && tsc -b && NODE_OPTIONS=--max_old_space_size=16384 node_modules/.bin/webpack --mode production --config webpack.prod.mjs",
-    "devstart": "npm run nearley --watch & npm run nodemon & npm run webpack-dev-server & npm run tailwind:watch & npx tailwindcss -i ./src/css/stylesheet.css -o ./public/css/stylesheet.css --watch=always",
+    "devstart": "npm run nearley --watch & npm run nodemon & npm run webpack-dev-server & npm run tailwind:watch & npx tailwindcss -i ./src/client/css/stylesheet.css -o ./public/css/stylesheet.css --watch=always",
```
KaelenProctor, line 21: Fix for devstart after recent backend refactor and moved file

src/client/components/DeckStacks.tsx
```diff
@@ -1,11 +1,12 @@
+import React from 'react';
+
+import Card from '../datatypes/Card';
```
KaelenProctor, line 3: Lint fixes

src/client/components/modals/TagColorsModal.tsx
```diff
@@ -90,6 +93,11 @@ const TagColorsModal: React.FC<TagColorsModalProps> = ({ isOpen, setOpen }) => {
     (event: any) => {
       const { active, over } = event;
 
+      //If sort (drag and drop) ends without a collision, eg outside the sortable area, do nothing
+      if (!over) {
```
KaelenProctor, line 97: Fix for JS error when dragging sortable off the sort area

PR #2522, opened 2025-01-07
GitHub actions linting (eslint/prettier) only for changed files

KaelenProctor:
Improve the developer experience by:

1. Only reporting on lint issues within the changed files of the PR, not across the entire repo (given there are hundreds of minor issues like unsorted imports).
2. Turn off no-restricted-imports rule because in the recent backend refactor we are using relative import paths again.
3. Reduce duplication in github actions on pull requests, as both the request and push triggered the same stuff. Now push only triggers on the master branch.

Annoyingly the best github action for prettier changed files hasn't been updated in a while and was using an old version of Reviewdog (which does the actual commenting/checks for github) that didn't check the github token permissions correctly. Instead of forking and publishing to the marketplace, copied into the repo (its MIT licensed so fine to take and modify as long as the LICENSE/copyright lines are kept).

See https://github.com/KaelenProctor/CubeCobra/pull/3 for an example of how the eslint/prettier issues are reported on the PR as comments (which also generates emails). An alternative is to use a different reporter that puts all the details into the Checks tab.

eg:
image

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Run Prettier on changed files
         # Always run both eslint and prettier
         if: always()
-        uses: EPMatt/reviewdog-action-prettier@v1
+        uses: ./.github/actions/prettier
         with:
           fail_on_error: true
           filter_mode: file
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,6 +1,12 @@
 name: CI Tests
 
-on: [push, pull_request]
+# Pull request will trigger when PR is opened or updated, and more (https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request)
+# Thus limit push checks to master branch so things run after the PR is merged. Without the limit multiple actions run on each PR
+on:
+  pull_request:
+  push:
+    branches:
+      - 'master'
 
 jobs:
   build:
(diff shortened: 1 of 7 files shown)


.github/workflows/actions.yml
```diff
@@ -1,22 +1,44 @@
...
+      - 'master'
 
 jobs:
   build:
     runs-on: ubuntu-latest
     timeout-minutes: 30
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
```
KaelenProctor, line 17: Didn't end up needing the upgrade, but why not be on latest

.github/workflows/actions.yml
```diff
@@ -1,22 +1,44 @@
...
           node-version: 20
       - name: Enable Corepack
         run: corepack enable
       - name: Setup
         run: npm ci
-      - name: Check Code Style
-        run: npm run lint
+      - name: Run eslint on changed files
+        # Always run both eslint and prettier
+        if: always()
```
KaelenProctor, line 27: Always run both lint steps, as you might as well know all the issues in one go. In my experience its super annoying to think you fixed all issues and then another tool starts yelling at you.

package.json
```diff
@@ -9,7 +9,8 @@
   },
   "scripts": {
     "build": "npm run nearley && tsc -b && npm run webpack && npx tailwindcss -i ./src/client/css/stylesheet.css -o ./public/css/stylesheet.css",
-    "lint": "npm run prettier-base --check && npm run eslint -- --max-warnings 0",
+    "lint": "npx prettier --check $(npm run --silent list-files); npx eslint --max-warnings 0 $(npm run --silent list-files)",
```
KaelenProctor, line 12: lint was removed recently which broke the current master actions. I've added it back just for developers to use and updated the list-files paths

PR #2528, opened 2025-01-09
Missing Mana Symbols

AlexanderKYoung:
Added missing mana symbols and updated parser to handle new hybrid phyrexian mana.


KaelenProctor:
1. Is there a place where the symbols are downloaded from that we can document?
2. Does the sizing of the new symbols affect the rendering in the page?

AlexanderKYoung:
> 1. Is there a place where the symbols are downloaded from that we can document?
> 2. Does the sizing of the new symbols affect the rendering in the page?

1. I pulled all of the images from here: https://mtg.fandom.com/wiki/Category:Symbol_images

2. I tested the card page for each card that would be affected. All of the sizing is done on the front end. Here are screen shots of every newly added symbol that appear on cards:
Screenshot 2025-01-09 094750
Screenshot 2025-01-09 094335
Screenshot 2025-01-09 094411
Screenshot 2025-01-09 094453
Screenshot 2025-01-09 094508
Screenshot 2025-01-09 094529
Screenshot 2025-01-09 094555
Screenshot 2025-01-09 094615
Screenshot 2025-01-09 094633

PR #2533, opened 2025-01-10
Replace Sharp with Jimp for pure JS image manipulation library

KaelenProctor:
# Testing

From testing a jpeg is around twice as large as a webp image and takes 3 times as long to generate (1.7s vs 5.9s). PNG images were 5 times as large.

Creating sample image that is of type jpeg
sampleimage-jpeg

There is a Jimp plugin for webp https://github.com/jimp-dev/jimp/tree/main/plugins/wasm-webp but I couldn't get it to work with error `Error: No "exports" main defined in /home/node/app/node_modules/@jimp/wasm-webp/package.json`, which I assume relates to the note about "only works in esm environments". As well looking through the webp plugin dependencies it eventually uses https://github.com/webmproject/libwebp which has to be built, so might have the same type of build issue as Sharp.

### DIFF
--- a/src/routes/cube/index.js
+++ b/src/routes/cube/index.js
@@ -21,7 +21,7 @@ const {
   generateSamplepackImage,
   cachePromise,
   isCubeViewable,
-  isCubeListed
+  isCubeListed,
 } = require('../../util/cubefn');
 
 const { CARD_HEIGHT, CARD_WIDTH, addBasics, bulkUpload, createPool, shuffle, updateCubeAndBlog } = require('./helper');
@@ -43,7 +43,6 @@ router.use('/blog', require('./blog'));
 router.use('/deck', require('./deck'));
 router.use('/api', require('./api'));
 router.use('/download', require('./download'));
-
 
 router.get('/dumpdraft/:id', async (req, res) => {
   const dump = await dumpDraft(req.params.id);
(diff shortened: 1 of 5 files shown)


src/routes/cube/index.js
```diff
@@ -1027,7 +1027,7 @@ router.get('/samplepackimage/:id/:seed', async (req, res) => {
     });
 
     res.writeHead(200, {
-      'Content-Type': 'image/webp',
+      'Content-Type': 'image/jpeg',
```
KaelenProctor, line 1030: This is only functional change in the file. The rest were update by prettier/eslint

KaelenProctor:
Did some more testing with various jpeg quality settings:

| Quality | File size (MB) | Duration (s) |
| -------- | ------- | ----- |
| 95 | 3.2    | 4.9 |
| 90 | 2.2     | 5.75 |
| 75 | 1.8    | 6 |
| 50 | 0.9    | 4.6 | 
| 25 | 0.6    | 5.6 |

I would guess the durations aren't very accurate coming from the local dev environment.

An example of the 25% quality image is below which is surprisingly decent, as I expected at 25% lots of artifacts or dithering.
25-quality

In comparison the 90% quality is. In comparison the full size in two browser tabs, you can definitely see some artifacts is the lower quality such as around the mirrored lotus lines or the generous plunderer background. I don't think those would be noticeable though in the embed sizing though. 
90-quality

dekkerglen:
I think for now we will stick with sharp - the perf hit is a bit too much to make this change worth it

PR #2534, opened 2025-01-10
feat: push docker container for scheduled jobs

almarcotte:
Continues from https://github.com/dekkerglen/CubeCobraCDK/pull/2

This will allow us to publish a Docker image containing all the jobs we want to schedule. Before merging we need to merge and deploy the changes in the PR above to create the role and the ECR repository then include the values in here.

### DIFF
--- a/.github/workflows/image-push.yml
+++ b/.github/workflows/image-push.yml
@@ -1,0 +1,34 @@
+name: Build and Push Docker Image to ECR
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build-and-push-prod:
+    name: Build and Push to ECR
+    runs-on: ubuntu-latest
+
+    env:
+      role: arn:aws:iam::123456789012:role/replace-me
+      region: us-east-2
+      repository: "replace-me!"
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ env.role }}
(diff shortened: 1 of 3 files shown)


.github/workflows/image-push.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build-and-push-prod:
+    name: Build and Push to ECR
+    runs-on: ubuntu-latest
```
KaelenProctor, line 15: Could you update this workflow and the main .github/workflows/actions.yml to add an if condition to restrict to the "dekkerglen/CubeCobra" repo? I've been meaning to add that for a couple days as running tests when things are pushed to master branch isn't that interesting for forks, and definitely we don't want forks pushing images to ECR.

Re: https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution#example-only-run-job-for-specific-repository

almarcotte: Good catch! I've updated the workflows 👍 OIDC would prevent forks from pushing to ECR but no point in even running the job in those just for it to fail

dekkerglen:
Is this ready for review?

almarcotte:
> Is this ready for review?

@dekkerglen Yup, sorry 👍

dekkerglen:
Incorporated into https://github.com/dekkerglen/CubeCobra/pull/2812

PR #2537, opened 2025-01-11
Fix most recent card printing option being used

KaelenProctor:
# Problem
As reported in Discord the Default printing option of recent vs first wasn't working correctly.

# Solution
A couple different things are play:
1. Pass the setting through to the `cardimage/:id` backend call in the Autocomplete hover
2. Correct the passing of defaultPrinting through to the `cube/api/getcardforcube` API
3. Correct the sort for recent vs first. Sorting by Release date is from oldest to newest, not newest to oldest as the code/comment thought

As a bonus I also turned off the autocomplete hover images from the Tag autocomplete in the Advanced search modal. Those always showed the card back since a tag isn't a card.

# Testing
There are no differences in the following autocompletes because they use they pre-populate/search all card editions:
* Cube Overview image
* Profile image
* Card for a package
* Customize basics
* Edit article/video - Though I didn't test this as I haven't setup my user locally to be a content contributor

No changes to the remove card part of editing a cube, as that uses the exact card edition in the cube.

## Before

Always used the oldest printing regardless of what your cube's default printing setting was. Screenshot from when cube printing = most recent but its using Serra angel from Alpha
oldest-printing-showing-when-most-recent-is-active

Card back showing when using Tag autocomplete:
current-card-back-image-when-hovering-tags

## After

With Cube set to first printing:
first-printing2
first-printing-bulk

With Cube set to recent printing:
most-recent-printing
most-recent-printing-bulk

No card images when using Tag autocomplete:
no-card-back-image-when-hovering-tags

### DIFF
--- a/src/client/components/EditCollapse.tsx
+++ b/src/client/components/EditCollapse.tsx
@@ -237,6 +237,7 @@ const EditCollapse: React.FC<EditCollapseProps> = ({ isOpen }) => {
                 autoComplete="off"
                 data-lpignore
                 className="square-right"
+                defaultPrinting={cube.defaultPrinting}
               />
               <Button color="primary" disabled={addValue.length === 0} onClick={(e) => handleAdd(e, addValue)}>
                 Add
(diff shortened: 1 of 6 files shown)


dekkerglen:
Could you resolve the conflicts?

KaelenProctor:
Rebase and re-tested afterwards

PR #2538, opened 2025-01-12
Some minor Package screen fixes

KaelenProctor:
# Problems
1. If no packages were in the initial state, then switching the status filter did not display newly loaded ones. **This is extremely minor and only affected local development**, given that in production the default status filter is Approved and there are many approved
2. Your packages filter did not apply keyword sorting
3. Limit property name in backend incorrect casing

# Solutions
1. Minor fix to check the packages state length instead of the original items length
2. Added keyword filtering into the your packages getting
    1. Used dynamodb querying when the sort is Date because the ByOwner index is sorted by date
    2. For the sorting by votes, that has to load all the items into memory in order to sort. Added filtering as a first step
3. Fixed Limit typo. Given the default limit is 36 and that is what the code wanted, it is not a functional difference but is correct

One minor annoyance I noticed is when there are keyword filters, the paging with dynamo doesn't work 100%. Because dynamo applies filters AFTER the limit (https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html) you can get less than your Limit number of results. So if the page size is small compared to the total amount of results (in this case by the status) then you will have to click "Load more" multiple times to get to the end, and sometimes those loads would return nothing. Given the production page size of 36 I doubt its noticeable to users.

I also spent some time looking into converting dynamo/models/package.js into Typescript. I got a bit hung up on types for the dynamo queries and result types, but will continue to iterate.

# Testing

## Before
When nothing loads at first, switching the status filter should load stuff in but it doesn't. Can see in gif that if I reload the page, there are items for the selected filter
old-packages-dont-show-if-none-loaded-first

When filtering by your packages, there is no filtering by keywords
old-no-keyword-filtering-for-your-packages

## After
Now packages load in even if no packages were found on page load
new-packages-show-even-if-none-loaded-first

Keyword filtering working for your packages:
keyword-filtering-your-packages
keyword-filtering

Date/vote sorting still working for your packages
vote-sorting
new-date-sorting

Testing a limit of 1 for paging correctly pages:
fixed-limit


src/client/components/modals/CreatePackageModal.tsx
```diff
@@ -42,7 +42,7 @@ const CreatePackageModal: React.FC<CreatePackageModalProps> = ({ isOpen, setOpen
   };
 
   const save = async () => {
-    const response = await csrfFetch(`/packages/submit/`, {
+    const response = await csrfFetch(`/packages/submit`, {
```
KaelenProctor, line 45: Lint fixes and dropped dangling slash (imo ugly)

src/client/pages/PackagesPage.tsx
```diff
@@ -1,37 +1,36 @@
...
+import useQueryParam from 'hooks/useQueryParam';
+import MainLayout from 'layouts/MainLayout';
 
 const CreatePackageModalLink = withModal(Button, CreatePackageModal);
 
 interface PackagesPageProps {
   loginCallback?: string;
   items: CardPackageData[];
   lastKey: string | null;
-  activePage: 'approved' | 'submitted' | 'user';
```
KaelenProctor, line 31: activePage was only used in a dependencies but not the function, and once that was fixed up it was clear the variable isn't used anywhere

src/client/pages/PackagesPage.tsx
```diff
@@ -81,21 +80,19 @@ const PackagesPage: React.FC<PackagesPageProps> = ({ loginCallback = '/', items,
...
-  }, [getData, currentLastKey, filter, sort, ascending, packages]);
+  }, [getData, type, filter, sort, ascending, currentLastKey, packages]);
 
   const getNewData = useCallback(
     async (t: string | null, f: string | null, s: string | null, a: string | null) => {
       setLoading(true);
       setPackages([]);
       const result = await getData(t, f, s, a, null);
 
-      console.log(result);
```
KaelenProctor, line 92: Couple of unnecessary and unhelpful console logs were left in

src/client/pages/PackagesPage.tsx
```diff
@@ -218,7 +213,7 @@ const PackagesPage: React.FC<PackagesPageProps> = ({ loginCallback = '/', items,
             </CreatePackageModalLink>
           </Flexbox>
         )}
-        {items.length === 0 ? (
+        {packages.length === 0 ? (
```
KaelenProctor, line 216: This fixes issue 1, as only the original items array was checked to decide to render packages or not

src/dynamo/models/package.js
```diff
@@ -82,6 +82,32 @@ const batchHydrate = async (packs) => {
   return packs;
 };
 
+const applyKeywordFilter = (query /*: Query*/, keywords /*: string*/) => {
```
KaelenProctor, line 85: Consolidating the logic to reduce duplication

PR #2540, opened 2025-01-14
Some Typescript conversions

KaelenProctor:
# Testing
In general I tested by adding breakpoints to the modified functions and then using the site/scripts.

By file:
* s3client.ts - Loading cube in CubeCobra, creating a new cube, and deleting a cube
* documentClient.ts - Using the site, fetching things like Cube comments, Blog posts, etc. Also editing and deleting those
* client.ts - Spun up a new localstack container locally and then ran `npm run setup:local:db` and validated all the Dynamo tables * were created successfully (apart from session which is created by a separate express DynamoDBStore thing)
* utils.ts
    * get - Using the site
    * createTable - Via  `npm run setup:local:db` 
    * scan - Via `npm run exports` (minus the upload part) and validating the files in temp/exports
    * put - Using the site, creating blog posts and following users
    * query - Getting cubes using the site
    * delete - Delete blog post, delete comments, delete cube
    * batchGet - From cube exploration page, batch gets all the cube owners. In exporting decks it also tested the chunking that changed due to the removed caching
    * batchPut - Follow another user
    * batchDelete - Deleting a cube deletes the cube hashes, metadata, and changelog


package.json
```diff
@@ -34,7 +34,7 @@
     "update-all": "npm run update-draft-history && npm run update-cube-history && npm run update-metadata-dict && npm run update-cards",
     "update-hashes": "node --max-old-space-size=8192 src/jobs/repair_hashes.js",
     "upload-exports": "aws s3 sync temp/export s3://cubecobra",
-    "exports": "node --max-old-space-size=8192 src/jobs/export_cubes.js && node --max-old-space-size=8192 src/jobs/export_decks.js && node --max-old-space-size=8192 src/jobs/export_simple_card_dict.js && npm run upload-exports",
+    "exports": "NODE_OPTIONS=--max-old-space-size=8192 ts-node src/jobs/export_cubes.js && NODE_OPTIONS=--max-old-space-size=8192 ts-node src/jobs/export_decks.ts && NODE_OPTIONS=--max-old-space-size=8192 ts-node src/jobs/export_simple_card_dict.js && npm run upload-exports",
```
KaelenProctor, line 37: I found that with node things failed to load src/util/carddb which is now typescript. And export_decks.ts needs it directly.

src/dynamo/client.ts
```diff
@@ -1,15 +1,21 @@
...
   s3ForcePathStyle: !!process.env.AWS_ENDPOINT,
   accessKeyId: process.env.AWS_ACCESS_KEY_ID,
   secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
   region: process.env.AWS_REGION,
 });
 
-module.exports = new AWS.DynamoDB({ apiVersion: '2012-08-10' });
+const client = new AWS.DynamoDB({
+  apiVersion: '2012-08-10',
+  endpoint: process.env.AWS_ENDPOINT || undefined,
```
KaelenProctor, line 17: Typescript told me that endpoint isn't a valid config property for AWS.config, but it is for each service config (per the SDK)

src/dynamo/client.ts
```diff
@@ -1,15 +1,21 @@
...
+});
+
+export default client;
```
KaelenProctor, line 20: Without this default export other files didn't understand the type of the client.

src/dynamo/models/comment.js
```diff
@@ -35,6 +35,7 @@ const client = createClient({
     [FIELDS.ID]: 'S',
     [FIELDS.DATE]: 'N',
     [FIELDS.PARENT]: 'S',
+    [FIELDS.OWNER]: 'S',
```
KaelenProctor, line 38: When testing creating dynamodb tables the comment table failed with error that one of the secondary indices refers to an attribute that didn't exist. The ByOwner index was added within the last month. At least from my local dynamo contents the owner field exists though the ByOwner index doesn't.

We should check if the ByOwner index exists in production.

src/dynamo/s3client.ts
```diff
@@ -1,53 +1,39 @@
...
         Key: key,
       })
       .promise();
-    const value = JSON.parse(res.Body.toString());
-
-    // Update cache
-    await put(key, value);
 
-    return value;
+    return JSON.parse(res!.Body!.toString());
```
KaelenProctor, line 30: Typescript complains that res or body might be undefined. However if getObject didn't fail pretty sure those must be defined, so using the `!.` to make typescript be OK

src/dynamo/util.ts
```diff
@@ -1,18 +1,45 @@
...
+
+interface ClientConfig {
+  name: string;
+  partitionKey: string;
+  attributes: Record<string, string>;
+  indexes: Array<IndexConfig>;
+  sortKey?: string;
+}
+
+interface ClientInterface {
```
KaelenProctor, line 27: Typed up everything I could from Dynamodb in the SDK

src/dynamo/util.ts
```diff
@@ -100,133 +121,102 @@ module.exports = function createClient(config) {
...
             TableName: tableName(config.name),
             ...params,
           })
           .promise();
-      } catch (error) {
+      } catch (error: any) {
         throw new Error(`Error scanning table ${config.name}: ${error.message}`);
       }
     },
-    getByKey: async (key) => {
```
KaelenProctor, line 124: Found no usages

PR #2546, opened 2025-01-15
Mobile view improvements in the Cube pages

KaelenProctor:
# Problems
1. The Cube card count does not show on mobile
2. The Reset filters button does not show on mobile

# Testing

## Before
Card count not showing. Nor the cube categories:
card-count-not-on-mobile
In comparison to desktop, with a hugely long subtitle:
cube-card-count-all-categories-desktop

Reset button not showing in mobile:
old-mobile-filter-without-reset

## After

Card count showing. Not including the categories because those can be very long and thus hard to fit on mobile. The card count is pretty constrained on how long it will be.
mobile-card-count-shows

Reset button showing on mobile:
new-mobile-filter-with-reset

How it looks on various mobile sizes:

Iphone 12 (390px)
new-mobile-filter-with-reset-iphone12-390px
550px:
new-mobile-filter-with-rest-550px-widge

**I was thinking that it might make sense to force the buttons to a new flex row, which might help with very small devices such that the input is short enough the placeholder text isn't visible**

### DIFF
--- a/src/client/components/FilterCollapse.tsx
+++ b/src/client/components/FilterCollapse.tsx
@@ -94,13 +94,28 @@ const FilterCollapse: React.FC<FilterCollapseProps> = ({
               }
             }}
           />
-          <Button color="primary" onClick={() => setFilterInput(searchFilterInput || '')}>
-            <span className="px-4 whitespace-nowrap">{buttonLabel}</span>
-          </Button>
+          <ResponsiveDiv md>
+            <Button color="primary" onClick={() => setFilterInput(searchFilterInput || '')}>
+              <span className="px-4 whitespace-nowrap">{buttonLabel}</span>
+            </Button>
+          </ResponsiveDiv>
+          <ResponsiveDiv md baseVisible={true}>
+            <Button color="primary" onClick={() => setFilterInput(searchFilterInput || '')} className="h-full">
+              <span className="px-1 whitespace-nowrap">{buttonLabel}</span>
+            </Button>
+          </ResponsiveDiv>
           {showReset && (
             <ResponsiveDiv md>
               <Button color="danger" onClick={reset} className="h-full">
                 <span className="px-4 whitespace-nowrap">Reset</span>
+              </Button>
+            </ResponsiveDiv>
+          )}
+          {/* For mobile */}
+          {showReset && (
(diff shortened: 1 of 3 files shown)


src/client/components/FilterCollapse.tsx
```diff
@@ -94,16 +94,31 @@ const FilterCollapse: React.FC<FilterCollapseProps> = ({
               }
             }}
           />
-          <Button color="primary" onClick={() => setFilterInput(searchFilterInput || '')}>
-            <span className="px-4 whitespace-nowrap">{buttonLabel}</span>
-          </Button>
+          <ResponsiveDiv md>
```
KaelenProctor, line 97: Oddly grouping the buttons by <ResponsiveDiv> for desktop then mobile made the sizing all wrong, thus leaving as all Apply buttons then all Reset buttons

src/client/components/FilterCollapse.tsx
```diff
@@ -94,16 +94,31 @@ const FilterCollapse: React.FC<FilterCollapseProps> = ({
...
               <Button color="danger" onClick={reset} className="h-full">
                 <span className="px-4 whitespace-nowrap">Reset</span>
               </Button>
             </ResponsiveDiv>
           )}
+          {/* For mobile */}
+          {showReset && (
+            <ResponsiveDiv md baseVisible={true}>
+              <Button color="danger" onClick={reset} className="h-full">
+                <span className="px-1 whitespace-nowrap">Reset</span>
```
KaelenProctor, line 118: Reduced padding on mobile so the text input doesn't get crowded out.

src/client/components/cube/CubeSubtitle.tsx
```diff
@@ -19,6 +23,9 @@ const CubeSubtitle: React.FC = () => {
       <ResponsiveDiv md>
         <Text md>({subtitle})</Text>
       </ResponsiveDiv>
+      <ResponsiveDiv md baseVisible={true}>
+        <Text md>({mobileSubtitle} Cube)</Text>
```
KaelenProctor, line 27: During testing I added a title attribute with the full subtitle name, but unfortunately the mobile browser doesn't show the title popup unlike on desktop. Oh well.

PR #2547, opened 2025-01-15
build: setup groundwork for automated testing

almarcotte:
## Foundation for testing

Added 3 tests that can be used as starting points (comments more than welcomed on how I'm doing those):

- `src/client/components/card/CardImage.test.tsx` is for a React component
- `src/client/utils/Affiliate.test.ts` is for helper code
- `src/router/routes/api/draftmancer/publish.test.ts` is for backend API handler

I figured that gave a good idea of how to write tests going forward. That also gave me the opportunity to make sure that we can test everything we want.

I created a `test-utils` folder where I plan to keep test boilerplate code to keep test cleans. For example I added ways to create test objects like `Card`s or `CardDetail`s with sane default while still allowing to override specific fields as needed:

```js
const details = createCardDetails({ name: "Lightning Bolt" })
```

@dekkerglen Quick note that I made a tiny change to the publish from draftmancer API to make testing the API token a bit easier.

### Coverage

Coverage will be calculated and posted in the workflow console but because of permissions (see below) there's no simple way to get it posted as a comment yet. You can run `npm run test:coverage` locally to generate it in `coverage/`.

### Notes on the setup

`babel.config.js`: For some reason Jest doesn't play well with `babel.config.mjs`? I have no clue why (or what the difference even is) but after logs of Googling this worked. I think it might have to do with the fact that we're using typescript and JS at the same time?

I find it very annoying that you need to keep Jest's `moduleNameMapper` setting in sync with what is in `tsconfig.json` but as far as I can tell if it's not explicitly defined in Jest everything breaks apart.

The Jest documentation mentions [projects](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig) which we could use to split backend and frontend code if we ever want to run those separately but I figured it wasn't worth it for now.

## ESlint / Prettier

There are some issues with the GitHub Actions for those. For security reasons the actions do not have write access to PRs or the repository itself so they can't suggest changes or event write comments. I'm adding [Husky](https://typicode.github.io/husky/get-started.html) so that ESlint and Prettier are automatically run on **staged files** as a pre-commit hook.

## Caching

I added caching for `~/.npm` and `node_modules` to hopefully speed up the pipeline a bit. It will skip doing `npm install` if no changes to `package.json` are found.


package.json
```diff
@@ -45,7 +43,23 @@
...
+    "prepare": "husky"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{ts,tsx}": [
+      "prettier --write",
```
KaelenProctor, line 60: While here could you update the .prettierignore to include public/css/stylesheet.css? That file is generated by tailwindcss and so shouldn't need to have prettier run on it

tests/test-utils/data.ts
```diff
@@ -0,0 +1,111 @@
...
+const createCard = (overrides?: Partial<Card>): Card => ({
+  cardID: 'cardId',
+  ...overrides,
+});
+
+/**
+ * Create a CardDetails for testing by providing sane defaults but allow for overriding
+ * @param overrides
+ */
+const createCardDetails = (overrides?: Partial<CardDetails>): CardDetails => ({
```
KaelenProctor, line 19: Love this! I've found this design of having default data creation functions so useful for testing overall

package.json
almarcotte: Done!

tests/test-utils/data.ts
almarcotte: `Partial<Type>` is really cool 👍

PR #2548, opened 2025-01-16
First batch of dynamo model conversions to typescript

KaelenProctor:
# Notes
1. I found a bug where deleting comments isn't working as the code is trying to set owner to null, but dynamodb rejects that. I can reproduce the problem on production
2. I dropped non-general model functions that I found no usages of, such as changelog.getChangelogFromBlog or Blog.convertBlog
3. I switched FIELDS from constant array to Enums because the former does not work in Typescript to reference a property of a type like `document[FIELDS.ID]`

# Coding standard questions
1. What is the best way for us to setup types to handle unhydrated vs hydrated dynamo documents? Most datatypes are the hydrated variants, but methods like put often need to handle both versions (for example the code checking if owner.id exists, which means it is a User type vs owner being a string of their id)
2. What form of module exports to use? For some reason in update_card_history it could not use exported enums of FIELDS and TYPES, though it could use them as part of the exported object 🤷 
3. Do we want to expose dynamo types like AttributeMap or PutItemOutput, or should we wrap those into our types?

# Other questions
1. Is the ban_user.js script needed anymore with the admin section functionality for banning?

# Testing
In all testing I browsed the site while having active NodeJS debugging and validated the code going through the changed methods of the models.

## Blog
- getById - Click into blog post
- getUnhydrated - Editing a blog post
- getByCube - Cube blog list
- getByOwner - User blog list
- put - Create blog post, edit one
- delete - Delete blog post
- batchPut - Nothing uses this
- batchGet - Nothing uses this
- changelogToText - RSS feed of cube

## CardHistory
- batchPut - Ran update_card_history job
- put - Nothing uses this
- getByOracleAndType - In tools/card/:id used in Elo and PlayRate tabs

## Changelog
- batchGet / hydrateChangelog - View Cube blog
- put / sanitizeChangelog - Edit a cube to generate a changelog
- batchPut - Nothing uses this
- scan - Ran update_cube_history job
- getByCube / getChangelog  - Cube history page

## Comment
- hydrate / getById -  Edit comment, view individual comment
- queryByParentAndType / batchHydrate - Load comments on blog posts
- queryByOwner - Admin ban user
- put - Create/edit comment, reply to comment, delete, admin remove comment, admin ban user
- batchPut - Nothing uses this
- scan - Not used, used in admin ban user script


src/dynamo/models/blog.ts
```diff
@@ -0,0 +1,257 @@
...
+
+enum FIELDS {
+  ID = 'id',
+  BODY = 'body',
+  OWNER = 'owner',
+  DATE = 'date',
+  CUBE_ID = 'cube',
+  TITLE = 'title',
+  CHANGELIST_ID = 'changelist',
+}
```
dekkerglen, line 22: I think we can refactor this out, since this was kind of a stopgap anyways. Instead we can just use typescript features accomplish this

dekkerglen (changes requested):

src/dynamo/models/blog.ts
dekkerglen: I like `UnhydratedBlog`, and `Blog`, yeah. These enums are really just a stopgap and not are are just clutter now

src/dynamo/models/changelog.ts
```diff
@@ -228,82 +178,32 @@ const Changelog = {
...
-          process.env.DATA_BUCKET,
+          getBucketName(),
           `changelog/${document.cubeId}/${document.id}.json`,
           sanitizeChangelog(document.changelog),
         ),
       ),
     );
   },
-  createTable: async () => client.createTable(),
-  getChangelogFromBlog: (blog) => {
```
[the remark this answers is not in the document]
dekkerglen: Yeah it was part of the migration script actually - we can get rid of it now

src/dynamo/models/blog.ts
KaelenProctor: @dekkerglen I've pushed a new commit in which I added UnhydratedBlogPost and make BlogPost "extend" (as best possible in typescript) it, removing the FIELDS array fully.

One annoyance I found was the lack of way to align the types with the dynamo table configuration, property names like "cube" have to be duplicated as strings in the config. Which isn't a huge deal as unless working directly in the model you don't really need to know.

While making the updates I did notice that the body of a Blog post could be null, which normally is only when its an automated post that has a changelog. However I did see that the create blog post server function does not require body to be a string with length > 0, though it does check that for title. Should I add that in?

src/client/components/EditBlogModal.tsx
```diff
@@ -16,7 +16,7 @@ interface EditBlogModalProps {
 }
 
 const EditBlogModal: React.FC<EditBlogModalProps> = ({ isOpen, setOpen, post, cubeID }) => {
-  const [markdown, setMarkdown] = useState<string>(post ? post.body : '');
```
KaelenProctor, line 19: With the more correct typing that body can be null, this complained that a non-string could be passed to the state. Updated the form like title below, which handles both the post being undefined and the body having no value

src/datatypes/BlogPost.ts
```diff
@@ -2,13 +2,28 @@ import { Changes } from './Card';
...
-}
+  title?: string;
+  changelist?: string;
+};
+
+/* "extend" the UnhydratedBlogPost with replacements for properties, such as where the prop is always present or
+ * by Hydrating it is no longer a simple type
+ */
+type BlogPost = Commentable &
+  Omit<UnhydratedBlogPost, 'id' | 'owner' | 'body' | 'date' | 'changelist'> & {
```
KaelenProctor, line 20: Interface extension in typescript does not allow the type of a property of the parent changed. eg cannot go from id? to id. There is a lot more capability with types to compose things including Omit here, where we replace the properties that differ on the hydrated.

PR #2551, opened 2025-01-17
Fix loose list styling so long lines do not appear on a line after the positional marker.

KaelenProctor:
The original fix added the spacing between list items, but for long lines the start of the text got pushed down from the marker unexpectedly. Should have tested with more realistic text.

I didn't go hog wild on checking the alignment of the list marker vs text above in the old vs new changes. It looks reasonable to me in that there is some indentation for the list itself compared to lines above.

# Testing

## Before
old-long-loose-list-lines-breaking

## After
new-long-loose-list-lines-in-markdown

### DIFF
--- a/public/css/stylesheet.css
+++ b/public/css/stylesheet.css
@@ -1214,10 +1214,6 @@ video {
   resize: both;
 }
 
-.list-inside {
-  list-style-position: inside;
-}
-
 .list-decimal {
   list-style-type: decimal;
 }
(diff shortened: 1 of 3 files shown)


public/css/stylesheet.css
```diff
@@ -1214,10 +1214,6 @@ video {
   resize: both;
 }
 
-.list-inside {
```
KaelenProctor, line 1217: This style comes from tailwind. I believe since it doesn't see the style used in any classes it elides it from the compiled

PR #2556, opened 2025-01-19
Fix failures in drafting when cards have non-integer mana values (cmc).

KaelenProctor:
* Human's would get error picking it (in theory bots too if auto-picking)
* Draft could not finish if a bot had the card in their pool
* Deck build for me would fail if in your pool.

# Testing
1. Exported cube https://cubecobra.com/cube/playtest/old_border to my local using CSV. 
2. Started a draft
3. Just happened to pick Assault//Battery and got React error about unable to do board[row][col].push(..)
4. Debugged to discover the CMC was 2.5!
5. Same error occurred in the deckbuild "Build for me"
6. Did a new draft and did not choose Assault//Battery thus a bot must pick it since its a 360 cube
7. Draft completed

### DIFF
--- a/src/client/pages/CubeDraftPage.tsx
+++ b/src/client/pages/CubeDraftPage.tsx
@@ -10,7 +10,6 @@ import Pack from 'components/Pack';
 import RenderToRoot from 'components/RenderToRoot';
 import { CSRFContext } from 'contexts/CSRFContext';
 import { DisplayContextProvider } from 'contexts/DisplayContext';
-import CardType from 'datatypes/Card';
 import Cube from 'datatypes/Cube';
 import Draft, { DraftStep } from 'datatypes/Draft';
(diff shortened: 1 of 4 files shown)


src/client/pages/CubeDraftPage.tsx
```diff
@@ -10,34 +10,23 @@ import Pack from 'components/Pack';
...
-import { setupPicks } from '../../util/draftutil';
+import { getCardDefaultRowColumn, setupPicks } from '../../util/draftutil';
 
 interface CubeDraftPageProps {
   cube: Cube;
   draft: Draft;
   loginCallback?: string;
 }
 
-const getCardsDeckStackPosition = (card: CardType): { row: number; col: number } => {
```
KaelenProctor, line 31: Moved into draft utils and updated logic to use cmcColumn from getDefaultPosition (which is used in grid draft)

PR #2559, opened 2025-01-20
feat: move comments and blogs handlers to typescript

almarcotte:
Migrating the comments API to typescript. Also adding typing on the notification types while updating related code.

I found an issue where the link to a specific comment from the notifications page errors out if the comment has been deleted in the interim. I think this is related to hydration and the fact that the `owner` on a comment is no longer nullable

image

Are we in agreement that a user with ID '404' is considered "null"?

## Test utils

Added a `call()` function to help test API handlers:

```typescript
await call(myHandler).as(user).withRequest({ id: '12345'}).send();
```

Everything between `call()` and `send()` is optional and will get sane defaults if not provided. `withRequest` takes a `Partial<Request>` so you can override just what you care about for the test.


src/datatypes/Comment.ts
```diff
@@ -5,8 +5,33 @@ export default interface Comment {
...
+  'podcast',
+  'video',
+  'episode',
+  'package',
+] as const;
+
+export type CommentType = (typeof allCommentTypes)[number];
+export type NotifiableCommentType = Exclude<CommentType, 'card'>;
+
+const createTypeGuard = <T extends string>(validValues: readonly T[]) => {
```
KaelenProctor, line 29: Could this be moved into a utility class?

src/router/routes/comment.ts
```diff
@@ -0,0 +1,216 @@
+import { Request, Response } from 'src/types/express';
+
+import { isCommentType, isNotifiableCommentType, NotifiableCommentType } from 'datatypes/Comment';
```
KaelenProctor, line 3: For consistency should start all imports with relative paths these days

src/datatypes/Comment.ts
```diff
@@ -5,8 +5,33 @@ export default interface Comment {
   id: string;
   parent: string;
   type: string;
-  owner: User;
+  owner?: User;
```
KaelenProctor, line 8: A Comment may not have an actual user, but after hydration it should always have a User object as we will in with anonymous details if owner is empty in dynamo. Is there a case you found where it could be empty after coming out of comments model?

almarcotte: Ah yes that makes sense. The old code use to do `delete comment.owner` which typescript won't let you do unless `owner` is `User | undefined`. I noticed the `hydrate` sets it to an anonymous user so I made the change to do the same thing here.

KaelenProctor:
@almarcotte That is an interesting situation which makes sense to handle. Checking for 404 makes sense to me, as even if we updated the remove comment logic which sets the owner to 404 to set it as undefined (like the admin remove comment function does) there are likely comments in the prod DB with owner=404

almarcotte:
@dekkerglen Let me know if you have a preference on how to organize the handlers. I saw that your new code for setting up the routes supports exporting multiple handlers from the same file or have each handler in its own file prefixed by whatever folder they're in. I figured this was to allow breaking large APIs into smaller chunks when needed and not necessarily something to do every time.

Basically do you prefer in this case a single `comments.ts` vs something like:

```
comments/
    ├── add.ts
    ├── edit
    ├── getcomments.ts
    ├── report.ts
    └── get.ts
```

almarcotte:
Moved back to draft since I've added a couple of test utilities to reduce boilerplate in tests and I want to use it in here.

PR #2565, opened 2025-01-22
Token Tracker improvements

### LINKED ISSUE #1221: Missing tokens tracker
This is an issue to track all missing tokens for the token analysis. Comment below with any additional missing tokens. Currently known:

- [x] Tokens and emblems for the back side of all flip cards.
- [ ] Morph and manifest tokens
- [ ] Adorned Pouncer
- [x] Young Pyromancer
- [ ] Nahiri, the Lithomancer
- [ ] Sword of Body and Mind

MuttonBash:
Fixes #1221

I noticed a bug where if Scryfall lists both a "token" and other "special case" trackers in the card parts, the "special case" will get ignored by the token tracker.

For example, "Forth Eorlingas" has both "rebel" tokens and "Monarch" tracker, but the "Monarch" will get ignored because it will use the rebel token as the complete list and ignore the "special case" logic.

Also added the Day Night tracker to the special trackers list since it was missing.

Wanted to put together a quick PR since it looks like an easy fix; I haven't set up my local environment so not sure how to test it.

### DIFF
--- a/src/jobs/update_cards.ts
+++ b/src/jobs/update_cards.ts
@@ -241,10 +241,12 @@ async function writeFile(filepath: string, data: any) {
   });
 }
 
+const miscTokens = ['Manifest', 'A Mysterious Creature']
+
 function getScryfallTokensForCard(card: ScryfallCard) {
   const allParts = card.all_parts || [];
   return allParts
-    .filter((element) => element.component === 'token' || element.type_line.includes('Emblem') || element.type_line.includes('Dungeon') || (element.type_line.includes('Card') && !element.name.includes('Checklist')))
+    .filter((element) => element.component === 'token' || element.type_line.includes('Emblem') || element.type_line.includes('Dungeon') || (element.type_line.includes('Card') && !element.name.includes('Checklist')) || miscTokens.includes(element.name))
     .map(({ id }) => id);
 }
 
--- a/src/jobs/update_cards.ts
+++ b/src/jobs/update_cards.ts
@@ -253,7 +253,7 @@ const specialCaseTokens = {
 function getScryfallTokensForCard(card: ScryfallCard) {
   const allParts = card.all_parts || [];
   return allParts
-    .filter((element) => element.component === 'token' || element.type_line.startsWith('Emblem'))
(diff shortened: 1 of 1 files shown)


MuttonBash:
I wonder if we can just check that the type line is "Card" instead of hardcoding all the "special cases". Here are all the possibilities according to scryfall as of today:

['Foretell', 'On an Adventure', 'The Monarch', 'Poison Counter', 'Ixalan Checklist', 'Magic Origins Checklist', 'Energy Reserve', "City's Blessing", 'Plot', 'Innistrad Checklist', 'Shadows Over Innistrad Checklist 2', 'Experience', 'Eldritch Moon Checklist', 'Radiation', 'Ticket Bucket-Bot', 'Core Set 2019 Checklist', 'Shadows Over Innistrad Checklist 1', 'Dark Ascension Checklist', 'Rivals of Ixalan Checklist', 'Acorn Stash']

Maybe "Card" - "Checklist"

KaelenProctor:
I checked out your branch and tested it locally, looking great!

# Testing
1. Ran the update cards job
2. Created a cube and added a bunch of cards that have tokens/related cards

## Before
Cube cards:
cube-cards

Only tokens listed are literal tokens:
old-tokens

## After
Additional related token cards like dungeons and stuff also show:
new-tokens

Even larger amount of cube cards:
new-more-cards

Equals more tokens!
new-more-tokens

MuttonBash:
Thanks for helping me test it!

I think this change should address 99% of the missing token bugs (including those mentioned in #1221). There are a few minor edge cases / followups left:

1. Goblin Rabblemaster doesn't generate a token with haste. Both scryfall and the physical printings have the "wrong" token and the "right" token doesn't exist yet, so this doesn't seem fixable right now.

2. Scryfall doesn't include the tokens generated by dungeons (skeletons, goblins, etc), so those probably need to be hardcoded until scryfall fixes that. This can be a separate PR later.

3. I feel like the logic for parsing the scryfall tokens is complete enough with this change that the complicated fallback logic can be deleted.

For this PR though, it's ready for review.

PR #2566, opened 2025-01-22
Fix playtest page error when not logged in

KaelenProctor:
# Problem
Reported on Discord. The issue came because when not logged in the User in the UserContext is undefined, though it is typed as either User or null.

In most pages the difference between undefined and null doesn't matter as the check is non-strict like `bool = user` or `user != null` but in the playtest page was doing strict `user !== null`.

# Solution
1. Update the page to use a non-strict check. Followed other page examples and in this case did `!!user`
2. Updated the UserContext value from reactProps so that it defaults to null. Thus the type in UserContext is accurate now, and undefined shouldn't be possible

I check all the usages of UserContext in the React components to double check, which lead me to a couple other minor updates for null checks for consistency.

# Testing

## Before
Page error on playtest:
old-playtest-page-fails-not-logged-in

Can see via React Dev tools that the user provider has undefined instead of the expected null:
old-user-context-is-undefined-when-not-logged-in

## After
Playtest page loads when not logged in:
new-playtest-page-loads-not-logged-in

User context is null when not logged in:
new-user-context-is-null-when-not-logged-in

In comparison the user context is an object when logged in:
new-user-context-is-object-when-logged-in

### DIFF
--- a/src/client/pages/CubeBlogPage.tsx
+++ b/src/client/pages/CubeBlogPage.tsx
@@ -33,7 +33,7 @@ const CubeBlogPage: React.FC<CubeBlogPageProps> = ({ cube, lastKey, posts, login
 
   return (
     <MainLayout loginCallback={loginCallback}>
-      <CubeLayout cube={cube} activeLink="blog" hasControls={user != null && cube.owner.id == user.id}>
+      <CubeLayout cube={cube} activeLink="blog" hasControls={!!user && cube.owner.id === user.id}>
         <Controls>
           <Flexbox direction="row" justify="start" gap="4" alignItems="center" className="py-2 px-4">
             <CreateBlogModalLink color="primary" modalprops={{ cubeID: cube.id, post: null }}>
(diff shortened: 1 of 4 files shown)


src/client/pages/CubeBlogPage.tsx
```diff
@@ -33,7 +33,7 @@ const CubeBlogPage: React.FC<CubeBlogPageProps> = ({ cube, lastKey, posts, login
 
   return (
     <MainLayout loginCallback={loginCallback}>
-      <CubeLayout cube={cube} activeLink="blog" hasControls={user != null && cube.owner.id == user.id}>
```
KaelenProctor, line 36: This form `user != null` worked before and after, so its more of a style change for consistency across pages. In most user was checked like `user ? A : B` or `!!user { .. }`

src/client/pages/CubePlaytestPage.tsx
```diff
@@ -56,7 +56,7 @@ const CubePlaytestPage: React.FC<CubePlaytestPageProps> = ({ cube, decks, decksL
 
   return (
     <MainLayout loginCallback={loginCallback}>
-      <CubeLayout cube={cube} activeLink="playtest" hasControls={user !== null && cube.owner.id === user.id}>
```
KaelenProctor, line 59: I tested with this original line and the RenderToRoot change, and in that case there was no bug because the User context was accurately a User object or null, no undefined. Either way, using less strict check again mainly for consistency

PR #2567, opened 2025-01-22
Dynamo model conversion #2

KaelenProctor:
# Changes
Improvements to already converted models and conversion of the Patron model to Typescript.

1. All models using defined Types for get/put actions instead of DynamoDB generic attribute maps/item lists
2. Using `keyof <Type>` for attribute names in DynamoDB queries, which allows typescript to statically check the string desired is in fact a property name of the corresponding type

Misc other improvements;
1. Use DefaultElo constant all over
2. Fixes to a couple of the scripts
3. Consolidation to using datatypes to reduce duplication
4. Also check Patron status (active or inactive) in the UI to show the "Add cube to featured queue" button, given that the backend checks that too

# Testing
1. Re-ran through my previous testing steps for things like Comments and Changelog, using node debugging to ensure the changed code ran
2. Partial testing of the Patron changes.
    1. Manually set myself as a patron using DyanmoDB, and tested different levels/statuses
    2. That allowed me to test the Patron UI parts such as showing my patron status (Hatchling etc) and featured cube actions
    3. **Did not test the actual Patreon linking/unlinking webhooks**. Are those something that can be tested locally in that they don't affect the actual Patreon status?

Related to https://github.com/dekkerglen/CubeCobra/pull/2559


src/dynamo/models/changelog.ts
```diff
@@ -161,30 +157,12 @@ const Changelog = {
...
-      [FIELDS.CUBE_ID]: cube,
-      [FIELDS.DATE]: new Date().valueOf(),
-    });
+      id,
+      cube,
+      date: new Date().valueOf(),
+    } as ChangelogType);
     return id;
   },
-  batchPut: async (documents: DocumentClient.PutItemInputAttributeMap[]): Promise<void> => {
```
KaelenProctor, line 170: No usages so removed

src/dynamo/models/comment.ts
```diff
@@ -111,86 +93,90 @@ const batchHydrate = async (items: DocumentClient.ItemList | undefined): Promise
...
+      items: await batchHydrate(result.Items as UnhydratedComment[]),
       lastKey: result.LastEvaluatedKey,
     };
   },
   queryByOwner: async (
     owner: string,
     lastKey?: DocumentClient.Key,
-  ): Promise<{ items?: DocumentClient.ItemList; lastKey?: DocumentClient.Key }> => {
+  ): Promise<{ items?: Comment[]; lastKey?: DocumentClient.Key }> => {
+    const ownerAttr: keyof UnhydratedComment = 'owner';
```
KaelenProctor, line 128: As an example of the static checking, if the string here was "ownar" then typescript would complain it isn't a known key of the UnhydratedComment type

src/dynamo/util.ts
```diff
@@ -13,7 +13,7 @@ const tableName = (name: string): string => `${process.env.DYNAMO_PREFIX}_${name
 interface IndexConfig {
   name: string;
   partitionKey: string;
-  sortKey: string;
+  sortKey?: string;
```
KaelenProctor, line 16: I had misread the code earlier and now see that sortKey is not required for an index

src/jobs/update_cube_history.ts
```diff
@@ -65,6 +67,34 @@ const loadCubesHistory = async (key: string): Promise<CubeDict> => {
   return cubes;
 };
 
+const mapTotalsToCardHistory = (
```
KaelenProctor, line 70: Consolidation and using types instead of maps

src/jobs/update_draft_history.ts
```diff
@@ -106,6 +107,9 @@ const loadAndProcessCubeDraftAnalytics = (cube: string) => {
   if (!fs.existsSync('./temp/all_drafts')) {
     fs.mkdirSync('./temp/all_drafts');
   }
+  if (!fs.existsSync('./temp/drafts_by_day')) {
+    fs.mkdirSync('./temp/drafts_by_day');
```
KaelenProctor, line 111: Ran into missing directory error on my machine when testing, ensure it exists

PR #2573, opened 2025-01-24
build: initial setup for automated release

almarcotte:
## Goals

1. _Automate_ the build, release and deployment of the application
2. Automate version management
3. Make sure manual input is still required

## Workflow

On manual input, the workflow should do the following:

1. Release a new version (which includes creating a tag and updating `package.json`) automatically
2. Build and publish to S3 (using the existing `publish.js` script)
3. Trigger a (currently non-existent) workflow in `CubeCobraCDK` to release the dev stack (it might actually just be the prod stack with different values)

Then the workflow will pause and wait for manual input for the next step which is:

4. Trigger the same workflow as above but deploy the prod stack.

## Semantic versioning

I'm bringing semantic versioning. It's used for a few things but primarily to keep the `package.json`'s `version` field up-to-date. It'll also create the tags and generate a `CHANGELOG.md` file. A lot of this depends on [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) so if it's too involved then we can just drop it. I know that @dekkerglen publishes a blog on the website for bigger releases so the automated change log might be useful.

## What's missing

* The workflow in `CubeCobraCDK`. I have an opened MR that adds a workflow that does releases on-demand. I can update it so it allows being triggered by other workflows.
* I need to test this properly 😅 I'll probably copy the workflow on an empty repository and give it a shot.


.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,83 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+
+      - name: Install Semantic Release
+        run: npm install semantic-release @semantic-release/git @semantic-release/npm -g
```
KaelenProctor, line 21: Rather than installing globally, wouldn't doing an npm install of dev resources be correct? Then executing the package using npx so it finds it within the installed node modules.
Installing globally wouldn't guarantee the same version of those packages are used in comparison to the lock file

almarcotte: Good catch. I've made sure all the needed modules are in `package.json` and updated the job to run `npm ci` and `npx semantic-release` instead 👍

PR #2576, opened 2025-01-25
build: CDK for infrastructure management and deployment

almarcotte:
### Goals

1. Automate the build, release and deployment of the application
1. Automate version management
1. Make sure manual input is still required

### How it works

On manual input, the workflow should do the following:

1. Release a new version (which includes creating a tag and updating package.json) automatically
1. Build and publish to S3 (using the existing publish.js script)
1. Trigger next workflow to deploy the CDK stack for both environments

Manual input will be needed for the deployments to actually go through. This has to be enabled in the environment settings on the repository:

image

I'm bringing semantic versioning. It's used for a few things but primarily to keep the package.json's version field up-to-date. It'll also create the tags and generate a CHANGELOG.md file. A lot of this depends on [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) so if it's too involved then we can just drop it. I know that @dekkerglen publishes a blog on the website for bigger releases so the automated change log might be useful.

## Deployment

Adding a workflow to deploy CDK changes. It can be used manually to deploy or be called by another workflow (like the app release) to trigger a deployment of the infrastructure with CDK.

image

The values picked there are passed as context values to CDK to determine which environment to deploy into.

It requires a role which can be created by CDK so this will require one last manual run by @dekkerglen to get the role created. Then, it needs to be added as a secret on the repository (`CDK_IAM_ROLE`). There are also a couple more secrets to add, mostly the values that are taken from the environment variables.

## Scheduled Jobs

Creating a Fargate cluster and a reusable way for us to deploy scheduled jobs to it. For a given job it will create:

* An ECR repository
* A task definition
* A schedule rule

All we need to do afterwards is push an image to the ECR repository and Fargate will take care of the rest. There is a bit of a chicken-and-egg situation where we need the repository created to be able to push to it but the task definition needs a repository. What would happen now is this:

1. We create a job in the stack (by adding a new entry in the jobs config)
2. Deploy the stack
3. The repository will be created, along with the job definition

From this point on the job will run on schedule but fail immediately (can't pull the image because it doesn't exist). I think we're fine with this? It means very minimal costs because nothing runs. Once we push the image to the repository the next scheduled run will execute as expected.

## A bit of restructuring 

The goal was to reorganize the code a bit so that the main stack file didn't grow too wildly. I wanted to make sure that the generated CloudFormation template was the same so I generated both and diffed them:

```diff
80c80,82
<     "ApplicationName": "CubeCobra-development-app",
---
>     "ApplicationName": {
>      "Ref": "Application"
>     },
97c99,101
<     "ApplicationName": "CubeCobra-development-app",
---
>     "ApplicationName": {
>      "Ref": "Application"
>     },
335c339
<     "Analytics": "v2:deflate64:H4sIAAAAAAAA/1WPwWrDQAxEvyX3tRq7FNpjMTnkFhzotShrOShea4NWTg7L/ntxmpD2NMw8mGEaqOsPWK/wmirfj1XgA+S9oR9dRynO6snhNX1nT2o8sEejCQWPpJDbZ+baQf7Y4hgnyF0MN/LQrSRD8bTTOHCg4ihgMvYHwoWEEXI7yOf5HJYWjuL+2y/SdE83cmGNMpFYcRpno7dXWKbIR+33ZOU2fP9QnMSe4JReLs0a6neoV6fEXOksxhNB96s/DdtzHA4BAAA="
---
>     "Analytics": "v2:deflate64:H4sIAAAAAAAA/1WPzW7CQAyEn4X7xiWgSu2xRBy4oSD1iszGqUw2XrR24LDad0ehVG1Po29m5J8V1PU7LBd408p3QxX4BPlg6AfXksYpeXJ402P2lIx79mg0ouAXJcjNr+eaXv5gcYwj5DaGR/KjO1FD8bRPsedAxekaNpMfyDao5CigGvsT4dwLA+Sml4/LJcwzOYr7j5+U9Olu5copykhixaU4Gb2uYV5MPqbuQFYeZzw/Kk5iR3DWl+tqCfUb1IuzMldpEuORoP3WO6rBml4cAQAA"
```


.github/workflows/cdk_deploy.yml
```diff
@@ -0,0 +1,103 @@
...
+        type: string
+
+env:
+  EMAIL_USER: ${{ secrets.EMAIL_USER }}
+  EMAIL_PASS: ${{ secrets.EMAIL_PASS }}
+  JOBS_TOKEN: ${{ secrets.JOBS_TOKEN }}
+  PATREON_CLIENT_ID: ${{ secrets.PATREON_CLIENT_ID }}
+  PATREON_CLIENT_SECRET: ${{ secrets.PATREON_CLIENT_SECRET }}
+  PATREON_HOOK_SECRET: ${{ secrets.PATREON_HOOK_SECRET }}
+  REDIS_HOST: ${{ secrets.REDIS_HOST }}
```
KaelenProctor, line 38: @dekkerglen is Redis still needed after the drafting changes? I do see the environment variable referenced in code though the redis utility doesn't appear to have references

almarcotte: My understanding is that it's no longer needed and that's why I didn't add it to CDK but I'll add it if @dekkerglen confirms it's still needed for something. I think it's ElastiCache right?

dekkerglen: Correct, redis is gone!

infra/jest.config.js
```diff
@@ -0,0 +1,8 @@
+export default {
```
KaelenProctor, line 1: I started to play around in this branch a bit and for me I could not get jest to run with export default, I had to switch it to module.export = . Is there something I might be missing? Within infra folder I did npm install, npm build, then npm test

almarcotte: Eh that's weird, this is exactly how `cdk init` sets it up but clearly there's a piece missing for it to work. I pushed a change that fixes it 👍

infra/lib/cubecobra-stack.ts
```diff
@@ -0,0 +1,139 @@
...
+    AWS_REGION: props?.env?.region || '',
+    CACHE_ENABLED: 'false',
+    CUBECOBRA_VERSION: params.version,
+    DATA_BUCKET: params.dataBucket,
+    DOMAIN: params.domain,
+    DOWNTIME_ACTIVE: params.downTimeActive ? 'true' : 'false',
+    DYNAMO_PREFIX: params.dynamoPrefix,
+    EMAIL_CONFIG_PASSWORD: params.emailPass,
+    EMAIL_CONFIG_USERNAME: params.emailUser,
+    ENV: params.env,
```
KaelenProctor, line 120: @almarcotte Could you add NODE_ENV here, using params.env? I was going to push that change to your branch but github permissions are rejecting that.

@dekkerglen Express will read NODE_ENV and that will pass to the PUG template for which React version to load, fixing production CubeCobra using react development. [Since NODE_ENV isn't set Express defaults to 'development'](https://expressjs.com/en/4x/api.html#app.settings.table). As well setting NODE_ENV=production should improve performance based on https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production

almarcotte: Added 👍

.github/workflows/build_and_release.yml
```diff
@@ -0,0 +1,76 @@
...
+          cache: npm
+
+      - run: npm ci
+
+      - name: Run Semantic Release
+        id: semantic_release
+        run: |
+          npx semantic-release --ci
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
almarcotte, line 29: This _should_ work because it's not a PR coming in from a fork. `semantic-release` needs write access to make a commit (update `version` in `package.json`) and create a tag.

.github/workflows/build_and_release.yml
```diff
@@ -0,0 +1,76 @@
...
+  deploy:
+    uses: ./.github/workflows/cdk_deploy.yml
+    needs:
+      - build
+      - release
+    strategy:
+      matrix:
+        target:
+          - development
+          - production
```
almarcotte, line 72: This will create 2 parallel jobs but both require manual approval before starting so development can be deployed first. We could have a `deploy_dev` and `deploy_prod` job that needs the dev one instead.

PR #2577, opened 2025-01-25
More dynamo conversions: Packages and Password Resets

KaelenProctor:
# Testing

## Password reset

Flow:
1. Logout and then go to forgot password from the login dialog
2. Enter valid email address
3. Extract the reset ID from dynamo since the email doesn't get sent locally
4. Go to the /user/passwordreset/<id> page using the id from dynamo
5. Complete the password reset
6. Login with new password

I plan to make some adjustments to the email sending to support local SMTP. As I've done in past jobs using a docker container like https://mailpit.axllent.org/docs/install/docker/ makes it easy to see the emails sent. That will require:
1. Refactor SMTP transport creation into a single place, use development environment variable to determine Gmail or local
2. Update email templates to take DOMAIN environment variable instead of CubeCobra.com hardcoded. Similarly handle HTTPS or HTTP based on an environment variable
3. Add mailer to the docker compose setup I use, which then probably should be checked in and documented

## Packages

1. Load packages page
2. Filter by keywords and validate packages listed match
3. Tested different status filters, sort by, and sort ascend vs descend
4. Vote and unvote a package
5. Create a new package
6. Delete a package
7. Add a package to a cube


src/client/components/card/CardPackage.tsx
```diff
@@ -46,7 +46,7 @@ const CardPackage: React.FC<CardPackageProps> = ({ cardPackage }) => {
       }
     }
     setLoading(false);
-  }, [cardPackage.id, loading, voted]);
+  }, [cardPackage.id, csrfFetch, loading, voted]);
```
KaelenProctor, line 49: Lint fix

src/client/pages/PackagesPage.tsx
```diff
@@ -34,7 +35,7 @@ const PackagesPage: React.FC<PackagesPageProps> = ({ loginCallback = '/', items,
   const user = useContext(UserContext);
   const { csrfFetch } = useContext(CSRFContext);
 
-  const [type, setType] = useQueryParam('t', 'a');
+  const [type, setType] = useQueryParam('t', CardPackageStatus.APPROVED);
```
KaelenProctor, line 38: Might as well use the constant/enum

src/datatypes/CardPackage.ts
```diff
@@ -1,16 +1,27 @@
...
+}
 
-export default interface CardPackage {
-  id: string;
+export type UnhydratedCardPackage = {
+  id?: string;
   title: string;
-  date: string;
+  date: number;
   owner: string;
```
KaelenProctor, line 13: Before owner was typed only as a string even though hydration clearly set it as a User

src/dynamo/models/package.ts
```diff
@@ -0,0 +1,278 @@
...
+};
+
+const hydrate = async (pack?: UnhydratedCardPackage): Promise<CardPackage | undefined> => {
+  if (!pack) {
+    return pack;
+  }
+
+  const owner = await User.getById(pack.owner);
+  const cards = pack.cards.map((c) => {
+    return cardFromId(c);
```
KaelenProctor, line 71: Removed the if statement for c.scryfall_id is set. The cards list stored in dynamo is a list of strings, so in hydration they are never objects

src/dynamo/models/package.ts
```diff
@@ -0,0 +1,278 @@
...
+    if (document.owner && typeof document.owner !== 'string' && document.owner.id) {
+      ownerId = document.owner.id;
+    } else if (document.owner && typeof document.owner === 'string') {
+      ownerId = document.owner;
+    }
+
+    let cardIds: string[] = [];
+    if (document.cards) {
+      cardIds = document.cards.map((card) => {
+        if (typeof card !== 'string' && card.scryfall_id) {
```
KaelenProctor, line 135: In comparison when saving a package it might be a hydrated version, such as when voting/unvoting, so have to handle the cards list being list of string or list of CardDetails

PR #2578, opened 2025-01-26
Started as a email refactor to use a local mail server, ended up fixing several bugs related to content

KaelenProctor:
# Refactoring
1. Consolidated the email transport/sending code into a utility file, with a simple sendEmail export
2. That makes it easy to configure different email transports for production or local. Though I did learn how to configure my gmail with an app password in order to send emails from local. As well the code now gracefully warns if the required email config options are not set, rather than causing a server error

Example of local email server with mailpit + docker:
mailpit-local-email-testing

# Bugs fixed
1. The content creator application submission was not doing anything
2. The admin routes related to content / contributors failed to send emails because the hydrated User properties don't have email on them
4. Also fixed bugs with content model where put actions would replace the user/owner object with just their ID before writing to dynamo, with the side effect of that changing the input variable. Then afterwards in the route the object is different and doesn't match the type (looking to the future with typescript)
5. Podcast and videos send for review used an incorrect review status
6. Fixed the video content page not loading. It expected a list of videos but was getting one

# Testing

## Content creator application submission

Before:
Clicking the button did nothing
old-submit-content-creator-broken

After:
new-submit-content-creator-working

## Failing to send content / contributor emails + model replacing user/owner

Before:
Can see the error as "no recipients" defined. The hydrated User object doesn't have email, have to explicitly get the user with sensitive data to have their email. Probably should adjust the email property on the User type for clarity.
old-failed-to-send-review-email

After:
new-content-published
new-content-reject-email
new-submit-content-creator-approve

## Podcast and videos send for review used an incorrect review status

Before:
old-podcast-submit
old-video-submit
old-podcast-not-showing-for-review

After:
new-video-for-review

## Video page not loading

Before:
old-video-broken

After:
new-video-working

# Next steps
1. Want to refactor the emails so that DOMAIN and http vs https are used as variables, so that local emails work well. Currently they all go to production so have to always copy and edit
2. Some errors in the content reject email, where the content type isn't printed


src/dynamo/models/content.js
```diff
@@ -191,16 +191,23 @@ module.exports = {
       throw new Error('Invalid document: No partition key provided');
     }
 
+    //If the owner is a User object, get its Id without modifying the incoming document
+    let ownerId;
     if (document.owner.id) {
-      document.owner = document.owner.id;
```
KaelenProctor, line 195: In the old code after this method finishes the document object was modified, which could cause unexpected failures in a situation like:
```
//in route
const content = Content.getById(...)
//content.owner = User object
content.title = 'blah'
await Content.update(content)
//content.owner = user id string
do something with content.owner.blah fails because its no longer an object
```
In the typescript conversions of dynamo I've been removing this behaviour unintentionally, because its creating a new object for the put action from the original document.

src/dynamo/models/notice.js
```diff
@@ -81,13 +81,17 @@ module.exports = {
     };
   },
   put: async (document) => {
+    let userId;
     if (document[FIELDS.USER].id) {
-      document[FIELDS.USER] = document[FIELDS.USER].id;
```
KaelenProctor, line 85: Same as in content

src/routes/admin_routes.js
```diff
@@ -70,41 +68,15 @@ router.get('/publish/:id', ensureAdmin, async (req, res) => {
...
-    juiceResources: {
-      webResources: {
-        relativeTo: path.join(__dirname, '..', 'public'),
-        images: true,
-      },
-    },
-    transport: smtpTransport,
-  });
+    //Normal hydration of User does not contain email, thus we must fetch it in order to notify about their application
+    const owner = await User.getByIdWithSensitiveData(document.owner.id);
```
KaelenProctor, line 73: Hydrated user objects don't have email. In order to send the emails to the applications we must load their User with their sensitive data in order to have email

src/routes/content_routes.js
```diff
@@ -476,7 +476,7 @@ router.post('/submitpodcast', ensureContentCreator, async (req, res) => {
   podcast.description = fields.description;
   podcast.podcastLink = fields.url;
   podcast.image = fields.image;
-  podcast.status = 'inReview';
+  podcast.status = Content.STATUS.IN_REVIEW;
```
KaelenProctor, line 479: Using constants

src/util/email.ts
```diff
@@ -0,0 +1,73 @@
...
+    return {
+      name: 'CubeCobra.com',
+      secure: true,
+      service: 'Gmail',
+      auth: {
+        user: process.env.EMAIL_CONFIG_USERNAME,
+        pass: process.env.EMAIL_CONFIG_PASSWORD,
+      },
+    };
+  } else if (process.env.EMAIL_CONFIG_SMTP_HOST && process.env.EMAIL_CONFIG_SMTP_PORT) {
```
KaelenProctor, line 22: The goal is to use a local mail server here such as mailpit with docker. That is an easier setup than getting developers to create app passwords for their gmail (in large part because the google UI for it sucks)

src/util/email.ts
```diff
@@ -0,0 +1,73 @@
...
+export const sendEmail = async (
+  to: string,
+  subject: string,
+  templateName: string,
+  templateLocals: Record<string, any> = {},
+): Promise<any> => {
+  const transport = getTransport();
+  if (!transport) {
+    // eslint-disable-next-line no-console -- Warn so clear why an email isn't being sent
+    console.warn('No email transport is configured, skipping sending');
```
KaelenProctor, line 48: Given the developer setup instructions don't include how to setup gmail for emails to start, the default behaviour meant sending emails just caused server errors. Rather than do that, might as well be graceful and not error.

KaelenProctor:
It might be worth checking in production if there are podcasts or videos in the wrong review status as `typeStatusComp = 'v:inReview'` or `typeStatusComp = 'p:inReview'`, and then either deleting them as old or updating inReview to `r`

PR #2580, opened 2025-01-27
Adjust plain name search grammar so that apostrophe and colons in names aren't rejected

KaelenProctor:
This allows name searching, without needing quotes, for names like:
* Jace, Vryn's Prodigy
* Circle of Protection: Black

For the details of the grammar I referenced https://github.com/dekkerglen/CubeCobra/pull/1647. From that it looks pretty safe to me to remove the apostrophe and colon exclusions from the last part of noQuoteStringValue which is trying to match all the remaining text in the token, after and/or are past. @lunakv if you are still active I'd love to hear your thoughts on the change

# Testing

## Before

Searching using double or single quoted

Searching in cube:

old-cop-1
old-cop-fail

old-jace-search1
old-jace-search-fail
old-jace-search2

Card search
old-apostorphe-search-fail
old-cop-card-search

## After

Now searching for cards with apostrophe and colon in their name works without quotes.

new-cop-card-search
new-apostorphe-search-fail
new-cop
new-jace

## Additional testing
I gathered all the unique characters in card names using:
```
jq '.[]' cubecobra/private/names.json -r | awk '{for(i=1;i<=NF;i++)if(!a[$i]++)print $i}' FS=""  | sort
```
Then using scryfall I grabbed one or two cards whose name contains non-alphabetic characters giving me a list of:

- Jace, Vryn's Prodigy
- "Ach! Hans, Run!" (quotes literal)
- Busted!
- Minsc & Boo, Timeless Heroes
- Hazmat Suit (Used)
- +2 mace
- Jace, Wielder of Mysteries
- Abzan Kin-Guard
- Mr. Orfeo, the Boulder
- T-45 Power Armor
- Borrowing 100,000 Arrows
- Circle of Protection: Black
- TL;DR
- Question Elemental?
- _____ Goblin
- The Ultimate Nightmare of Wizards of the Coast® Customer Service
- Ratonhnhaké꞉ton

After the change of those only `Hazmat Suit (Used)` and `Ratonhnhaké꞉ton` don't work, see below.

# Some expressions that don't pass still

* `name:'Circle of protection: red'` - It parses correctly but returns 2 results from the filter parser, and the code only expects 1. Interestingly `name:"Circle of protection: red"` works
* `Hazmat Suit (Used)` - Similarly fails because of 2 results
* `n:Circle of protection: black` - Again 2 results
* `'Jace, Vryn's Prodigy'` - Fails to parse
* `Ratonhnhaké꞉ton` - They filter parses but the name_lower of Ratonhnhaké꞉ton replaced the é with e as part of the normalization (from update_cards using cardutil.normalizeName), so in the comparison function it doesn't match.

### DIFF
--- a/nearley/values.ne
+++ b/nearley/values.ne
@@ -167,7 +167,7 @@ noQuoteStringValue ->
     | "and"i [^ \t\n"'\\=<>:] 
     | "o"i [^rR \t\n"'\\=<>:]
     | "or"i [^ \t\n"'\\=<>:]
-    ) [^ \t\n"'\\=<>:]:* {% ([startChars, chars]) => startChars.concat(chars).join('').toLowerCase() %}
+    ) [^ \t\n"\\=<>]:* {% ([startChars, chars]) => startChars.concat(chars).join('').toLowerCase() %}
 # "
 
 manaCostOpValue -> equalityOperator manaCostValue {% ([op, value]) => manaCostOperation(op, value) %}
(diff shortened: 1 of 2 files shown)


KaelenProctor:
On a related note, I have a working branch where I've converted nearley to output typescript but I am planning on finishing the dynamo conversions first. Relevant with respect to all the eslint errors from the CI tests.

KaelenProctor:
This unexpectedly broke filters like rarity, so that sucks. Will try and figure out why...

PR #2583, opened 2025-01-29
Continue email refactor to consolidate http vs https and domain

KaelenProctor:
These changes make life easier for a developer working locally with CubeCobra. Also helps with the CubeCobra dev site though that seems to have fallen to the wayside.

Changes:
1. HTTP_ONLY environment variable with must be "true" (string) in order to use http:// scheme
2. Pass base url consisting of scheme + DOMAIN (eg http://localhost:8080) to email templates and to the frontend client
3. Renamed DomainContext to BaseUrlContext to reflect the difference
4. Used baseUrl instead of hard-coded cubecobra.com in various view

An additional toss in was deleting CubeDraftStaging which I found no usages of, which makes sense since the DraftMancer integration there is no organization of drafts with other people on CubeCobra.

# Testing
I also tested the QR code for cubes but there isn't a great way to screenshot that to prove lol

## Before
old-share-comment-scheme-doesnt-match
old-meta-url-matches-prod
old-email-matches-prod

## After
Share comment URL matches local dev env:
new-share-comment-base-url-matches-dev-env

URL links in emails match local dev env:
new-email-with-local-domain-gmail
new-email-with-local-domain

Meta URLs match:
new-meta-url-matches-dev-env3
new-meta-url-matches-dev-env2
new-meta-url-matches-dev-env


src/client/components/cube/CubeOverviewCard.tsx
```diff
@@ -35,7 +36,7 @@ interface PrivateCubeIconProps {
 }
 
 const PrivateCubeIcon: React.FC<PrivateCubeIconProps> = ({ visibility }) => {
-  const visibilityWord = visibility == 'pr' ? 'private' : 'unlisted';
+  const visibilityWord = visibility === 'pr' ? 'private' : 'unlisted';
```
KaelenProctor, line 39: This and the eslint ignores are to cleanup style warnings

src/router/routes/cube/blog.ts
```diff
@@ -212,7 +213,7 @@ export const getBlogPostsForCubeHandler = async (req: Request, res: Response) =>
           `Cube Cobra Blog: ${cube.name}`,
           cube.description,
           cube.image.uri,
-          `https://cubecobra.com/cube/blog/${encodeURIComponent(req.params.id)}`,
+          `${baseUrl}/cube/blog/${encodeURIComponent(req.params.id)}`,
```
KaelenProctor, line 216: I thought about removing the base URL here and appending within generateMeta but decided not to because it seemed like the meta URL doesn't have to always be to cubecobra.com. The image URI meta is only a couple times (eg sample pack) pointing to cube cobra

src/util/util.js
```diff
@@ -90,7 +90,9 @@ function hasProfanity(text) {
 
 function validateEmail(email) {
   if (email.includes('+')) {
-    throw new Error('CubeCobra does not support plus email addressing, please remove the "+descriptor" from your email and try again.');
+    throw new Error(
```
KaelenProctor, line 93: Style fixes my IDE threw in..

emails/application_approve/html.pug
```diff
@@ -11,7 +11,7 @@ html
                             tbody
                                 tr
                                     td(align='center', bgcolor='#343A40', style='padding: 40px 0 30px 0; color: #ffffff; font-size: 28px; font-weight: bold; font-family: Arial, sans-serif;')
-                                        img(src='https://cubecobra.com/content/banner.png', alt='Cube Cobra', height='50', style='display: block;')
+                                        img(src=`${baseUrl}/content/banner.png`, alt='Cube Cobra', height='50', style='display: block;')
```
KaelenProctor, line 14: FYI I found out that by default the email-template library will inline, as base64 data, images smaller than a default threshold (because we have juiceResources.webResources.image = true, per https://github.com/forwardemail/email-templates?tab=readme-ov-file#options). However many email clients including gmail don't support base64 data per https://www.caniemail.com/features/image-base64/, causing the image not to show.

So wondering if we want to disable that option (or ignore with data-inline-ignore attributes) so the image loads via the web?

dekkerglen (approved):

PR #2584, opened 2025-01-30
Update draft card picking such that unfiltered slots are done after all filtered slots

KaelenProctor:
# Problem
As reported in discord a complex custom draft format would fail to fire with the reason that no cards could match a filter, but each time the filter that failed would change. We investigated and determined the problem was that the unfiltered card slots, which are processed in seat/pack order, would randomly choose a card resulting in subsequent filtered slots to have no cards remaining that matched the filter. This is a problem that is most likely to occur in a 360 card cube for a regular 8 seat, 3 packs, 15 cards per pack, though more generally if the number of cards in the cube matched the total cards needed for the draft format.

Example of one of the packs setup for the failing draft:
card-lots-for-blb-draft

# Solution
As discussed the logical path is to complete all the filtered cards slots before unfiltered ones. That way all the filtered card slots are satisfied from the total cube set of cards, and then the unfiltered slots pull randomly from the remaining set.

# Testing
Used the BLB cube (https://www.cubecobra.com/cube/list/blbsetcube) referenced by the person having problems drafting. Exported their cube via CSV to my local dev environment, then setup a custom draft format the same as theirs.

I wrote the unit tests after the implementation change which is not exactly the best practice. I did run them against master but without using a truly random nextCardFunction it is hard to clearly see the master branch failing and this branch passing (and true random in tests is not ideal for consistency).

## Before

In 3 tries to launch the draft got 3 different filters which failed to find a card.
old-blb-fail3
old-blb-fail2
old-blb-fail1

## After

With the changes the draft works. I verified the cards in each slot match the filters from the pack 1 draft setup; admittedly I had to double check a bunch mainly because of the special guests (which vary in rarity and set they come from) and because the "commoncolorless" tag includes lands, artifacts, and multicolored cards in this cube.
new-successful-pack-creation

Also updated the custom draft modal to explain the change of unfiltered slots being done last
new-description


src/client/drafting/createdraft.ts
```diff
@@ -142,38 +142,128 @@ export const getDraftFormat = (params: DraftParams, cube: Cube): DraftFormat =>
   return createDefaultDraftFormat(params.packs, params.cards || 15);
 };
 
-const createPacks = (format: DraftFormat, seats: number, nextCardFn: NextCardFn): CreatePacksResult => {
-  let ok = true;
-  let messages: string[] = [];
```
KaelenProctor, line 147: The jest tests were failing because messages was always empty in the result even when the stub nextCardFn in my test returned messages. On further reflection I think both messages and ok would never be correctly set on the result object, but that wasn't noticeable because the actual nextCardFn implementation throws an Error when it cannot find a card, thus bypassing result being returned in an error state.

Also exacerbated by the fact that the method returned a variable named "result" and the return of nextCardFn was assigned to "results" as well. This I improved by renaming the latter to cardResult.

tests/cube/draft/createPacks.test.ts
```diff
@@ -0,0 +1,448 @@
+import seedrandom from 'seedrandom';
```
KaelenProctor, line 1: Even though createPacks lives in the createdraft.ts file, it felt more logical to me to name the test after the function

PR #2585, opened 2025-01-31
Tag color contrast between background and foreground

### LINKED ISSUE #2579: Color Contrast Issues with Tags
**Describe the bug**
When selecting a tag color, the contrast between the background color and the text often makes it difficult to read what the card name is. 

**To Reproduce**
1. Open a card
2. Add a tag to the card
3. Click "Display -> Set Tags Colors"
4. Select a color from the available dropdown
5. Save & view the color with text

**Expected behavior**
- Whenever a tag color is selected that does not meet a11y accessibility guidelines, the text will invert to use a lighter color.
- Tag colors are preselected to provide proper contrast against the text in order to prevent the readability issues.

**Screenshots**
I have included screenshots demonstrating the problematic color combinations, excluding Orange and Yellow.

Image
Image
Image
Image

 - OS: macOS 15.2
 - Browser Arc
 - Version 1.78.1

KaelenProctor:
Fixes #2579

# Problem
Tag background colors did not contrast with the foreground text color well in many cases, with different colors matching poorly with the text depending on light or dark mode.

# Solution
In order to have good contrast, which I've defined as WCAG 2.0 level AAA ratio of 7.5 (which a site like https://webaim.org/resources/contrastchecker/ checks) we need to set the text color for each tag instead of depending on the theme text color. As well I had to adjust some of the background colors a bit to hit the 7.5 target.

A couple other improvements I tossed in:
1. The tags in the autocard hover now are colored basesd on the cube settings. Had to pass the TagColors of the cube through the WithAutocard component into the AutocardContext.showCard, given that AutocardContext provider is above the TagColor provider and thus would not have any tag colors initialized if it tried to use the provider.
2. In changelogs for edited cards, use the new card version instead of the old. **This makes sense to me but if it is not desired I can revert**

# Testing

## Before

Can see in light and dark themes that some of the text colors don't contrast well against the background colors:
old-tag-colors-dark
old-tag-colors-light
In the list view:
old-tag-colors-in-list-view-light
old-tag-colors-in-list-view-dark

Can also see that in the autocard hover tags don't have a background color even though the card itself in the cube list does
old-tag-colors-dont-show-in-autocard

In the changelog the cube shows the old version of a card. Here I added 4 tags to the card, then removed 2 of them. In the 2nd changelog it is showing the old version of the card with 4 tags when it makes more sense to show the updated state with 2 tags:
old-changelog-edit-shows-old-version

## After

Now can see the text color doesn't depend on the theme, but is chosen based on contrast to the background. In order to use only 2 text colors I had to modify slightly some of the background colors to hit the desired contrast ratio.
new-tag-colors-light-mode
new-tag-colors-dark-mode

In the table view:
new-tag-colors-in-list-view
In the list view:
new-tag-colors-in-table-view

### Tag colors in the autocard
new-tag-color-in-autocard
new-multiple-tags-in-autocard

### Changelog shows new version of edited card
Now in the latest changelog we see the card after 2 of the 4 tags were removed:
new-autocard-in-changelog-shows-tags2

### Outside the cube cards page the autocard has no differences:
Cube tags:
cube-tags-unchanged

Recommendations page
new-autocard-outside-cube-list
Autocomplete
new-autocard-in-autocomplete
Markdown
autocard-in-markdown
Card versions
card-page-versions-autocard

### DIFF
--- a/public/css/tags.css
+++ b/public/css/tags.css
@@ -1,57 +1,69 @@
 .tag {
-	background-color: var(--header);
-	font-size: 85%;
-	padding: 0.5em .75em;
-	margin: .25em .1em;
-	display: inline-block;
-	border: 1px solid var(--border-light);
-	transition: all .1s linear;
+  background-color: var(--header);
+  font-size: 85%;
+  padding: 0.5em 0.75em;
+  margin: 0.25em 0.1em;
+  display: inline-block;
+  border: 1px solid var(--border-light);
+  transition: all 0.1s linear;
 }
 
 .tag-sort-placeholder {
   height: 44px;
-	background-color: var(--bg-hover);
+  background-color: var(--bg-hover);
   margin-left: -15px;
   margin-right: -15px;
 }
 
 .tag-sort-helper {
-	background-color: var(--bg);
-	border: 1px solid var(--border-light);
+  background-color: var(--bg);
+  border: 1px solid var(--border-light);
 }
 
-.tag-color,
-.tag-color .tdname,
-.tag-color > a {
-  color: var(--bg) !important;
-}
+/* In order to have WCAG 2.0 level AA contrast between the tag background and foreground colors,
+ * we must explicitly set the foreground color rather than based on light or dark theme. Used
+ * https://webaim.org/resources/contrastchecker/ to check contrast and found colors that fit well
+ * for the backgrounds, though had to make small color adjustments to some background to get passing
+ * grades. Targeted WCAG 2.0 level AAA which needs a 7.5 contrast ratio between text and background,
+ * compared to AA which is 4.5 ratio.
+ */
 .tag-red {
   background-color: #8e1600 !important;
+  color: #f0f0f0 !important;
 }
 .tag-brown {
   background-color: #654321 !important;
+  color: #f0f0f0 !important;
 }
 .tag-orange {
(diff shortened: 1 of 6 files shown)


public/css/tags.css
```diff
@@ -1,57 +1,69 @@
 .tag {
-	background-color: var(--header);
-	font-size: 85%;
-	padding: 0.5em .75em;
-	margin: .25em .1em;
-	display: inline-block;
-	border: 1px solid var(--border-light);
-	transition: all .1s linear;
+  background-color: var(--header);
```
KaelenProctor, line 2: Style changes made by IDE

public/css/tags.css
```diff
@@ -1,57 +1,69 @@
...
-	background-color: var(--bg);
-	border: 1px solid var(--border-light);
+  background-color: var(--bg);
+  border: 1px solid var(--border-light);
 }
 
-.tag-color,
-.tag-color .tdname,
-.tag-color > a {
-  color: var(--bg) !important;
```
KaelenProctor, line 26: Removed this for a couple reason:
1. The text color needs to be based on the tag background color not the theme in order to have good contrast
2. I found no usages of tdname
3. The only time a tag renders as a link is in the CubeOverview and there it doesn't have tag-color class above it. Thus redundant

PR #2586, opened 2025-01-31
feat(deckbuilder): Display mana breakdown as deck stats

almarcotte:
Adding a new information panel when deck building that shows (for now) the mana breakdown of the deck. It includes the following:

1. Number (and percentage) of cards of each colour
2. Percentage of mana symbols for each colour
3. CMC curve by colour
4. Number of mana sources for each colour

It's inspired by (i.e. stolen) what Moxfield displays at the bottom of deck lists.

## How it works

I added a bunch of utilities to cards and "decks" to generate the numbers needed to populate the panel. This includes things like mana curve by colour, extracting all symbols from mana costs (including hybrid mana) and finding all mana sources in a deck.

If a deck includes fetches (either the OG fetches or anything that fetches a basic like Fabled Passage) those will count as a source for all the colours they can fetch. The tests probably can explain how this works better than I can but in short: If your deck has an Arid Mesa, a Steam Vent and a Plains then Arid Mesa will count as a source of U, R and W. Similarly if the deck has a Fabled Passage it will count as a source of each colour of basics in the deck.

A lot of this piggybacks the `produced_mana` field from the Scryfall data which we weren't ingesting before so I've updated `update_cards.ts`. It's worth mentioning that `produced_mana` isn't only for lands (Dark Ritual for instance has `produced_mana = ['B']`) but for now a "source" is a land. This means things like moxen are also not counted as "sources".

Is some of this overkill? Maybe. But I figured these tools might be useful going forward especially since some of these metrics could be applied to a whole cube (number of sources of each colour seems particularly useful).

Cards will count towards the count of each colour they have in their cost. Mana symbols are similar except {W}{W} would add 2 to the white count. Hybrid mana counts as 2 separate symbols. Leyline of the Guildpact for example has 8 symbols. Phyrexian mana adds a symbol for the colour. 

## I suck at frontend

I tried as best as I could to replicate the look of similar components I could find but I have no clue what I'm doing. Ideally someone would checkout this branch and do a pass at the look and feel.

### Classic example

Testing

### Hybrid Mana

Testing with hybrid mana symbols

### Vertical 

mobile-vertical

### Horizontal

mobile-horizontal

### DIFF
--- a/src/client/utils/cardutil.ts
+++ b/src/client/utils/cardutil.ts
@@ -1,13 +1,16 @@
-import Card, {
-  CardDetails as CardDetailsType,
-  COLOR_CATEGORIES,
-  ColorCategory,
-  DefaultElo,
+import {
   isGenericHybridManaSymbol,
   isHybridManaSymbol,
   isManaSymbol,
   isPhyrexianManaSymbol,
   ManaSymbol,
+} from 'datatypes/Mana';
+
+import Card, {
+  CardDetails as CardDetailsType,
+  COLOR_CATEGORIES,
+  ColorCategory,
+  DefaultElo,
 } from '../../datatypes/Card';
 import CategoryOverrides from '../res/CategoryOverrides.json';
 import LandCategories from '../res/LandCategories.json';
--- a/src/client/utils/cardutil.ts
+++ b/src/client/utils/cardutil.ts
@@ -3,6 +3,10 @@ import Card, {
   COLOR_CATEGORIES,
   ColorCategory,
   DefaultElo,
+  isGenericHybridManaSymbol,
+  isHybridManaSymbol,
+  isManaSymbol,
+  ManaSymbol,
 } from '../../datatypes/Card';
 import CategoryOverrides from '../res/CategoryOverrides.json';
 import LandCategories from '../res/LandCategories.json';
@@ -42,6 +46,15 @@ export const COLOR_COMBINATIONS: string[][] = [
   ['W', 'U', 'B', 'R'],
   ['W', 'U', 'B', 'R', 'G'],
 ];
+
+export const BASIC_LAND_MANA_MAPPING: { [key: string]: ManaSymbol } = {
+  Plains: 'W',
+  Island: 'U',
+  Swamp: 'B',
+  Mountain: 'R',
+  Forest: 'G',
+  Waste: 'C',
+};
 
 // export const COLOR_INCLUSION_MAP: Record<string, Record<string, boolean>> = fromEntries(
 //   COLOR_COMBINATIONS.map((colors) => [
@@ -417,6 +430,10 @@ export const cardColors = (card: Card): string[] => {
   return [];
 };
 
+export const cardColorsAsManaSymbols = (card: Card): ManaSymbol[] => {
+  return cardColors(card).map((symbol: string) => symbol.toUpperCase() as ManaSymbol);
+};
+
 export const cardLanguage = (card: Card): string => card.details?.language ?? '';
 
 export const cardMtgoId = (card: Card): number => card.details?.mtgo_id ?? -1;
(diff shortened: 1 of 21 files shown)


src/datatypes/Card.ts
```diff
@@ -152,6 +156,18 @@ export interface Changes {
   version?: number;
 }
 
+export const BasicLands = ['Plains', 'Island', 'Swamp', 'Mountain', 'Forest', 'Waste'] as const;
+export type BasicLand = (typeof BasicLands)[number];
+
+export const BASIC_LAND_MAPPING: Record<BasicLand, ManaSymbol> = {
```
KaelenProctor, line 162: This is duplicate within src/client/utils/cardutil.ts, should consolidate

src/util/ml.js
```diff
@@ -2,7 +2,7 @@ const tf = require('@tensorflow/tfjs-node');
 const fs = require('fs');
 const cloudwatch = require('./cloudwatch');
 
-const indexToOracle = JSON.parse(fs.readFileSync('./model/indexToOracleMap.json'));
+const indexToOracle = process.env.NODE_ENV === 'test' ? {} : JSON.parse(fs.readFileSync('./model/indexToOracleMap.json'));
```
KaelenProctor, line 5: In https://github.com/dekkerglen/CubeCobra/pull/2585/files#diff-4c3cfde2c8298e64ca709feb98ce653c1f25d523dc113f5fc097d9314ec785ef I've set Jest to automock this ML classes because of all of these model loads. We should pick one of the two solutions to that for now, until ml.js can be updated to not auto-load models (instead app.js calls a function and awaits)

yarn.lock
```diff
@@ -1686,29 +1686,41 @@
   resolved "https://registry.npmjs.org/@humanwhocodes/retry/-/retry-0.3.0.tgz"
   integrity sha512-d2CGZR2o7fS6sWB7DG/3a95bGKQyHMACZ5aW8qGkkqQpUoZV6C0X7Pc7l4ZNMZkfNBf4VWNe9E1jRsf0G146Ew==
 
-"@img/sharp-libvips-linux-x64@1.0.4":
+"@img/sharp-darwin-arm64@0.33.5":
```
KaelenProctor, line 1689: @almarcotte Were any libraries changed in the PR? I see not package.json or package-lock.json updates. Since yarn wasn't working in CI we are supposed to be using npm instead of yarn, so perhaps the yarn.lock isn't even needed anymore...
Given @dekkerglen has noted that sharp can have install problems during release, it is worrying to see the packages change from linux-x64 to arm64

almarcotte: That looks like WebStorm doing some stuff on its own, I'll revert this change because it definitely doesn't belong there. Good catch 👍

src/util/ml.js
almarcotte: I like your solution a lot more, I wasn't a fan of having to modify the actual code. I'll remove this and rebase once your PR is merged (which I'm guessing is going to be before this one anyway)

src/datatypes/Card.ts
almarcotte: Oops! Good catch. Not sure what I was doing there. I'll remove this one.

src/client/utils/cardutil.ts
```diff
@@ -233,38 +249,7 @@ export const cardColorCategory = (card: Card): ColorCategory => {
...
-    if (colors.includes('R')) {
-      return 'Red';
-    }
-    if (colors.includes('G')) {
-      return 'Green';
-    }
-  }
-
-  return 'Colorless';
+  return cardColorIdentityCategory(card);
```
almarcotte, line 252: This was duplicated in `cardColorIdentityCategory` so I'm combining both.

src/jobs/update_cards.ts
```diff
@@ -243,14 +246,23 @@ async function writeFile(filepath: string, data: any) {
...
+      .filter(
+        (element) =>
+          element.component === 'token' ||
+          element.type_line.includes('Emblem') ||
+          element.type_line.includes('Dungeon') ||
+          (element.type_line.includes('Card') && !element.name.includes('Checklist')) ||
+          miscTokens.includes(element.name),
+      )
+      .map(({ id }) => id)
+  );
```
almarcotte, line 265: IDE linting, no actual changes

src/jobs/update_cards.ts
```diff
@@ -56,6 +57,8 @@ interface ScryfallCard {
   cmc: number;
   colors: string[];
   color_identity: string[];
+  keywords: string[];
+  produced_mana: string[];
```
almarcotte, line 61: `keywords` contains things like 'Trample' or 'Flying' but also ['Landcycling', 
Mountaincycling'] on cards like Oliphaunt. Lots of potential here with improving stats when deck building and other places.

I'm using `produced_mana` when finding the sources in a deck.

tests/test-utils/data.ts
```diff
@@ -25,6 +26,10 @@ export const createCard = (overrides?: Partial<Card>): Card => ({
   ...overrides,
 });
 
+export const createCardFromDetails = (overrides?: Partial<CardDetails>): Card => {
+  return createCard({ details: createCardDetails(overrides) });
+};
+
```
almarcotte, line 32: I was doing `createCard({details: createCardDetails({}))` a lot...

PR #2587, opened 2025-02-01
Conversion of notifications dynamo model to typescript and a few bug fixes found on the way

KaelenProctor:
# Bugs

1. Notifications for cloning a draft deck (endpoint is rebuilding), following a cube, and starting a sealed pool were not sending. Putting the updated cube into dynamo mucks with the owner value from an object to a string. Then util.addNotification trying `${to.id}` gets undefined and though that stores into dynamo, the notification does nothing
2. Notifications for mentions in comments/blog posts were not sending. The mentions request is a semi-colon separated string but the backend treated it like an array, meaning that it actually iterated over each character to get the user by id which wouldn't find the user
3. Adding your cube to the featured cube caused a backend error and did nothing. The UI used a link button meaning it did a GET, but the backend expected a POST

# Solutions
1. Before the cube is updated store the cube owner in a variable and use that for util.addNotification
2. Split the mentions body by semi-colon and then iterate. Updated tests accordingly
3. Used onClick action with csrfFetch to do a POST

# Testing

Validation of the semi-colon separated mentions format:
mentions-form-format

## Before

Creating a sealed pool in another person's cube doesn't notify the owner:
old-sealed-doesnt-notify-owner
Following another person's cube doesn't notify the owner:
old-cube-follow-doesnt-notify-owner
Mentioning users in blog posts or comments doesn't notify them:
old-mention-doesnt-notify-user

Can see from debugging how after the cube is updated when incrementing numDecks, the owner is now the user's id instead of the user object.
old-cube-owner-mucked-with-after-save

Error trying to add your cube to featured:
old-failed-to-add-cube-to-featured

## After

Test cases:

- Actions that created notifications, as seen in screenshots below
- Seeing unread notifications in the nav bar
- Going to notifications list page
- Clicking on a notification which marks it as read and redirects you to the notification context (eg cube, comment)
- Marking all notifications as read

Deck clone notification to cube owner:
new-rebuilt-deck-notification-to-cube-owner
Deck clone notification to originator of the deck:
new-rebuilt-deck-notification-to-original-deck-creator
Following a cube:
new-cube-follow-notification
Cloning a cube:
new-cloned-cube-notification
Following a user:
new-user-follow-notification
Content creator notifications:
new-content-notifications2
Add cube to featured:
image
Mention in comment/blog post notifies the mentionees:
image
Notification for starting a sealed:
image


src/util/util.js
```diff
@@ -262,7 +262,7 @@ function fromEntries(entries) {
   return obj;
 }
 
-async function addNotification(to, from, url, text) {
+async function addNotification(to/*: User*/, from/*: User*/, url, text) {
```
KaelenProctor, line 265: Loving typescript more and more, hopefully we can speed up the migration

tests/comments/api.test.ts
```diff
@@ -490,7 +490,7 @@ describe('Add Comment', () => {
           body: 'This is a new comment',
           parent: 'parent-id',
           type: 'comment',
-          mentions: ['mention-user-1', 'mention-user-2'],
+          mentions: 'mentionuser1;mentionuser2',
```
KaelenProctor, line 493: The tests passed because the code looked like it took mentions as an array, but it is actually a string. I think to improve this we can create types for the request body and add validation middlewares

PR #2589, opened 2025-02-02
Fix deck delete button

KaelenProctor:
# Problem
Clicking button to delete deck navigates to the deck before you can take action in the modal

# Solution
Prevent the click event on the delete button from triggering the row click action

# Testing

## Before
old-deleting-deck-opens-modal-then-leaves

## After
new-deleting-deck-opens-modal-without-leaving

### DIFF
--- a/src/client/components/WithModal.tsx
+++ b/src/client/components/WithModal.tsx
@@ -6,11 +6,13 @@ export interface WithModalProps<U> {
   className?: string;
   modalprops?: Omit<U, 'setOpen' | 'isOpen' | 'toggle'>;
   altClick?: () => void;
+  //Commonly would set stopProgagation to true if the button to open the modal lives within another clickable block like a row
+  stopProgagation?: boolean;
 }
 
 const withModal = <T extends ElementType, U>(Tag: T, ModalTag: ComponentType<U>) => {
   const Result: React.FC<WithModalProps<U> & ComponentProps<T>> = (allProps: WithModalProps<U> & ComponentProps<T>) => {
-    const { children, className, modalprops = {}, altClick } = allProps;
+    const { children, className, modalprops = {}, altClick, stopProgagation } = allProps;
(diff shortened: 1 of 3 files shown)


src/client/components/WithModal.tsx
```diff
@@ -29,10 +31,14 @@ const withModal = <T extends ElementType, U>(Tag: T, ModalTag: ComponentType<U>)
...
+        }
+
         event.preventDefault();
         return toggle();
       },
-      [altClick, toggle],
+      [altClick, stopProgagation, toggle],
     );
 
     return (
```
KaelenProctor, line 44: Down here where <Tag ..> is setup the onClick is set to handleClick. So even if allProps had onClick, such as because Button supports onClick, its value won't be used.

src/client/components/base/Button.tsx
```diff
@@ -38,20 +38,20 @@ const Button: React.FC<ButtonProps> = ({
       'text-button-text': ['primary', 'danger', 'accent', 'secondary'].includes(color) && !outline,
       'text-button-text-secondary': !['primary', 'danger', 'accent', 'secondary'].includes(color) && !outline,
       'text-button-primary bg-transparent border border-button-primary hover:bg-button-primary hover:text-button-text':
-        outline && color == 'primary',
+        outline && color === 'primary',
```
KaelenProctor, line 41: These are unrelated lint fixes

src/client/components/base/Button.tsx
```diff
@@ -71,7 +71,6 @@ const Button: React.FC<ButtonProps> = ({
     <button
       className={classes}
       onClick={(e) => {
-        type !== 'submit' && e.preventDefault();
```
KaelenProctor, line 74: IDE warning that this line does nothing, since its not an if nor is the value stored in a variable

PR #2593, opened 2025-02-04
More Dynamo conversions: Feeds and Notices

KaelenProctor:
# Regression test cases
As usual I used debugging and breakpoints to ensure all the changes of the models were executed successfully.

## Feeds
- See feed items on your Dashboard
- Create blog post and see all cube followers have new items in their feed
- Adding cards to the cube from a package also adds to the follower's feeds

## Notices

- Report user, cube, and comments
- As admin, see those reports
- Ignore user, cube, and comment reports
- Remove comment makes it "Deleted by moderator"
- Ban user because of user or cube reports


src/dynamo/models/feed.ts
```diff
@@ -0,0 +1,102 @@
...
+      return {
+        items: [],
+        lastKey: result.LastEvaluatedKey,
+      };
+    }
+
+    const items = result.Items as UnhydratedFeed[];
+
+    //Using [key in FeedTypes]? as unlike Record<FeedTypes, ..> it doesn't require the initial object to have all keys set
+    const byType: { [key in FeedTypes]?: string[] } = {};
```
KaelenProctor, line 58: Simplified this a bit by putting the ids into byType instead of the full Feed object. Since in the end those are used as they keys of itemsById and whose values are a BlogPost

src/dynamo/models/notice.ts
```diff
@@ -0,0 +1,123 @@
...
+      ScanIndexForward: false,
+    });
+    return {
+      items: await batchHydrate(result.Items as UnhydratedNotice[]),
+      lastKey: result.LastEvaluatedKey,
+    };
+  },
+  put: async (document: NewNotice | Notice): Promise<void> => {
+    //Explicitly generate an id if one isn't set
+    const id = document.id || uuidv4();
```
KaelenProctor, line 98: By doing this, like in some other models, it bypasses the problem of auto-generating the id.

Notification is an example where the document put doesn't have an id property so client.put saw the lack and generated an id. But as I was converting notice I originally had:
```
await client.put({
      id: document.id,
  ...
```
which failed to write to dynamo because id wasn't set. Even though client.put saw the lack of id (because the property is undefined), due to the order of the object with spread the new id wasn't added.

src/routes/dev_routes.js
```diff
@@ -1,11 +1,13 @@
 const express = require('express');
 
 const { csrfProtection, ensureRole } = require('./middleware');
-const { render, redirect } = require('../util/render');
+const { render } = require('../util/render');
```
KaelenProctor, line 4: redirect wasn't used, removed as a lint fix

PR #2595, opened 2025-02-05
Fix duplicate card removals that also remove unintended cards

### LINKED ISSUE #2482: Removing groups of cards when they are already marked for removal causes unrelated cards to be removed
Removing groups of cards when they are already marked for removal causes unrelated cards to be removed.

**To Reproduce**
Go to a cube list.
Click on a card (say, Lightning bolt), and click "Remove from cube"
Click on the header of the group of cards containing Lightning Bolt. (With default sorting, it should say something like "Instant (2)")
Click "Remove all from cube".
Click "Save changes" to apply those changes to the cube.
All cards in the group will now be removed, along with one additional card that was not in the group (and was not intended to be removed).

**Expected behavior**
I expected the duplicate removal to be ignored, and for all the cards in the group to be removed from the cube but no others.

 - OS: Windows 11 Home
 - Browser: Firefox 133.0

KaelenProctor:
Fixes #2482

# Problem
Bulk removal of a group does not check whether the card is already in the removal set, nor if it is in the edits/swaps. If the same card (by the cube index not card name) is in the removal set multiple times then when the backend removes the cards (which it does in decreasing index order to keep the indexes accurate) it will remove the card in question, and then whatever card was after it because that card now has the same index.

# Solution
Added to the bulk removal a check if the card already in the removal set by card index. Also check if the card is in the edits or swaps sets, and if so then remove from there as well. That aligns the bulk remove behaviour with single card remove.

## Thoughts
* Should we make the logic exact by having single remove/add/edit call the bulk versions of the action?
* There are some cases not handled such as if stage swapping card A for B and then edit A. Currently both the swap and edit are in the changelist. Do we want to define all these cases better or leave it?
* Should the frontend/backend be updated to construct Sets of the add, remove, edit, swap changes? That would (in theory) also have prevented the original problem

# Testing

## Before

Edit a card and then bulk remove the group, the edited card stays alongside it being removed:
old-edit-card-then-bulk-remove-has-duplicates

Reverting a bulk removal leaves the first card in the changest still being removed:
old-revert-bulk-removal-leaves-one

Delete a card and then bulk remove the group, the card is listed twice in the removes:
old-delete-card-then-bulk-remove-has-duplicates

### Original case reproduction
Here I have a cube with 3 red cards in the list and after those cards are Railway Brawler and then Fabled Passage.
cube-cards-order

When I delete lightning bolt and then the red cards group and save, you can see that Railway Brawler also was removed.
old-removed-with-duplicate-also-removed-railway-brawler

Then after a page reload can see the cards list again without Railway Brawler:
cube-cards-order-after-reload

## After

Edit a card and then bulk remove the group, edit is replaced by remove:
new-edit-card-then-bulk-remove-no-duplicates

Reverting a bulk removal leaves clean slate:
new-bulk-remove-revert-clean-slate

Delete a card and then bulk remove the group, the card is listed once:
new-delete-card-then-bulk-remove-no-duplicates

### DIFF
--- a/src/client/contexts/CubeContext.tsx
+++ b/src/client/contexts/CubeContext.tsx
@@ -848,10 +848,16 @@ export function CubeContextProvider({
         if (!newChanges[remove.board]) {
           newChanges[remove.board] = { adds: [], removes: [], swaps: [], edits: [] };
         }
-        newChanges[remove.board]?.removes?.push({
-          index: remove.index,
-          oldCard: cube.cards[remove.board][remove.index],
-        });
+
+        const removes = newChanges[remove.board]?.removes || [];
+        const removeIndex = removes.findIndex((e) => e.index === remove.index);
+        //Don't add the same card to the removals list
+        if (removeIndex === -1 && removes) {
+          newChanges[remove.board]?.removes?.push({
+            index: remove.index,
+            oldCard: cube.cards[remove.board][remove.index],
+          });
+        }
       }
 
       setChanges(newChanges);
--- a/src/client/contexts/CubeContext.tsx
+++ b/src/client/contexts/CubeContext.tsx
@@ -16,7 +16,7 @@ import { Object } from 'core-js';
 import { cardName, normalizeName } from 'utils/cardutil';
 import { deepCopy } from 'utils/Util';
 
-import Card, { BoardType, boardTypes, Changes } from '../../datatypes/Card';
+import Card, { BoardType, boardTypes, Changes, CubeCardChange } from '../../datatypes/Card';
 import { CardDetails } from '../../datatypes/Card';
 import Cube, { TagColor } from '../../datatypes/Cube';
 import { UncontrolledAlertProps } from '../components/base/Alert';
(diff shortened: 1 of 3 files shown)


src/client/contexts/CubeContext.tsx
```diff
@@ -813,10 +829,7 @@ export function CubeContextProvider({
       const newChanges = deepCopy(changes);
 
       for (const edit of list) {
-        const editIndex = newChanges[edit.board]?.edits?.findIndex((e) => e.index === edit.index) || -1;
```
KaelenProctor, line 816: The reason why reverting bulk removes left one item is because findIndex returned zero but since that is falsey the `|| -1` occurs and thus it doesn't splice the array

src/datatypes/Card.ts
```diff
@@ -144,11 +144,16 @@ export type FilterValues = {
 export const boardTypes = ['mainboard', 'maybeboard'] as const;
 export type BoardType = (typeof boardTypes)[number];
 
+export type CubeCardChange = { index: number; oldCard: Card };
```
KaelenProctor, line 147: This probably isn't necessary with the end state of removeCardByIndexFromChangeset. Originally I had it returning the modified array because it used filter() instead of findIndex + splice. But typescript didn't like the filter() syntax because the various add/edits/removes/swaps arrays are documented as possibly undefined.

src/client/contexts/CubeContext.tsx
```diff
@@ -831,19 +844,27 @@ export function CubeContextProvider({
...
         }
-        newChanges[remove.board]?.removes?.push({
-          index: remove.index,
-          oldCard: cube.cards[remove.board][remove.index],
-        });
+
+        const removes = newChanges[remove.board]?.removes || [];
+        const removeIndex = removes.findIndex((e) => e.index === remove.index);
+        //Don't add the same card to the removals list
+        if (removeIndex === -1 && removes) {
```
KaelenProctor, line 862: Here is the new check that the card being bulk removed isn't already in the removal set

PR #2602, opened 2025-02-09
Allow card search with apostrophe

KaelenProctor:
This allows name searching, without needing quotes, for names like:

* Jace, Vryn's Prodigy

* Urza's Bauble

For the details of the grammar I referenced https://github.com/dekkerglen/CubeCobra/pull/1647. From that it looks pretty safe to me to remove the apostrophe from the last part of noQuoteStringValue which is trying to match all the remaining text in the token, after and/or are past. Unlike my previous PR https://github.com/dekkerglen/CubeCobra/pull/2580 it is NOT possible to remove the colon from the exclusions, as that would result in ambiguity with `<filter>:...` filters.

**Without a ton more testing I cannot guarantee that this didn't break other filters.** The changed noQuoteStringValue is used in other filters though they are not likely to contain anything with a single quote.

# Testing
## Before
Searching using double or single quoted

Searching in cube:

image
image
image

Card search
image

## After
Now searching for cards with apostrophe works without quotes.

image
image
image

image
image

### How I gathered unique characters for testing
I gathered all the unique characters in card names using:

```
jq '.[]' cubecobra/private/names.json -r | awk '{for(i=1;i<=NF;i++)if(!a[$i]++)print $i}' FS=""  | sort
```
Then using scryfall I grabbed one or two cards whose name contains non-alphabetic characters giving me the names used in the test file.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -44,7 +44,7 @@ jobs:
         uses: tj-actions/eslint-changed-files@v25
         with:
           config_path: 'eslint.config.mjs'
-          extra_args: '--max-warnings=0'
+          extra_args: '--max-warnings=0 --no-warn-ignored'
           reporter: github-pr-review
 
       - name: Run Prettier on changed files
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -8,7 +8,7 @@ import tseslint from 'typescript-eslint';
 
 export default [
   {
-    ignores: ['dist/**/*', '.git/**/*', 'node_modules/**/*', 'src/generated/**/*', 'jobs/archived/*.js'],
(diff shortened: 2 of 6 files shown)


dekkerglen:
Tests are failing

KaelenProctor:
> Tests are failing

The linting is failing, specially on the nearley generated src/client/generated/filtering/cardFilters.js file. It looks like the eslint configured ignores are out of date, as it has src/generated but not src/client/generated. I'll update.

I've been playing around with having nearley write the cardFilters as typescript and that solves most of the lint issues, though some ignores are still needed.

The tests are passing:
```
Test Suites: 10 passed, 10 total
Tests:       119 passed, 119 total
Snapshots:   0 total
Time:        99.408 s
Ran all test suites in 2 projects.
```

KaelenProctor:
@dekkerglen All passing now!

PR #2609, opened 2025-02-13
Change tag filter to be partial matching instead of exact

KaelenProctor:
# Problem
The tag filter with a string value is doing exact matching, which is less useful that a partial match and if you compare it to name filtering the `:` operation is for partial not exact matching.

Because the tag filter supports both strings (the tag value matches the string) and numbers (the count of tags) it is not really possible to support both `=` and `:` for strings, as that would cause ambiguity in the parser given that all numbers are a subset of a string.

# Solution
Update setElementOperation so that the `:` operator does partial matching of the set element value.

Though setElementOperation is also used in the various legality filters (legality, banned, restricted and variants), the change for `:` to partial match doesn't affect it because the grammar defines all the possible formats as full strings. Thus even though setElementOperation will be doing a partial match, its equivalent to an exact match given that only the exact format strings are allowed.

# Testing

## Before
It can match tags that are a single word (no spaces) without quotes:
old-tag-string-exact-match-single-word

But if the tag has a space, double quotes are required (otherwise the filter used is tag:word AND name contains word2)
old-tag-string-quoted-exact-match-necessary-for-spaces

And per the above examples, Tag filter string only does exact matching so "Spell" doesn't match the tags with "Spells matter"
old-tag-string-only-exact-match

## After
Now tag does partial match:
new-tag-string-partial-matching

Even with quotes for multiple words:
new-tag-string-partial-matching-with-qoutes

Numeric matching still works:
new-tag-numeric-matching-still-works2
new-tag-numeric-matching-still-works

### DIFF
--- a/src/client/filtering/FuncOperations.js
+++ b/src/client/filtering/FuncOperations.js
@@ -283,6 +283,7 @@ export const setCountOperation = (op, value) => {
 export const setElementOperation = (op, value) => {
   switch (op.toString()) {
     case ':':
+      return (fieldValue) => fieldValue?.some((elem) => elem?.toLowerCase().includes(value));
     case '=':
       return (fieldValue) => fieldValue?.some((elem) => elem?.toLowerCase() === value);
     case '<>':
--- a/src/client/filtering/FuncOperations.js
+++ b/src/client/filtering/FuncOperations.js
@@ -183,6 +183,7 @@ const canCastWith = (mana, cost) => {
   cost = [...cost]
     .filter((symbol) => symbol[0] === 'x' || symbol[0] === 'y' || symbol[0] === 'z')
     .map((symbol) => {
+      // eslint-disable-next-line no-console -- Debugging
       console.debug(symbol);
       if (symbol.length === 1) {
         const intValue = parseInt(symbol[0], 10);
@@ -210,6 +211,7 @@ const canCastWith = (mana, cost) => {
         intValue = 1;
       }
     }
+    // eslint-disable-next-line no-console -- Debugging
     console.debug(intValue);
     if (Number.isInteger(intValue)) {
       const remove = [];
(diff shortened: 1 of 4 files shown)


haganbmj:
I'll raise the counterpoint to this PR in that partial matches renders me unable to filter for certain tags.

I use the tags `signed` and `unsigned`, which now means that `tag:signed` returns both and I need to construct a set of exclusions instead (eg. `tag:signed -tag:unsigned`) because there isn't an option for performing exact matches.

KaelenProctor:
> I'll raise the counterpoint to this PR in that partial matches renders me unable to (individually) filter for certain tags.
> 
> I use the tags `signed` and `unsigned`, which now means that `tag:signed` returns both and I need to construct a set of exclusions instead (eg. `tag:signed -tag:unsigned`) because there isn't an option for performing exact matches.

Thanks for the feedback. To support both cases my initial thought is to change `tags:"<phrase>"` back to exact match, as that also aligns better with how name filtering works. Per the examples on https://cubecobra.com/filters:
* `goblin blood` = cards whose names contain both "blood" and "goblin".
* `"goblin blood"` = cards whose names contain exactly "goblin blood".

@dekkerglen Does that sound reasonable to you, or any other ideas?

haganbmj:
@KaelenProctor 
I'm bad at keeping up with things on Github, but yeah I think you're right that consistent with the other filter methods is probably preferred. Notes and Card Type might be the closest comparisons and those use colon for partial match, equals sign for exact match. Maybe `tag=<phrase>` as an exact match is a way to better align those.
Quotes has its own baggage because you are allowed to have multi-word tags so you can't really use that to differentiate behavior.

KaelenProctor:
@haganbmj Circling back to let you know we made adjustments based on your feedback. Due to technical limitations I went with a solution where `tag:"signed"` does exact match (eg surrounded with quotes) vs `tag:signed` does partial match. The updates are described on the filter syntax page.

PR #2611, opened 2025-02-14
Reapply draftmancer deck picks breakdown fix

KaelenProctor:
When validating the recent bug fixes in the most recent deployment I noticed that the deck picks for Draftmancer drafts did not get fixed. Found that my change was not fully incorporated into a change to show the draft bot predictions.

# Testing

## Before
Always showing the first pack contents and the bolding of the selected card is also on the first pack in the left nav
old-deck-picks-draftmancer-always-first-pack

## After
Back to correctly showing different packs / bolding the current pack card.
new-deck-picks-draftmancer-draft

### DIFF
--- a/src/client/components/DecksPickBreakdown.tsx
+++ b/src/client/components/DecksPickBreakdown.tsx
@@ -167,8 +167,8 @@ const CubeBreakdown: React.FC<BreakdownProps> = ({ draft, seatNumber, pickNumber
   const onPickClick = (packIndex: number, pickIndex: number) => {
     let picks = 0;
     for (let i = 0; i < packIndex; i++) {
-      if (draft.InitialState?.[0]?.[i]?.cards?.length) {
-        picks += draft.InitialState[0][i].cards.length;
+      if (picksList[i]?.length) {
+        picks += picksList[i].length;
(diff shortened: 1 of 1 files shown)


src/client/components/DecksPickBreakdown.tsx
```diff
@@ -167,8 +171,8 @@ const CubeBreakdown: React.FC<BreakdownProps> = ({ draft, seatNumber, pickNumber
   const onPickClick = (packIndex: number, pickIndex: number) => {
     let picks = 0;
     for (let i = 0; i < packIndex; i++) {
-      if (draft.InitialState?.[0]?.[i]?.cards?.length) {
-        picks += draft.InitialState[0][i].cards.length;
+      if (picksList[i]?.length) {
+        picks += picksList[i].length;
```
KaelenProctor, line 175: Only functional change (the rest is prettier). Since draft.InitialState only exists for Cube Cobra drafts it wasn't adding to the index for Draftmancer packs. In the refactored code both types of drafts have picksList as the array of packs the current drafter did, so equivalent to `draft.InitialState?.[0]?.[i]`

PR #2616, opened 2025-02-17
Allow users to set their default printing preference and use in card searches/etc

KaelenProctor:
The user's default printing setting applies to searches outside of cubes. A card added to a cube is a specific printing, so even when drafting it uses that print rather than an equivalent print of the card that matches the user's preference. **It is likely possible to apply the user's preference but would be a lot more work**

### Taking suggestions on a better type name than PrintFilter =)
Has been updated to PrintingPreference

# Changes
1. Add defaultPrinting setting to the User object. If that isn't set, it defaults to recent at hydration because that is what the code currently does. Unless I am mistaken, a Cube also always has a default printing set
2. Add setting default printing in the User's display preferences
3. Pass either the user's or cube's default setting into all the methods that get a reasonable version of a card by name or filter. This should make all the default parameter values of the most reasonable methods never be used, due to #1
4. Removed a couple unused files/methods that related to printing/most reasonable
5. Added server-side validation for user's display preferences so invalid themes or printing aren't saved

# Testing

# Before
Since the default printing setting for the get most reasonable methods is recent, searching for cards such as with `set:dft` gives a bunch of yellow cards.

As seen in production:
image

# After

Searching DFT when user's default printing is "First":
new-user-first-printing-search

Searching DFT when user's default printing is "Recent" (or unset):
new-card-search-recent-printing

Searching for DFT cards when the cube's printing is recent but user's is first, the cube setting is used:
image
new-cube-printing-setting-takes-precedence1
new-cube-printing-setting-takes-precedence2
When the user searches the same card outside the cube, sees the first printing as desired
new-cube-printing-setting-takes-precedence3

Searching for DFT cards when the cube's printing is first but user's is recent, the cube setting is used:
cube-first-printing
new-cube-first-printing-takes-precdence

Server side validation for display preferences:
new-user-default-printing-validation
new-user-theme-validation


src/routes/cube/api.js
```diff
@@ -327,23 +327,6 @@ router.post(
   }),
 );
 
-router.get(
```
KaelenProctor, line 330: Found no usages of "getimage" (whole world) or "cube/getimage". Found when backtracking from getMostReasonable usages

src/datatypes/Card.ts
```diff
@@ -167,6 +167,12 @@ export interface Changes {
 export const BasicLands = ['Plains', 'Island', 'Swamp', 'Mountain', 'Forest', 'Waste'] as const;
 export type BasicLand = (typeof BasicLands)[number];
 
+export enum PrintFilter {
```
KaelenProctor, line 170: The more I see PrintFilter the less I like it =(

src/routes/search_routes.js
```diff
@@ -297,7 +299,7 @@ const searchCubes = async (query, order, lastKey, ascending, user) => {
   // separate query into tokens, respecting quotes
   const tokens = tokenize(query);
 
-  const cardQueries = getCardQueries(tokens);
+  const cardQueries = getCardQueries(tokens, user?.defaultPrinting);
```
KaelenProctor, line 302: Even though we hydrate defaultPrinting in User, since search and tools can be used anonymously we still must use `?.` to handle no user

src/routes/tools_routes.js
```diff
@@ -267,7 +295,7 @@ router.get('/cardimage/:id', async (req, res) => {
   try {
     let { id } = req.params;
 
-    const defaultPrinting = req?.query?.defaultPrinting;
+    const defaultPrinting = req?.query?.defaultPrinting || req?.user?.defaultPrinting;
```
KaelenProctor, line 298: In previous fix we pass defaultPrinting from the frontend to the cardimage endpoint when inside the cube context. But cardimage is also used outside of cubes, so adding the user setting as the next fallback

src/datatypes/Card.ts
Eetai: ArtFilter?

Eetai: I guess that's technically less descriptive than Printing since multiple printings can share the same art...

Eetai: CardPrintingFilter?

Eetai: CardPrintingFilter?

KaelenProctor: Hmm, maybe DefaultPrintingPreference or CardPrintingPreference?

Eetai: I like DefaultPrintingPreference, that gets the idea across clearly

PR #2617, opened 2025-02-17
Follow up to #2549, pull the promo_type from Scryfall onto Cards.

KaelenProctor:
This supports promo_type being used to filter variants during "reasonable cards" search, eliminating things like alternate frames, booster fun, etc.

cc @DankTrainTom 

Works with https://github.com/dekkerglen/CubeCobra/pull/2549

With https://github.com/dekkerglen/CubeCobra/pull/2616 that gives User's the choice of most recent or first printing, we could add more allowed values such that someone can want to see promos or at least alternative frames. For example Eldraine adventure cards have the scrollwork frame are tracked as promos because they have "showcase" frame_effects, but I could see someone wanting to use that printing yet exclude literal promos. The best example I can come up with is Brazen Borrower, whose showcase frame is https://scryfall.com/card/eld/281/brazen-borrower-petty-theft but then has later promos like https://scryfall.com/card/spg/30/brazen-borrower-petty-theft (special guest) or secret lairs. But of course that is a topic for another day, just adding as context.

## Edit
Added unit tests for reasonableCard and getMostReasonable. The latter required me to extract the catalog loading from carddb.ts to a separate file, in order to be able to mock the catalog with Jest. Tthough https://jestjs.io/docs/mock-functions#mocking-partials worked to mock the catalog as imported to the test, it did not allow changing the catalog as it existed within the carddb file.

For additional regression testing I re-ran the update cards function and re-tested the card search, using debugging to validate the catalog as imported into carddb has contents even if coming from cardCatalog and async filled in.

# Testing

## Before
Most recent printing includes Tamiyo promo (https://scryfall.com/card/mh3/443/tamiyo-inquisitive-student-tamiyo-seasoned-scholar) which is not considered a literal promo (card.promo === false) but does have promo_types (bundle and boosterfun)
old-tamiyo-most-recent-includes-promo-type

## After
Those promo like variants are filtered and then the original printing is returned:
new-tamiyo-most-recent-filters-promos

### DIFF
--- a/scripts/clean_followers.js
+++ b/scripts/clean_followers.js
@@ -1,6 +1,6 @@
 const _ = require('lodash');
 const Cube = require('../src/dynamo/models/cube');
-const { initializeCardDb } = require('../src/util/carddb');
+const { initializeCardDb } = require('../src/util/cardCatalog');
 
 (async () => {
   try {
--- a/scripts/rebuild_cube.js
+++ b/scripts/rebuild_cube.js
@@ -1,6 +1,6 @@
 const cubeId = '63ddcf7a75838178427ea858';
 
-const { initializeCardDb } = require('../src/util/carddb');
+const { initializeCardDb } = require('../src/util/cardCatalog');
 
 const ChangeLog = require('../src/dynamo/models/changelog');
 const Cube = require('../src/dynamo/models/cube');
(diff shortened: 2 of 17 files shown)


dekkerglen:
Can you fix the conflict?

KaelenProctor:
Conflicting with my own changes, *shakes head*. Fixed and regressioned the npm tests + searching cards

KaelenProctor:
@dekkerglen Conflicts resolved

PR #2620, opened 2025-02-20
Use user or cube printing preference in more places cards are shown (ML stuff)

KaelenProctor:
# Changes
1. Related cards on the tool/card page now will use the user's printing preference (eg drafted with, cubed with, synergy)
2. Analysis page adds/cuts will now use the cube's printing preference
3. Change the default printing preference to first per Dekkaru's ask. This applies to new users and user's that haven't yet saved their preference

**If you can think of other places cards are shown based on name or oracle (and thus we can choose any printing) let me know and I can investigate**

# Testing

## Before
On tool/card page showing related:
old-first-printing-preference-not-conformed-in-cubed-with
old-first-printing-preferenc-not-conformed-in-related

In add/cuts recommender:
old-recommender-neither-first-or-most-recent-printing
The Lunarch veteran is not a promo but it does have special frame effects. However update_cards.ts doesn't look for those effects, only extendedart and showcase. But either way it isn't the first or most recent printing
old-recommends-can-include-promos-that-arent-reasonable

### First printing
Related cards:
new-first-printing-preference-for-related
new-first-printing-prefernced-for-related-cubed-with
new-first-printing-prefernced-for-related-synergy

Add/cuts
new-first-printing-recommended-adds
new-first-printing-recommended-cuts

### Most recent printing
Related cards:
new-most-recent-printing-prefernced-for-related
new-most-recent-printing-prefernced-for-related-cubed-with
new-most-recent-printing-prefernced-for-related-synergy

Add/cuts
new-most-recent-printing-recommended-adds
new-most-recent-printing-recommended-cuts

### DIFF
--- a/src/util/carddb.ts
+++ b/src/util/carddb.ts
@@ -191,6 +191,14 @@ export function getReasonableCardByOracle(oracleId: string): CardDetails {
   return getFirstReasonable(ids);
 }
 
+export function getReasonableCardByOracleWithPrintingPreference(
+  oracleId: string,
+  printingPreference: PrintingPreference,
+): CardDetails {
+  const ids = catalog.oracleToId[oracleId];
+  return getMostReasonableByPrintingPreference(ids, printingPreference)!;
+}
+
 export function isOracleBasic(oracleId: string): boolean {
   return cardFromId(catalog.oracleToId[oracleId][0]).type.includes('Basic');
 }
--- a/src/util/carddb.ts
+++ b/src/util/carddb.ts
@@ -111,13 +111,21 @@ export function getMostReasonable(
   printing: PrintingPreference = PrintingPreference.RECENT,
   filter?: FilterFunction,
 ): CardDetails | null {
-  let ids = getIdsFromName(cardName);
+  const ids = getIdsFromName(cardName);
   if (ids === undefined || ids.length === 0) {
     // Try getting it by ID in case this is an ID.
 
     return getMostReasonableById(cardName, printing);
   }
 
+  return getMostReasonableByPrintingPreference(ids, printing, filter);
+}
+
+export function getMostReasonableByPrintingPreference(
+  ids: string[],
+  printingPreference: PrintingPreference = PrintingPreference.RECENT,
+  filter?: FilterFunction,
+): CardDetails | null {
   if (filter) {
     ids = ids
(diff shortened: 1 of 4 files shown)


src/util/carddb.ts
```diff
@@ -111,13 +111,21 @@ export function getMostReasonable(
...
   if (ids === undefined || ids.length === 0) {
     // Try getting it by ID in case this is an ID.
 
     return getMostReasonableById(cardName, printing);
   }
 
+  return getMostReasonableByPrintingPreference(ids, printing, filter);
+}
+
+export function getMostReasonableByPrintingPreference(
```
KaelenProctor, line 124: In my experience I've found optional parameters can get out of hand quickly, whereas alternate versions of the function with descriptive name differences is easier. Unless you get to having so many parameters that it makes sense to define a config object type just for it.

But if you'd prefer consolidating such that printing preference could be set or not, I can refactor.

src/util/carddb.ts
```diff
@@ -147,7 +155,7 @@ export function getMostReasonable(
   ids = cards.map((card) => card.details.scryfall_id);
```
KaelenProctor, line 155: Since we have the cards + details here, we could reverse that array rather than creating the id map and returning that. That would reduce the calls to cardFromId (at end of this method and within reasonableId called per id). Though I dunno which would be more efficient, reversing (in place) an array of "big" objects or fewer id lookups on the dictionaries.

src/util/carddb.ts
```diff
@@ -217,37 +246,45 @@ export function getRelatedCards(oracleId: string): Record<string, Record<string,
...
+      creatures: related.synergistic.creatures.map(mapper),
+      spells: related.synergistic.spells.map(mapper),
+      other: related.synergistic.other.map(mapper),
     },
   };
 }
 
 // if the oracle id is not in the training data, we will use the similar card in the metadata dict
-export function getOracleForMl(oracleId: string): string {
+export function getOracleForMl(oracleId: string, printingPreference: PrintingPreference | null): string {
```
KaelenProctor, line 276: getOracleForMl is called by drafting functions and there they don't need printing preference. After determining the oracle id/rating, usually there will be sorting or filtering against a card set which are already specific printings

PR #2621, opened 2025-02-21
Keep cube overview changes around in local storage so that validation errors don't lose your work

KaelenProctor:
# Solution
As Dekkaru described using local storage makes sense and it's a pattern used in many places.

Because the overview save is a full page reload, there is no good way to clear local storage on success like Dekkaru suggested. That meant I had to handle a minor discrepancy with the card count used in the modal, by taking that from the cube passed into the modal component.

Alternatively, but more work, would have been to convert the endpoint to an API with a JSON response though at that point local storage probably isn't needed (though would be nice in case you accidentally refresh the page).

I noticed via chrome dev tools that there is a lot of local storage items staying around even when likely never usable again, such as individual blog post title/body. Since browser's don't implement a least recently used strategy for local storage, eventually we might fill it up and then nothing new can be saved.

# Testing

## Before
The discord report was with an empty cube, I tested with banned words filter. Can see that after trying to save, all my changes were lost
old-cube-overview-modal-forgets-changed-on-validation-error

## After
Now the changes are saved in local storage and used the next time.
new-cube-overview-modal-keeps-changes-around-even-on-validation-error

### DIFF
--- a/src/client/components/CSRFForm.tsx
+++ b/src/client/components/CSRFForm.tsx
@@ -24,4 +24,6 @@ const CSRFForm = forwardRef<HTMLFormElement, CSRFFormProps>(({ children, method,
   );
 });
 
+CSRFForm.displayName = 'CSRFForm';
+
 export default CSRFForm;
--- a/src/client/components/modals/CubeOverviewModal.tsx
+++ b/src/client/components/modals/CubeOverviewModal.tsx
@@ -5,6 +5,7 @@ import { getCubeDescription } from 'utils/Util';
 import Cube from '../../../datatypes/Cube';
 import Image from '../../../datatypes/Image';
 import { CSRFContext } from '../../contexts/CSRFContext';
+import useLocalStorage from '../../hooks/useLocalStorage';
 import AutocompleteInput from '../base/AutocompleteInput';
 import Button from '../base/Button';
 import { Card, CardHeader } from '../base/Card';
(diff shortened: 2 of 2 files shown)


src/client/components/CSRFForm.tsx
```diff
@@ -24,4 +24,6 @@ const CSRFForm = forwardRef<HTMLFormElement, CSRFFormProps>(({ children, method,
   );
 });
 
+CSRFForm.displayName = 'CSRFForm';
```
KaelenProctor, line 27: Lint fix, giving the component a name. Nice when using react dev tools to look at the component tree

src/client/components/modals/CubeOverviewModal.tsx
```diff
@@ -48,14 +49,22 @@ const CubeOverviewModal: React.FC<CubeOverviewModalProps> = ({ isOpen, setOpen,
...
+
   const changeImage = useCallback(
     (image: string) => {
       setImagename(image);
       if (imageDict[image.toLowerCase()]) {
         setState({ ...state, imageName: image, image: imageDict[image.toLowerCase()] });
       }
     },
-    [imageDict, setState, state],
+    [imageDict, setImagename, setState, state],
```
KaelenProctor, line 67: Lint fix about full dependencies

dekkerglen:
I think if we make a change on one device, then on a second, editing the overview on the initial device will show the old content. We need some mechanism here to invalidate the locally cached content if there is a change made elsewhere

KaelenProctor:
The simplest solution for that would be more in line with your initial suggestion on clearing on save. I'd need to convert the save settings to an API endpoint instead of a page request. Would that work for you?

KaelenProctor:
Or is there a way to check the cube version against the localstorage one and discard localstorage if lower?

dekkerglen:
Using the version is tricky because the cards use that, and updating the cards would then invalidate the local storage. (and vice versa).

Moving it to an async api request could work, but we did that for a while and managing the state of the overview page became complicated

KaelenProctor:
Yeah my plan with the API would be to trigger a page reload on success. So
functionally the same behavior but with the ability to clear local storage.
Agreed don't want to deal with state syncing.

KaelenProctor:
Closing this as replaced by https://github.com/dekkerglen/CubeCobra/pull/2638. Rather than using local storage which has consistency issues per discussion, the replacement PR prevents the overview from being edited until the cube has cards and also saves using an API so if there are validation errors they are shown and the user has a chance to fix before trying to save again.

PR #2623, opened 2025-02-22
Convert first tool routes to Typescript/new router

KaelenProctor:
# Testing
Regression cases
1. Go to top cards page
2. Successfully see top cards by ELO descending
3. Filter by name
4. Change sort direction
5. Change sort

### DIFF
--- a/src/client/utils/Sort.ts
+++ b/src/client/utils/Sort.ts
@@ -213,6 +213,7 @@ export const ORDERED_SORTS: string[] = [
   'Pick Count',
   'Collector number',
 ];
+export type OrderedSortsType = (typeof ORDERED_SORTS)[number];
 
 /**
  * Pad the collector number with zeros in order for string sorting to work numerically like.
--- a/src/router/routes/api/draftmancer/publish.ts
+++ b/src/router/routes/api/draftmancer/publish.ts
@@ -1,4 +1,3 @@
-import { NextFunction, Request, Response } from 'express-serve-static-core';
 import Joi from 'joi';
 
 import { cardCmc, cardType, detailsToCard } from '../../../../client/utils/cardutil';
@@ -8,6 +7,7 @@ import DraftType, { DraftmancerLog, DraftmancerPick } from '../../../../datatype
 import type DraftSeatType from '../../../../datatypes/DraftSeat';
 import Cube from '../../../../dynamo/models/cube';
 import Draft from '../../../../dynamo/models/draft';
+import { NextFunction, Request, Response } from '../../../../types/express';
 import { cardFromId, getReasonableCardByOracle } from '../../../../util/carddb';
 import { deckbuild } from '../../../../util/draftbots';
 import { setupPicks } from '../../../../util/draftutil';
@@ -260,7 +260,6 @@ export const handler = async (req: Request, res: Response) => {
       draftId,
     });
   } catch (error) {
-    // @ts-expect-error TODO: define request type with logger
     req.logger?.error('Error publishing draft', error);
     return res.status(500).json({ error: 'Error publishing draft' });
   }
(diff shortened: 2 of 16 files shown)


src/client/utils/Sort.ts
```diff
@@ -213,6 +213,7 @@ export const ORDERED_SORTS: string[] = [
   'Pick Count',
   'Collector number',
 ];
+export type OrderedSortsType = (typeof ORDERED_SORTS)[number];
```
KaelenProctor, line 216: Useful for Joi validation on the sort parameter

src/router/routes/api/draftmancer/publish.ts
```diff
@@ -8,6 +7,7 @@ import DraftType, { DraftmancerLog, DraftmancerPick } from '../../../../datatype
 import type DraftSeatType from '../../../../datatypes/DraftSeat';
 import Cube from '../../../../dynamo/models/cube';
 import Draft from '../../../../dynamo/models/draft';
+import { NextFunction, Request, Response } from '../../../../types/express';
```
KaelenProctor, line 10: Use our extended express types which includes req.logger.error

src/router/routes/draft/finish.ts
```diff
@@ -56,6 +56,7 @@ const validateBody = (req: Request, res: Response, next: NextFunction) => {
   const { error } = FinishDraftBodySchema.validate(req.body);
   if (error) {
     res.status(400).json({ error: error.details[0].message });
+    return;
```
KaelenProctor, line 59: Luckily I copied the Joi validation from this first and then kept getting headers already sent errors, because the next() function was still being called. I checked all the other new routers and only this one didn't return on validation error

src/router/routes/tool/api/topcards.ts
```diff
@@ -0,0 +1,74 @@
...
+  s: Joi.string()
+    .valid(...validSortNames)
+    .required(),
+  d: Joi.string()
+    .valid(...SortDirections)
+    .required(),
+  p: Joi.number().integer().min(0).required(),
+}).unknown(true); // allow additional fields
+
+const validateQuery = (req: Request, res: Response, next: NextFunction) => {
```
KaelenProctor, line 23: I'm not sure how best to test this. It wouldn't be a unit test as much as an integration test, since we want to test the rules against the request for real.

src/routes/tools_routes.js
```diff
@@ -16,92 +15,23 @@ import carddb, {
...
 const generateMeta = require('../util/meta');
 const util = require('../util/util');
 const { csrfProtection } = require('./middleware');
 const { handleRouteError, render, redirect } = require('../util/render');
 
-const { PrintingPreference } = require('../datatypes/Card');
 const CardHistory = require('../dynamo/models/cardhistory');
 const Cube = require('../dynamo/models/cube');
 
+const { searchCards } = require('../util/tools');
```
KaelenProctor, line 27: Not sure if I like putting the common stuff in util. Since its used in both top cards and card search, which logically should be two different route files, it needs a common place.

tests/test-utils/transport.ts
```diff
@@ -8,6 +8,9 @@ export const createMockRequest = (partialRequest?: Partial<Request>): Request =>
     query: {},
     params: {},
     headers: {},
+    logger: {
```
KaelenProctor, line 11: Align mock request with the express request as we modify it

tests/draftmancer/api.test.ts
```diff
@@ -29,7 +30,8 @@ describe('Publish', () => {
   beforeAll(async () => {
     app = express();
     app.use(express.json());
-    app.post('/api/draftmancer/publish', ...routes[0].handler);
+    //Need to type as our RequestHandler or Typescript gets angry
+    app.post('/api/draftmancer/publish', ...(routes[0].handler as RequestHandler[]));
```
KaelenProctor, line 34: By using CubeCobra's extended express types in the handler Typescript complains the handlers here don't match with logger and flash params. Hence explicitly saying its our RequestHandler as well.
Probably sensible to have a simpler interface for express app in our tests so we don't have to copy/paste

dekkerglen (approved):

PR #2627, opened 2025-02-23
Fix card name sorting on playtest data, fix invalid dates in date added sorting (v2)

KaelenProctor:
# Fixes
1. Card name sorting on the playtest data was using a sort function that didn't work for the fact the values of the column are Card objects and not scalar values
2. Fix invalid dates appearing in sorting because the addedTmsp can be stored as a unix timestamp string or as a parsable date string like `2025-01-19T18:56:15.683Z`

# Testing

## Before

Sorting names in playtest not doing anything:
old-card-name-not-sorting-playtest-data

Examples of invalid dates for cards
old-card-invalid-date-if-parsable-string
old-card-invalid-dates

## After
Sorting names in playtest working:
new-card-name-sorting-playtest-data

Parsing the dates better so no Invalid dates
new-card-date-parsing-better
new-sort-date-added

### DIFF
--- a/src/client/analytics/PlaytestData.tsx
+++ b/src/client/analytics/PlaytestData.tsx
@@ -6,7 +6,7 @@ import { fromEntries } from 'utils/Util';
 import Card, { DefaultElo } from '../../datatypes/Card';
 import { Flexbox } from '../components/base/Layout';
 import ErrorBoundary from '../components/ErrorBoundary';
-import { compareStrings, SortableTable } from '../components/SortableTable';
+import { SortableTable } from '../components/SortableTable';
 import withAutocard from '../components/WithAutocard';
 import CubeContext from '../contexts/CubeContext';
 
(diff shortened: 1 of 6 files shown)


src/client/analytics/PlaytestData.tsx
```diff
@@ -89,7 +93,7 @@ const PlaytestData: React.FC<PlaytestDataProps> = ({ cubeAnalytics }) => {
             { key: 'mainboards', title: 'Mainboard Count', sortable: true, heading: false },
           ]}
           data={data}
-          sortFns={{ label: compareStrings }}
+          sortFns={{ card: compareCardNames }}
```
KaelenProctor, line 96: The card name column has key `card` so the sort function for label did nothing. Even if it was used, that would have compared the object as a string which would have been `[object Object]` and thus no sorting either.

PR #2634, opened 2025-02-27
Async load cube or user followers when the modal is opened, page results

KaelenProctor:
# Problem
As identified by endpoint duration logging there are outliers where a huge number of followers slows the backend down, having to fetch all the users from dynamo and send them in the response. Most noticeable is in the Modo Vintage cube which has over 3000 followers.

# Solution
As proposed by Dekkaru, load the results asynchronously on the frontend and do so using paging. That reduces the size of the initial page response/duration and keeps each fetch for the new page small.

I choose a page size of 100 as that is in line with other paging. For Modo Vintage Cube that would require 30+ clicks to see all the users, but then again I can't really see why anyone would do that.

## Other improvements
Updated imagedict loads that exist in modals to only do so when the modal is opened and if it hasn't been fetched yet. We discussed changing the caching headers for this more so that the server has to do less work, but this should make a dent in the amount of traffic already

# Testing
I created 50+ fake users by inserting directly into dynamodb, then added those users as followers to a cube and user. Scale wise nowhere close to Modo Vintage cube, but alongside reducing the page size to 10 while testing it allowed me to validate the paging.

## Before

On my local the response size for the Cube overview is 11,039,624 bytes when it contains the followers. The average response duration across 6 samples was 1,378 ms; it is not a super interesting number. The production logs will be a much better indication of the performance improvement.

Both packages and Cube overview loaded the imagedict right away, even though its only needed in the various modals. Perhaps it would make sense to have a ImageDictContext in the future? (I am still a context newbie with react)
old-packages-loads-imagedict-right-away
old-overview-loads-imagedict-twice-right-away

## After

On my local the response size for the Cube overview is 10,999,615 bytes when it only has followers count. The average response duration across 6 samples was 972 ms; again it is not very interesting.

Paged loading of followers in the modal:
new-paged-loading-of-followers2

The network requests for the paged loading:
new-paged-loading-of-followers-network

**Fetching the image dict occurs after modals are opened.**

On cube overview load no fetches:
new-fetching-imagedict-in-modals-occurs-on-open
Opening customize basics does a fetch:
new-fetching-imagedict-in-modals-occurs-on-open2
Opening edit overview does its own fetch:
new-fetching-imagedict-in-modals-occurs-on-open3

### DIFF
--- a/src/client/components/modals/CreatePackageModal.tsx
+++ b/src/client/components/modals/CreatePackageModal.tsx
@@ -22,14 +22,18 @@ const CreatePackageModal: React.FC<CreatePackageModalProps> = ({ isOpen, setOpen
   const [cardName, setCardName] = useState<string>('');
   const [packageName, setPackageName] = useState<string>('');
   const [imageDict, setImageDict] = useState<{ [key: string]: { id: string } }>({});
+  const [fetched, setFetched] = useState(false);
 
   useEffect(() => {
-    fetch('/cube/api/imagedict')
-      .then((response) => response.json())
-      .then((json) => {
-        setImageDict(json.dict);
-      });
-  }, []);
+    if (isOpen && !fetched) {
+      fetch('/cube/api/imagedict')
+        .then((response) => response.json())
+        .then((json) => {
+          setImageDict(json.dict);
+          setFetched(true);
+        });
+    }
+  }, [isOpen, fetched]);
 
   const submitCard = () => {
     if (imageDict) {
--- a/src/client/pages/UserBlogPage.tsx
+++ b/src/client/pages/UserBlogPage.tsx
@@ -14,7 +14,7 @@ import UserLayout from 'layouts/UserLayout';
 
 interface UserBlogPageProps {
   owner: User;
-  followers: User[];
+  followersCount: number;
   following: boolean;
   posts: BlogPostType[];
   loginCallback?: string;
@@ -24,7 +24,7 @@ interface UserBlogPageProps {
 const PAGE_SIZE = 10;
 
 const UserBlogPage: React.FC<UserBlogPageProps> = ({
-  followers,
+  followersCount,
   following,
   posts,
   owner,
@@ -64,7 +64,7 @@ const UserBlogPage: React.FC<UserBlogPageProps> = ({
 
       setLoading(false);
     }
-  }, [owner.id, currentLastKey, items, page]);
+  }, [csrfFetch, owner.id, currentLastKey, items, page]);
 
   const pager = (
     <Pagination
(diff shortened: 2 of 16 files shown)


src/client/components/modals/CreatePackageModal.tsx
```diff
@@ -22,14 +22,18 @@ const CreatePackageModal: React.FC<CreatePackageModalProps> = ({ isOpen, setOpen
...
+  const [fetched, setFetched] = useState(false);
 
   useEffect(() => {
-    fetch('/cube/api/imagedict')
-      .then((response) => response.json())
-      .then((json) => {
-        setImageDict(json.dict);
-      });
-  }, []);
+    if (isOpen && !fetched) {
```
KaelenProctor, line 28: Here we could check if imageDict is empty or not instead of having explicit fetched. I implemented this first for the loading of followers and since a cube/user may have no followers, I needed an explicit state to check if loaded as I could not depend on data size. So to be consistent used the same fetched state in these modals.

src/client/pages/UserBlogPage.tsx
```diff
@@ -64,14 +64,15 @@ const UserBlogPage: React.FC<UserBlogPageProps> = ({
 
       setLoading(false);
     }
-  }, [owner.id, currentLastKey, items, page]);
+  }, [csrfFetch, owner.id, currentLastKey, items, page]);
```
KaelenProctor, line 67: Lint fix for dependencies

src/datatypes/User.ts
```diff
@@ -17,8 +17,8 @@ export default interface User {
   about?: string;
   hideTagColors?: boolean;
   followedCubes?: string[];
-  followedUsers?: string[];
-  following?: string[];
+  followedUsers?: string[]; //Who this user is following
```
KaelenProctor, line 20: Comments mainly because I confused myself when looking at the dynamo results directly lol

tests/followers.api.test.ts
```diff
@@ -0,0 +1,141 @@
...
+import CubeFn from '../src/util/cubefn';
+import { createCube, createUser } from './test-utils/data';
+import { expectRegisteredRoutes } from './test-utils/route';
+import { call, middleware } from './test-utils/transport';
+
+jest.mock('../src/dynamo/models/user');
+jest.mock('../src/dynamo/models/cube');
+jest.mock('../src/util/cubefn');
+
+describe('Followers API', () => {
```
KaelenProctor, line 13: Gotta hand it to Copilot, it did a very good job generating these tests when I told it to follow the style of an existing api test file we have. I only had to add the helper createXYZ methods for more accurate objects and fix `.withUser()` -> `.as()` in the call stuff

PR #2637, opened 2025-02-28
feat: Sort by word count

### LINKED ISSUE #2636: Feature request: Sort by word count
I would like to be able to sort my cube by the word count of the oracle text.

I would use this sort option to identify cards that are complexity outliers.

**Additional context**
N/A

oscarhermoso:
This PR adds an additional sort option for "Word Count".

Would appreciate it if someone could check out my branch & confirm the feature is working as expected. Thanks!

Closes #2636

### DIFF
--- a/src/client/utils/Sort.ts
+++ b/src/client/utils/Sort.ts
@@ -20,6 +20,7 @@ import {
   cardStatus,
   cardTix,
   cardType,
+  cardWordCount,
   COLOR_COMBINATIONS,
   convertFromLegacyCardColorCategory,
 } from './cardutil';
@@ -212,6 +213,7 @@ export const ORDERED_SORTS: string[] = [
   'Cube Count',
   'Pick Count',
   'Collector number',
+  'Word Count',
 ];
 export type OrderedSortsType = (typeof ORDERED_SORTS)[number];
 
@@ -279,6 +281,7 @@ export const SortFunctions: Record<string, (a: any, b: any) => number> = {
     }
     return 0;
   },
+  'Word Count': (a, b) => cardWordCount(a) - cardWordCount(b),
 };
 
 export const SortFunctionsOnDetails = (sort: string) => (a: any, b: any) =>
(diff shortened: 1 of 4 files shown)


src/client/utils/cardutil.ts
```diff
@@ -466,6 +466,52 @@ export const cardLayout = (card: Card): string => card.details?.layout ?? '';
...
+    const byteAtIndex = BITMAP[byteIndex];
+    const isMatch = ((byteAtIndex >> bitIndex) & 1) === 1;
+    if (isMatch && (shouldCount || byteAtIndex === 255)) count++;
+    shouldCount = !isMatch;
+  }
+
+  if (shouldCount) count++;
+
+  return count;
+}
```
dekkerglen, line 503: I think this is a bit overkill - `oracle_text.split(' ').length` should be sufficient I think

src/datatypes/Card.ts
```diff
@@ -75,6 +75,7 @@ export interface CardDetails {
   cubeCount?: number;
   pickCount?: number;
   isExtra?: boolean;
+  wordCount?: number;
```
KaelenProctor, line 78: With the addition to the type it would be good to add to the logic within update_cards.ts convertCard() function. Otherwise the field will be undefined unless the card has passed through the word count sort.

src/client/utils/Sort.ts
```diff
@@ -523,6 +540,17 @@ export function getLabelsRaw(cube: Card[] | null, sort: string, showOther: boole
...
     }
     ret = res;
+  } else if (sort === 'Word Count') {
+    const labels: string[] = [];
+    for (const card of cube || []) {
+      const wordCount = cardWordCount(card);
+      const label = wordCountBucket(wordCount);
+      if (!labels.includes(label)) {
+        labels.push(label);
+        if (labels.length === wordCountBuckets.length) break;
```
KaelenProctor, line 550: small nit: for style consistency would be nice to have {} brackets with the if and break

KaelenProctor (commented):
Looking good!

PR #2639, opened 2025-03-01
Allow custom back of card images to make the Transform button appear.

KaelenProctor:
Requires the display setting of show custom images.

# Testing

## Before
If the card doesn't have a back image by nature of it being a DFC/transforming card, adding a custom back image doesn't allow transforming.
old-adding-custom-back-image-on-single-faced-card-doesnt-transform

## After

Still works with cards that already are double faced:
custom-images-with-already-double-faced-card

Can add a custom back image to a single-faced card and then it can transform. And if you are looking at the custom back image then remove it, the UI reverts to the front face
can-transform-when-custom-images-are-used

If the custom back image is removed and the card normally has a back image, it does not revert to the front face
removing-custom-back-image-for-double-faced-stays-on-back

If "show custom images" is off, then cannot transform to custom back side:
no-transform-if-hide-custom-images

### DIFF
--- a/src/client/components/card/CardModal.tsx
+++ b/src/client/components/card/CardModal.tsx
@@ -131,8 +131,8 @@ const CardModal: React.FC<CardModalProps> = ({
 
   const updateField = useCallback(
     (field: keyof Card, value: any) => {
-      //Handle edge case of the back image being removed while looking at the back
-      if (field === 'imgBackUrl' && value.trim() === '' && !isFrontImage) {
+      //Handle edge case of the back image being removed while looking at the back, and there is no actual back image on the card details
+      if (field === 'imgBackUrl' && value.trim() === '' && !card?.details?.image_flip && !isFrontImage) {
         setIsFrontImage(true);
         setImageUsed(getCardFrontImage(card));
       }
--- a/src/client/components/card/CardModal.tsx
+++ b/src/client/components/card/CardModal.tsx
@@ -99,20 +99,16 @@ const CardModal: React.FC<CardModalProps> = ({
     }
   }, [card, versionDict]);
 
-  const updateField = useCallback(
-    (field: keyof Card, value: any) => {
-      editCard(card.index!, { ...card, [field]: value }, card.board!);
-    },
-    [card, editCard],
-  );
-
   const disabled = !canEdit || card.markedForDelete;
 
   const { showCustomImages } = useContext(DisplayContext);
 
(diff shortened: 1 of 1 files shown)


src/client/components/card/CardModal.tsx
```diff
@@ -133,6 +129,18 @@ const CardModal: React.FC<CardModalProps> = ({
     setImageUsed(getCardBackImage(card));
   }
 
+  const updateField = useCallback(
```
KaelenProctor, line 132: Moved updateField down so that all the functions it calls are defined above it

src/client/components/card/CardModal.tsx
```diff
@@ -99,20 +99,16 @@ const CardModal: React.FC<CardModalProps> = ({
...
-  );
-
   const disabled = !canEdit || card.markedForDelete;
 
   const { showCustomImages } = useContext(DisplayContext);
 
-  const getCardFrontImage = function (card: Card) {
-    return (showCustomImages && cardImageUrl(card)) || card?.details?.image_normal;
-  };
+  const getCardFrontImage = useCallback(
```
KaelenProctor, line 106: Converted to useCallback now that it gets used within a hook

src/client/components/card/CardModal.tsx
```diff
@@ -133,6 +129,18 @@ const CardModal: React.FC<CardModalProps> = ({
     setImageUsed(getCardBackImage(card));
   }
 
+  const updateField = useCallback(
+    (field: keyof Card, value: any) => {
+      //Handle edge case of the back image being removed while looking at the back
+      if (field === 'imgBackUrl' && value.trim() === '' && !isFrontImage) {
```
KaelenProctor: Without this you could look at the custom back image, remove it, then see the MTG card back with no way to see the front face. That applies to a card which normally has only a front face, but the custom back image allowed the transform.

src/client/components/card/CardModal.tsx
```diff
@@ -151,7 +159,7 @@ const CardModal: React.FC<CardModalProps> = ({
                   fallbackSrc="/content/default_card.png"
                   alt={cardName(card)}
                 />
-                {card?.details?.image_flip && (
+                {getCardBackImage(card) && (
```
KaelenProctor, line 162: Dum dum by me. Everywhere else was using functions which ended up using functions from cardutil which handles custom images then fallback to the details images.

PR #2640, opened 2025-03-01
Add form validation to the create cube modal, leveraging built-in browser input/form validation.

KaelenProctor:
While working on the edit overview UX improvements I thought about doing the same for the create cube modal.

# Testing

## Before
No validation in the UI. Meaning it easy to have a cube name shorter than 5 characters.

## After
Min name length is highlighted when invalid / provides clear message
new-min-length-cube-name-validation

Security question field required
new-required-security-question

Captcha not being set uses Alerts. Couldn't see how to attach form validation through the Recapthca library:
new-required-captcha

Gif of the validation in total:
new-create-cube-validation-total

### DIFF
--- a/src/client/components/base/Input.tsx
+++ b/src/client/components/base/Input.tsx
@@ -21,6 +21,7 @@ export interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> 
   onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
   onEnter?: () => void;
   disabled?: boolean;
+  otherInputProps?: React.InputHTMLAttributes<HTMLInputElement>;
 }
 
 const Input: React.FC<InputProps> = ({
@@ -40,6 +41,7 @@ const Input: React.FC<InputProps> = ({
   autoCapitalize = 'sentences',
   autoCorrect = 'on',
   spellCheck = true,
+  otherInputProps = {},
 }) => {
   return (
     <div className="block w-full">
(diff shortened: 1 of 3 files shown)


src/client/components/base/Input.tsx
```diff
@@ -40,6 +41,7 @@ const Input: React.FC<InputProps> = ({
   autoCapitalize = 'sentences',
   autoCorrect = 'on',
   spellCheck = true,
+  otherInputProps = {},
```
KaelenProctor, line 44: Originally I used the react idomatic code of `...props` and `<input {...props}` but that went wonky in the forms. For example we are setting name properties on the <Input> fields but since those aren't explicit props handled here, they don't go on the actual inputs. But with `...props` they were added and then alongside the `<CSRFForm>` formdata all the fields were twice in the actual post body.

I am not a fan of otherInputProps but I couldn't see a way a good alternative that didn't involve reviewing all <Input> usages across the code for cases like that, nor did I want to add explicit props for the various <input> fields.

If there are alternatives I didn't consider, please let me know!

src/client/components/modals/CreateCubeModal.tsx
```diff
@@ -41,17 +86,26 @@ const CreateCubeModal: React.FC<Props> = ({ isOpen, setOpen }) => {
             <Input
               label="Cube name:"
               value={name}
-              onChange={(e) => setName(e.target.value)}
-              maxLength={1000}
-              name="name"
```
KaelenProctor, line 46: name isn't used by <Input> so it isn't necessary

PR #2642, opened 2025-03-04
Power/toughness comparison filters

KaelenProctor:
Feature request from Discord, to implement the same power/toughness comparison filters as in Scryfall https://scryfall.com/docs/syntax#stats. 

eg `pow>toughness`, `toughness!=power`, `toughness>=power`


jest.config.js
```diff
@@ -28,11 +28,7 @@ module.exports = {
   collectCoverage: true,
   coverageDirectory: 'coverage',
   coverageReporters: ['json', 'text'],
-  collectCoverageFrom: [
-    'src/**/*.{ts,tsx}',
-    '!src/**/*.d.ts',
-    '!src/**/index.{js,ts}',
-  ],
+  collectCoverageFrom: ['src/**/*.{ts,tsx}', 'src/client/filtering/*.js', '!src/**/*.d.ts', '!src/**/index.{js,ts}'],
```
KaelenProctor, line 31: There are other JS files we may want coverage on but for now I only added a single new directory

PR #2643, opened 2025-03-05
use batchpredict results before user picks, add draft weights to picks

Eetai:
Major Changes

**Faster Draft UI**

The batchPredict call (which fetches ML weights used for bot picks) now runs and updates the draft as the pack loads, ensuring the next pack is ready immediately instead of waiting on the server between picks.

**Retry on Failure**

Added a retry button for batchPredict failures. Previously, failures could corrupt the local draft state, as a makePick call might start but not complete. While that should no longer happen due to the sequencing changes, users still need a way to recover from a failed batchPredict that prevents them from making a next pick. The retry button addresses this.

**ML Prediction Weights**

Added a button to view ML prediction weights for the current pack, similar to the existing pick-by-pick breakdown. If users have frequent questions about the numbers, we may want to add helper text for clarification. The blue-highlighted card represents the choice a bot would make in the user’s position.

**TODOs**

Refactor CubeDraftPage— still too wild, draft logic should move to a custom hook.
Properly handle EndDraft failures
Prefetch next pack's images for even smoother UI
Add fancy draft animations

### DIFF
--- a/public/css/stylesheet.css
+++ b/public/css/stylesheet.css
@@ -786,6 +786,10 @@ video {
   top: 100%;
 }
 
+.bottom-\[5\%\] {
+  bottom: 5%;
+}
+
 .z-10 {
   z-index: 10;
 }
@@ -1150,6 +1154,10 @@ video {
 
 .w-96 {
   width: 24rem;
+}
+
+.w-\[140px\] {
+  width: 140px;
 }
 
 .w-full {
@@ -2234,6 +2242,10 @@ video {
 
 .italic {
   font-style: italic;
+}
+
+.tracking-tight {
+  letter-spacing: -0.025em;
 }
 
 .text-bg {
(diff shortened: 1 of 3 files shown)


KaelenProctor:
From testing this a few pieces of feedback on the UX:

1. I expected "Show bot ratings" to be a toggle, rather than having to click it each pack
2. Also thought it was odd there is no way to hide them after
3. Maybe just me, but I would suggest putting the "Show bot ratings" / "Waiting for bot picks" next to the "pack X, pick Y" information so it is all next to each other. Having to look at opposite ends of the header above the cards is jarring
4. The pack can show before the bot picks are done and so you can click on them, but nothing occurs. The cursor does change from pointer to grabber when its ready, but can be easily missed. So perhaps a translucent overlay on the cards while the bot pick is finishing would help indicate to people the readiness

PR #2646, opened 2025-03-08
Fix error where frontend causes a 404 on redirect because the short id changed.

KaelenProctor:
Return the direct portion of the URL to use instead of a reload.

# Testing

## Before

Save a cube without a short id, reloads fine
old-save-cube-add-short-id-continues-using-guid

Saving a cube to add a short id reloads using the GUID

Save a cube with a short id change, 404
old-save-cube-change-short-id-404

Save a cube and remove the short id, 404
old-save-cube-remove-shortid-404

## After

Save a cube without a short id, redirects to the GUID URL
new-save-cube-without-short-id-redirects-to-guid

Save a cube adding a short id, redirects to the short ID URL
new-save-cube-adding-short-id-redirects-to-shortid

Save a cube with a short id, redirects to it
new-save-cube-with-shortid-redirects-to-it

Save a cube and change the short id, redirects to the new one
new-save-cube-change-shortid-redirects-to-new

Save a cube and remove the short id, redirects to the GUID URL
new-save-cube-remove-shortid-redirects-to-guid

### DIFF
--- a/src/client/components/modals/CubeOverviewModal.tsx
+++ b/src/client/components/modals/CubeOverviewModal.tsx
@@ -75,11 +75,11 @@ const CubeOverviewModal: React.FC<CubeOverviewModalProps> = ({ isOpen, setOpen, 
       }),
     });
 
+    const body = await response.json();
     if (response.ok) {
       //Reload page to ensure state is updated
-      window.location.reload();
+      window.location.replace(body.redirect);
     } else {
-      const body = await response.json();
       setAlerts([{ color: 'danger', message: body.error }]);
     }
   }, [csrfFetch, state]);
--- a/tests/cube/api/editoverview.api.test.ts
+++ b/tests/cube/api/editoverview.api.test.ts
@@ -16,6 +16,7 @@ jest.mock('../../../src/dynamo/models/cubeHash');
 describe('Edit overview API', () => {
   afterEach(() => {
     jest.clearAllMocks();
+    jest.resetAllMocks();
   });
 
   it('should fail if the cube isnt found', async () => {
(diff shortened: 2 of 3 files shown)


src/client/components/modals/CubeOverviewModal.tsx
```diff
@@ -75,11 +75,11 @@ const CubeOverviewModal: React.FC<CubeOverviewModalProps> = ({ isOpen, setOpen,
       }),
     });
 
+    const body = await response.json();
     if (response.ok) {
       //Reload page to ensure state is updated
-      window.location.reload();
+      window.location.replace(body.redirect);
```
KaelenProctor, line 81: Are there any security implications of this? I am returning a root relative path

tests/cube/api/editoverview.api.test.ts
```diff
@@ -16,6 +16,7 @@ jest.mock('../../../src/dynamo/models/cubeHash');
 describe('Edit overview API', () => {
   afterEach(() => {
     jest.clearAllMocks();
+    jest.resetAllMocks();
```
KaelenProctor, line 19: Weirdly without both, things like mockRejectedValue were persisting between tests. Caused me lots of confusion!

tests/cube/api/editoverview.api.test.ts
```diff
@@ -193,6 +194,7 @@ describe('Edit overview API', () => {
     const owner = createUser({ id: '12345' });
     const existingCube = createCube({ shortId: 'foo', owner });
     const updatedCube = createCube({
+      ...existingCube,
```
KaelenProctor, line 197: More accurate setup with the updated cube being the existing one but with specific things updated

KaelenProctor:
Ug, the failing test is existing and already put up https://github.com/dekkerglen/CubeCobra/pull/2645 to fix

PR #2647, opened 2025-03-08
Fix Playtest data CSV containing card ID instead of name, so different than the UI

KaelenProctor:
Reported in Discord.

Most changes are utilizing cardutil functions more, for consistency. As also asked in contributor Discord, we can easily make the export CSV use the same sorting the UI is which seems like a nice quality of life improvement.

# Testing

## Before

Screenshot showing both CSV output and table in the UI, see "card" column has card oracle ids.
old-playtest-csv-card-ids-instead-of-name

## After

Again showing both, but "card" is the card name
new-playtest-csv-card-name

### DIFF
--- a/src/client/analytics/PlaytestData.tsx
+++ b/src/client/analytics/PlaytestData.tsx
@@ -1,6 +1,6 @@
 import React, { useContext, useMemo } from 'react';
 
-import { encodeName, mainboardRate, pickRate } from 'utils/cardutil';
+import { cardName, cardNameLower, cardOracleId, encodeName, mainboardRate, pickRate } from 'utils/cardutil';
 import { fromEntries } from 'utils/Util';
 
 import Card, { DefaultElo } from '../../datatypes/Card';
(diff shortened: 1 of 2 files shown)


src/client/analytics/PlaytestData.tsx
```diff
@@ -62,7 +57,7 @@ const PlaytestData: React.FC<PlaytestDataProps> = ({ cubeAnalytics }) => {
         .filter(([oracle]) => cardDict[oracle])
         .map(([oracle, { elo, mainboards, sideboards, picks, passes }]) => ({
           card: {
-            exportValue: oracle,
+            exportValue: cardName(cardDict[oracle]),
```
KaelenProctor, line 60: Functional change, use the cardName as the export value. The card is an object in order for renderCardLink to display the name and autocard hover, which means we also need the export value.

KaelenProctor:
Update the export CSV to use the same sorting as the visual table:

Examples from the various analytics tables:
new-sorted-export-csv3
new-sorted-export-csv2
new-sorted-export-csv
aaa

PR #2649, opened 2025-03-10
Update Tile component to use anchor instead of onClick so browser's normal ctrl/cmd click behaviour opens the cube in a new tab.

KaelenProctor:
# Testing

## After

Opening from dashboard or profile, using ctrl (Chrome and Firefox shown, also tested edge). It also works holding windows key (though that also opens the start menu so not advised)
new-tile-component-ctrl-click-opens-new-window
new-tile-component-ctrl-click-opens-new-tab2
Regular clicking unaffected

### DIFF
--- a/src/client/components/base/Tile.tsx
+++ b/src/client/components/base/Tile.tsx
@@ -12,10 +12,8 @@ interface TileProps {
 
 const Tile: React.FC<TileProps> = ({ className, children, href }) => {
   return (
-    <div
-      onClick={() => {
-        window.location.href = href;
-      }}
+    <a
+      href={href}
       className={classNames(
         'block bg-bg-accent shadow overflow-hidden border border-border',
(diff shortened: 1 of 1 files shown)


src/client/components/base/Tile.tsx
```diff
@@ -12,18 +12,16 @@ interface TileProps {
...
 const Tile: React.FC<TileProps> = ({ className, children, href }) => {
   return (
-    <div
-      onClick={() => {
-        window.location.href = href;
-      }}
+    <a
+      href={href}
       className={classNames(
         'block bg-bg-accent shadow overflow-hidden border border-border',
```
KaelenProctor, line 18: Since already explicitly block styled, the style is the same in the end even though anchors are normally inline. Used chrome dev tools to compare computed before and after, the computed before are all in the after

PR #2651, opened 2025-03-11
create pendingPick state and allow users to load a pick while the bot…

Eetai:
This PR updates the previous draft changes to enhance UX consistency and performance.  

## Key Changes  

- **Introduces `pendingPick` Concept in the code**
  - Allows users to make picks before bots complete their pick cycle.  
  - Pending picks are stored in state, and a `useEffect` converts them into real picks once the bot call resolves.  
  - This immediately triggers the next `fetchBatchPredict` call.  

- **UI Adjustments**  
  - Replaces the grey "Waiting for Bot Picks" button with *"Waiting for Bot Picks…"* in the top left to match the *"Waiting for next pack…"* pattern.  

- **Simplified State Management**  
  - Removed the lingering board mutation from `makePick` to simplify logic.  
  - The *mainboard/sideboard* state is now only affected by `onMoveCard`, while draft state changes occur inside `makePick`.  
  - (`makePick` might be better named something like `moveDraftForward`?)  

With these changes, users can once again pick as soon as cards appear on screen while still benefiting from the speed improvements and draft state consistency introduced in prior updates.  

## TODOs  

- Further UI improvements.  
- Start refactoring `CubeDraftPage` for style and readability once logic changes are verified.  
- Fix a bug from the last PR related to a custom draft pack ending with a "pass" step. This likely requires an additional line of logic in `makePick`.

### DIFF
--- a/src/client/components/Pack.tsx
+++ b/src/client/components/Pack.tsx
@@ -51,13 +51,13 @@ const Pack: React.FC<PackProps> = ({
             {retryInProgress ? 'Retrying...' : 'Bot picks failed. Try again?'}
           </Button>
         ) : (
-          ratings &&
-          ratings.length > 0 &&
-          !showRatings && (
-            <Button onClick={() => setShowRatings(true)} color="primary">
-              Show CubeCobra Bot Ratings
-            </Button>
-          )
+          <Button
+            className={ratings && ratings.length > 0 && !showRatings ? '' : 'invisible'}
+            onClick={() => setShowRatings(true)}
+            color="primary"
+          >
+            Show CubeCobra Bot Ratings
+          </Button>
         )}
       </CardHeader>
       <CardBody>
--- a/src/client/components/Pack.tsx
+++ b/src/client/components/Pack.tsx
@@ -26,7 +26,6 @@ interface PackProps {
 const Pack: React.FC<PackProps> = ({
   pack = [],
   loading = false,
-  loadingPredictions = false,
   title = 'Pack',
   disabled = false,
   ratings,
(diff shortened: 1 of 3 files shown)


src/client/pages/CubeDraftPage.tsx
```diff
@@ -477,22 +547,44 @@ const CubeDraftPage: React.FC<CubeDraftPageProps> = ({ cube, draft, loginCallbac
       if (target.type === locations.pack) {
         return;
       }
-
+      //Dragged a card from the pack to the deck or sideboard (the latter is off)
```
KaelenProctor, line 550: We can remove the `(the latter is off)` from the comment now, the sideboard is available during picks

src/client/components/Pack.tsx
```diff
@@ -51,10 +50,6 @@ const Pack: React.FC<PackProps> = ({
           <Button onClick={onRetry} color="danger" disabled={retryInProgress}>
             {retryInProgress ? 'Retrying...' : 'Bot picks failed. Try again?'}
           </Button>
-        ) : loadingPredictions ? (
-          <Button color="secondary" disabled>
-            Making Bot Picks...
-          </Button>
         ) : (
           ratings &&
           ratings.length > 0 &&
```
KaelenProctor, line 55: Minor UX thing but I noticed that when the button isn't showing (ratings loading or showing the ratings) the height of the header changes, as seen in gif:
header-height-shifting
My suggested solution is to move the conditional onto the button to set class invisible eg `className={ratings && ratings.length > 0 && !showRatings ? '' : 'invisible'}`. That way the button is always in the layout for height calculations even if not visible currently.

KaelenProctor (commented):
Tested it out and the drafting definitely feels snappier than before!

PR #2652, opened 2025-03-12
fix pick by pick breakdown bug that came from not tracking user pick order

Eetai:
Add new array picksInOrder to keep pick order sequenced - necessary for Pick by Pick Breakdown UI.

### DIFF
--- a/src/client/pages/CubeDraftPage.tsx
+++ b/src/client/pages/CubeDraftPage.tsx
@@ -100,8 +100,7 @@ const CubeDraftPage: React.FC<CubeDraftPageProps> = ({ cube, draft, loginCallbac
   const [sideboard, setSideboard] = useLocalStorage(`sideboard-${draft.id}`, setupPicks(1, 8));
   const [ratings, setRatings] = useState<number[]>([]);
   const [currentPredictions, setCurrentPredictions] = useState<PredictResponse | null>(null);
-
-  // Add state to track the pending pick made during predictionsLoading
+  const [picksInOrder, setPicksInOrder] = useLocalStorage<number[]>(`picks-${draft.id}`, []); // Add state to track the pending pick made during predictionsLoading
   const [pendingPick, setPendingPick] = useState<number | null>(null);
 
   // Draft Status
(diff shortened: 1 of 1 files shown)


KaelenProctor:
At this point the next thing we should look at for this area is adding tests for the logic, given we've had a few different bugs in the area.

Testing wise, have you tried custom draft formats with trash picks or auto pick/trashing?

src/client/pages/CubeDraftPage.tsx
```diff
@@ -100,8 +100,7 @@ const CubeDraftPage: React.FC<CubeDraftPageProps> = ({ cube, draft, loginCallbac
   const [sideboard, setSideboard] = useLocalStorage(`sideboard-${draft.id}`, setupPicks(1, 8));
   const [ratings, setRatings] = useState<number[]>([]);
   const [currentPredictions, setCurrentPredictions] = useState<PredictResponse | null>(null);
-
-  // Add state to track the pending pick made during predictionsLoading
+  const [picksInOrder, setPicksInOrder] = useLocalStorage<number[]>(`picks-${draft.id}`, []); // Add state to track the pending pick made during predictionsLoading
   const [pendingPick, setPendingPick] = useState<number | null>(null);
```
KaelenProctor, line 104: Given localstorage is used for picksInOrder I assume that is to help with page reloads not messing with the draft? If so, then do any other pieces of state need local storage?

Eetai: Yeah now that picksInOrder is the source of truth for the user's picks we need to make sure it's not lost on page load.  Ratings/currentPredictions will be regenerated by the backend on refresh so not necessary to store on localstorage.

I suppose setPendingPick should be localStorage for the cases where:

a. they've made a pendingPick before the last round of bot picks came in
b. then the bot picks fail and we're in error state

I'd rather not more complexity w/ this bugfix but can tweak that with the next PR

Eetai:
> At this point the next thing we should look at for this area is adding tests for the logic, given we've had a few different bugs in the area.
> 
> Testing wise, have you tried custom draft formats with trash picks or auto pick/trashing?

Yeah, I agree, I think it will be easier to build tests now the card click behavior is separated from the draft state changes.

I would like to turn the remaining makePick logic into something even easier to test, maybe it can be turned into a pure function with some more refactoring (and/or a few smaller functions) that doesn't live in CubeDraftPage.tsx. 

I have tested with basic custom drafts I made locally with auto pick/trashing, but I tried not to touch any of pick/trash/pass logic itself with my changes.

Eetai:
1. Prior version called setUserPicksInOrder too early, just a logic issue
2. There was a React race condition issue that I solved by giving endDraft its own useEffect instead of having it live in makePick. This is another step in reducing the workload and complexity of makePick.

PR #2654, opened 2025-03-13
Fix delete deck modal actions redirecting to the deck

KaelenProctor:
# Bug
From Discord bug report, which included that the deck may not even be deleted. Reported in Firefox on Android.

From testing locally I saw that the onclick of the DeckPreview row, which redirects to the deck page, is triggered after the clicks of the actions in the modal. I'm not 100% sure why given that the modal is rooted outside the row itself, but there was a similar bug in this area fixed a while ago with the same behaviour.

# Solution
Stop the event propagation when the delete confirmation modal buttons are clicked.

# Testing

## Before
Can see the click on close redirects to the deck page. Though it didn't occur on my local, I can see why the bug reporter saw the deleting the deck didn't actually delete but redirected them; if there is some latency of the delete API call the browser could cancel it before the server even begins to process it, or the deck page load gets the deck before the delete finishes syncing to the database.
old-mobile-deck-delete-actions-act-weird

## After

Clicking close doesn't redirect, nor does delete.
new-mobile-deck-delete-actions-sane

### DIFF
--- a/src/client/components/modals/ConfirmDeleteModal.tsx
+++ b/src/client/components/modals/ConfirmDeleteModal.tsx
@@ -28,10 +28,24 @@ const ConfirmDeleteModal: React.FC<ConfirmDeleteModalProps> = ({ isOpen, setOpen
       </ModalBody>
       <ModalFooter>
         <Flexbox direction="row" className="w-full justify-end" gap="2">
-          <LoadingButton block color="danger" onClick={submitDelete}>
+          <LoadingButton
+            block
+            color="danger"
+            onClick={(e: React.MouseEvent) => {
+              e.stopPropagation();
+              submitDelete();
+            }}
+          >
             Delete
           </LoadingButton>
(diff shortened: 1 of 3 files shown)


src/client/components/modals/ConfirmDeleteModal.tsx
```diff
@@ -28,10 +28,24 @@ const ConfirmDeleteModal: React.FC<ConfirmDeleteModalProps> = ({ isOpen, setOpen
       </ModalBody>
       <ModalFooter>
         <Flexbox direction="row" className="w-full justify-end" gap="2">
-          <LoadingButton block color="danger" onClick={submitDelete}>
+          <LoadingButton
```
KaelenProctor, line 31: The only other component using ConfirmDeleteModal is the Blog Delete action which I tested and works the same. The delete button on a blog doesn't live in any other clickable area, so nothing else was listening for the click action.

PR #2656, opened 2025-03-16
Fix bug in card full names that resulted in Art Cards appearing before the actual cards

KaelenProctor:
From Discord feature request https://discord.com/channels/592787488523943937/1350090569023619092

Once I dug into it and walked through the tree names -> card names logic, I found that the full names of art cards had two spaces between the name and "art card". After searching the tree through each character the user has typed, it then does a depth first search from there until a maximum number of names are found. From testing spaces are before other characters in the tree, whereas [ are before other characters. Thus due to the double space in the full name for art cards, they took precedence over the rest. See images below to help illustrate

I've typed "island" and this shows the tree after it has traversed through those characters:
image
The `[` sub-tree is all the full names of island cards, whereas `a` takes you to the art cards and `f` takes you to "island fish jasconius" (that's a card???).

In comparison this is the same in the current CubeCobra. Due to the double space, the art cards end up before the `[` sub-tree.
image
Depending on how many printings a card name has the art cards will use up the maximum number of names to show.

# Testing

## Before

When search names, the art cards were after the actual card:
old-art-cards-last-with-no-versions
old-art-cards-after-with-no-versions2
old-art-cards-after-with-no-versions3

But when searching versions the art cards were first:
old-art-cards-first
old-art-cards-first2
old-art-cards-can-be-full

## After

When search names, same behaviour:
new-art-cards-after-actual-no-versions
new-art-cards-after-actual-no-versions2
new-art-cards-after-actual-no-versions3

When searching versions, now art cards are after real cards:
new-art-cards-after-actual
new-art-cards-after-actual2
new-art-cards-after-actual3

### DIFF
--- a/src/jobs/utils/update_cards.ts
+++ b/src/jobs/utils/update_cards.ts
@@ -99,9 +99,13 @@ export function convertName(card: ScryfallCard, preflipped: boolean) {
     str = str.substring(0, str.indexOf('/')); // first name
   }
 
+  //Trim the card name here before potentially adding art series suffix.
+  //Important the first or second name was extracted
+  str = str.trim();
+
   if (card.layout === 'art_series') {
     str = `${str} ${ART_SERIES_CARD_SUFFIX}`;
   }
 
-  return str.trim();
+  return str;
 }
--- a/src/jobs/utils/update_cards.ts
+++ b/src/jobs/utils/update_cards.ts
@@ -1,0 +1,107 @@
+import { ART_SERIES_CARD_SUFFIX } from '../../client/utils/cardutil';
+
+export interface ScryfallCard {
+  id: string;
+  name: string;
+  lang: string;
+  set: string;
+  collector_number: string;
+  set_name: string;
+  released_at: string;
+  reprint: boolean;
+  border_color: string;
+  promo: boolean;
+  promo_types: string[];
+  digital: boolean;
+  finishes: string[];
+  prices: {
+    usd: string | null;
+    usd_foil: string | null;
+    usd_etched: string | null;
+    eur: string | null;
+    tix: string | null;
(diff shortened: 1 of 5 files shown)


src/jobs/utils/update_cards.ts
```diff
@@ -0,0 +1,111 @@
+import { ART_SERIES_CARD_SUFFIX } from '../../client/utils/cardutil';
```
KaelenProctor, line 1: Happy to adjust the file name / path if there is a preference for something else

src/jobs/utils/update_cards.ts
```diff
@@ -0,0 +1,111 @@
...
+    str = str.substring(str.indexOf('/') + 2); // second name
+  } else if (card.name.includes('/') && card.layout !== 'split') {
+    // NOTE: we want split cards to include both names
+    // but other double face to use the first name
+    str = str.substring(0, str.indexOf('/')); // first name
+  }
+
+  //Trim the card name here before potentially adding art series suffix.
+  //Important the first or second name was extracted
+  str = str.trim();
```
KaelenProctor, line 104: Moving the trim before the addition of "Art Card" is the functional difference.
The same fix could have been done by adjusting `str = str.substring(0, str.indexOf('/'));` to take from 0 to index - 1, though that only works if only 1 space between end of first name and `//`

tests/utils/utils.test.ts
```diff
@@ -0,0 +1,292 @@
+import { normalizeName } from '../../src/client/utils/cardutil';
```
KaelenProctor, line 1: I thought I was going to be changing the turnToTree logic so I wrote tests, which are always nice to have regardless.

PR #2658, opened 2025-03-19
Prevent last step of a pack in custom draft format from being a "pass", which stalls the draft

KaelenProctor:
# Problem
If the last step of a pack is "pass" the draft cannot proceed. When building the step queue from the pack steps, after each pack there is an "endpack" step added which is what swaps to the next pack.

# Solution
As discussed in Discord we want to:
a) Validate custom drafts cannot be saved with a "pass" as the last step in a pack
b) Fix (normalize) custom draft formats when loaded, so a bad state isn't used to create a draft

# Testing
In the master branch I setup a couple of draft formats for a cube where the last step is a "pass".
Lots of unit tests added to help prevent regressions in the future.

## Before

Gif of what happens when getting to the end of pack 1 on master and gets stuck on the last "pass" step.
old-pack-ending-pass-stalls-draft

Image of the format stored as loaded into the playtest page, where you can see the last "pass" step.
old-ending-pass-steps-stored

Same thing but looking direct at dynamo content.
custom-draft-with-ending-pass-step-stored

Custom draft format UI not highlighting the error of having a "pass" as the last step in a pack.
old-ending-pass-step-not-marked-as-invalid

## After

New validation error for a pack ending in a "pass" step.
new-custom-draft-validation-on-ending-pass-step

And after fixing that:
new-custom-draft-add-step-is-valid

Additionally now the "Add step" button adds pass then pick (instead of pick then pass) which prevents it from being an immediate validation error in relation to the last step being a valid action.
new-custom-format-add-slot-then-add-step-is-valid

Gif of a draft using a custom format that is saved with a last step of "pass", but the normalization on read fixed that.
new-draft-doesnt-get-stuck-at-end-of-pack

Due to the normalization on load, editing the custom draft doesn't show the last step pass and then saving the format will fix the stored version to a good state.
new-pack-normalization-fixes-last-step-pass-on-next-save

### DIFF
--- a/src/client/layouts/MainLayout.tsx
+++ b/src/client/layouts/MainLayout.tsx
@@ -48,4 +48,6 @@ const MainLayout: React.FC<MainLayoutProps> = ({ children, loginCallback = '/' }
   );
 };
 
+MainLayout.displayName = 'MainLayout';
+
 export default MainLayout;
--- a/src/client/pages/CubeDraftPage.tsx
+++ b/src/client/pages/CubeDraftPage.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useContext, useEffect, useMemo, useState } from 'react';
 
 import { DndContext } from '@dnd-kit/core';
-import type { State } from 'src/router/routes/draft/finish.ts';
 
 import { Card, CardBody, CardHeader } from 'components/base/Card';
 import Text from 'components/base/Text';
@@ -11,7 +10,7 @@ import RenderToRoot from 'components/RenderToRoot';
 import { CSRFContext } from 'contexts/CSRFContext';
 import { DisplayContextProvider } from 'contexts/DisplayContext';
 import Cube from 'datatypes/Cube';
-import Draft, { DraftStep } from 'datatypes/Draft';
+import Draft from 'datatypes/Draft';
 import DraftLocation, { addCard, location, removeCard } from 'drafting/DraftLocation';
 import { locations } from 'drafting/DraftLocation';
 import useLocalStorage from 'hooks/useLocalStorage';
@@ -19,7 +18,7 @@ import CubeLayout from 'layouts/CubeLayout';
 import MainLayout from 'layouts/MainLayout';
 import { makeSubtitle } from 'utils/cardutil';
 
-import { getCardDefaultRowColumn, setupPicks } from '../../util/draftutil';
+import { getCardDefaultRowColumn, getInitialState, setupPicks } from '../../util/draftutil';
 
 interface CubeDraftPageProps {
   cube: Cube;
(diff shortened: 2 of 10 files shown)


src/client/layouts/MainLayout.tsx
```diff
@@ -48,4 +48,6 @@ const MainLayout: React.FC<MainLayoutProps> = ({ children, loginCallback = '/' }
   );
 };
 
+MainLayout.displayName = 'MainLayout';
```
KaelenProctor, line 51: Added displayName because I was having challenges debugging react components with React Dev tools, when most names are "Anonymous"

src/datatypes/Draft.ts
```diff
@@ -26,80 +26,6 @@ export type DraftState = {
   steps: DraftStep[];
 }[][];
 
-export const DEFAULT_STEPS: DraftStep[] = [
```
KaelenProctor, line 29: Happy to adjust, but it felt odd to me to have logic in the datatypes file so I moved them to draftutils and added tests.

src/client/pages/CubeDraftPage.tsx
```diff
@@ -68,31 +67,6 @@ const processPredictions = (json: PredictResponse, packCards: any[]) => {
   return packCards.map((card) => predictionsMap.get(card.oracle_id) || 0);
 };
 
-const getInitialState = (draft: Draft): State => {
```
KaelenProctor, line 71: Moved to draftutils and added tests. Ideally later we can do this in the route and pass the initial state direct to the page.

src/dynamo/models/cube.js
```diff
@@ -165,6 +166,12 @@ const hydrate = async (cube) => {
   cube.owner = await User.getById(cube.owner);
   cube.image = getImageData(cube.imageName);
 
+  const draftFormats = cube?.formats || [];
+  //Correct bad custom draft formats on load, so any page using them are using good versions
+  for (let format of draftFormats) {
+    format = normalizeDraftFormatSteps(format);
```
KaelenProctor, line 172: Fix the formats on load

src/util/draftutil.ts
```diff
@@ -82,15 +83,20 @@ export const flattenSteps = (steps: Step[], pack: number): FlattenedStep[] => {
   return res;
 };
 
-export const defaultStepsForLength = (length: number): Step[] =>
-  new Array(length)
+export const defaultStepsForLength = (length: number): Step[] => {
```
KaelenProctor, line 86: Not a functional change. Using normalizeDraftSteps so we don't have duplicate logic of removing any last "pass" steps

src/util/draftutil.ts
```diff
@@ -82,15 +83,20 @@ export const flattenSteps = (steps: Step[], pack: number): FlattenedStep[] => {
...
     ])
-    .flat()
-    .slice(0, length * 2 - 1) // Remove the final pass.
-    .map((action) => ({ ...action }));
+    .flat();
+
+  return normalizeDraftSteps(steps).map((step) => ({
+    action: step.action,
+    //We know not null here
+    amount: step.amount!,
```
KaelenProctor, line 97: For whatever reason Step type doesn't allow amount to be null but DraftStep does. I kinda think the return type should be DraftStep anyways, since this isn't doing anything with pick/cardsInPack properties which are part of Step.

src/util/draftutil.ts
```diff
@@ -103,28 +109,6 @@ export const getStepList = (initialState: any[]): FlattenedStep[] =>
     ])
     .flat();
 
-export const nextStep = (draft: Draft, cardsPicked: number): string | null => {
```
KaelenProctor, line 106: Function no longer used

src/util/draftutil.ts
```diff
@@ -314,3 +298,146 @@ export const setupPicks: (rows: number, cols: number) => any[][][] = (rows: numb
...
+    let amount = 0;
+
+    if (!pack.steps) {
+      // this is ok, it just means the pack is a default pack
+      continue;
+    }
+
+    const stepsLength = pack.steps.length;
+    const lastStep = pack.steps[stepsLength - 1];
+    if (lastStep?.action === 'pass') {
```
KaelenProctor, line 355: This block has the new validation logic

src/util/draftutil.ts
```diff
@@ -314,3 +298,146 @@ export const setupPicks: (rows: number, cols: number) => any[][][] = (rows: numb
...
+    for (const pack of seat) {
+      const stepsLength = pack.steps.length;
+      if (stepsLength === 0) {
+        continue;
+      }
+
+      stepQueue.push(...pack.steps);
+
+      //Backwards compatability, add endpack step to the end of the pack if the backend hasn't already
+      if (pack.steps[stepsLength - 1]?.action !== 'endpack') {
```
KaelenProctor, line 426: Slight change where endpack is only added if the last step in the pack isn't already that. This made more sense when getInitialState was still in the page component, but now it seems more likely we'd do this in the route and pass the computed initial state to the frontend; if so then this logic isn't that useful.

PR #2663, opened 2025-03-23
Fix blog post redirection on edit/delete if the cube is deleted

KaelenProctor:
# Problem

The user's blog shows blog posts from cubes that are deleted. You cannot go to the post itself (it will 404) but it shows the buttons to edit or delete. Those buttons work, except that they will redirect you to the cube's blog page which will then 404. You see the successful edit/delete but also an error saying the blog post wasn't found.

# Solution
The initial solution was to check if the cube exists or not when deciding where to redirect to, and if the cube doesn't exist anymore redirect to /dashboard so the user can still see the successful action flash.

Then I decided to improve the UX more to return the user to the page they were at when taking the action if possible, else go to /dashboard if the cube doesn't exist. Blog posts can be seen in the following places:
1. The cube's blog
2. The user's blog
3. The blog post itself - Extra check that the post wasn't deleted, as cannot return you to it once it is gone

# Questions/concerns
1. Overall the redirecting back to the page you can from could be more complicated than it is worth
2. The cube blog post view doesn't have any button/link back to the cube blog, which could make navigation annoying
3. Should we allow blog posts that belong to deleted cubes to be accessible on their blog post page? They can be edited and deleted from the user blog already

# Testing

### Happy cases
Successfully deleting a blog post from the user's blog redirects to the cube blog page:
old-delete-blog-from-user-blog-redirects-to-cube-blog

Successfully deleting a blog post from itself redirects to the cube blog page:
old-delete-blog-from-itself-redirects-to-cube-blog

Successfully deleting a blog post from the cube blog page, redirects to the same:
old-delete-blog-redirects-to-cube-blog

Successfully editing a blog post from itself redirects to the cube blog page:
old-edit-blog-from-itself-redirects-to-cube-blog

Successfully editing a blog post from the user's blog redirects to the cube blog page:
old-edit-blog-from-user-blog-redirects-to-cube-blog

Successfully editing a blog post from the cube blog page, redirects to the same:
old-edit-blog-from-cube-blog

### Edge cases

Deleting a blog of a deleted cube gives you both success and fail:
old-delete-blog-in-deleted-cube-error-page2

### Happy paths

Creating a blog post from the cube blog:
new-create-blog-post

Editing a blog post from itself returns you there:
new-edit-blog-post-from-itself

Editing a blog post from the user blog returns you there:
new-edit-blog-post-from-user-blog

Editing a blog post from the cube blog returns you there:
new-edit-blog-post-from-cube-blog

Deleting a blog post from the user blog returns you there:
new-delete-blog-post-from-user-blog

Deleting a blog post from the user blog, when the cube is deleted, returns you there:
new-delete-blog-post-from-user-blog-when-cube-is-deleted

Deleting a blog post from the cube blog returns you there:
new-delete-blog-post-from-cube-blog

#### Deleting blog post from itself

Redirects to the cube blog, if the cube still exists:
new-delete-blog-post-from-itself-redirects-to-cube-blog

#### Cube no longer exists

Editing a blog post of a deleted cube from user blog, redirects back to user blog. Also shows that you cannot go to the blog post itself because the cube is deleted.
new-edit-blog-post-when-cube-is-deleted-from-user-blog

Deleting a blog post of a deleted cube, from the cube blog, redirects to dashboard:
new-delete-blog-post-when-cube-is-deleted

Editing a blog post of a deleted cube, from the cube blog, redirects to dashboard:
new-edit-blog-post-when-cube-is-deleted

#### Referrer that doesn't belong to cubecobra

Setup:
new-override-referer-to-evil
new-evil-referer-used

From the user's blog, the redirect goes to the cube blog because it exists and because the referrer wasn't trusted:
new-edit-blog-in-user-blog-redirects-to-cube-when-evil-referer

From the user's blog, the redirect goes to the dashboard because the cube doesn't exist and because the referrer wasn't trusted:
new-edit-blog-in-user-blog-redirects-to-dashboard-when-evil-referer-and-cube-is-gone

### DIFF
--- a/src/router/routes/cube/blog.ts
+++ b/src/router/routes/cube/blog.ts
@@ -15,15 +15,29 @@ import CubeType from '../../../datatypes/Cube';
 import { FeedTypes } from '../../../datatypes/Feed';
 import UserType from '../../../datatypes/User';
 
-const getRedirectUrl = async (cubeId: string): Promise<string> => {
+const getRedirectUrl = async (req: Request, cubeId: string): Promise<string> => {
   const cube = await Cube.getById(cubeId);
-  return await getRedirectUrlForCube(cube);
+  return await getRedirectUrlForCube(req, cube);
 };
 
-const getRedirectUrlForCube = async (cube: CubeType): Promise<string> => {
-  if (cube) {
-    const id = cube.shortId || cube.id;
-    return `/cube/blog/${encodeURIComponent(id)}`;
+const getRedirectUrlForCube = async (req: Request, cube: CubeType): Promise<string> => {
+  const referrer = util.getSafeReferrer(req);
+  //If not a valid referrer on this website, either send to the cube or the dashboard
+  if (referrer === null) {
+    if (cube) {
+      return `/cube/blog/${encodeURIComponent(cube.id)}`;
+    } else {
+      return '/dashboard';
+    }
+  }
+
+  //Use the referrer to know where to return the user, based on the places blog posts can be seen
+  if (referrer.includes('/user/blog/')) {
+    return referrer;
+  } else if (referrer.includes('/cube/blog/blogpost/')) {
+    return referrer;
+  } else if (referrer.includes('/cube/blog/') && cube) {
+    return referrer;
   } else {
     return '/dashboard';
   }
@@ -34,7 +48,7 @@ export const createBlogHandler = async (req: Request, res: Response) => {
     const cubeId = req.params.id;
     //Generally going to assume the cube exists here. Definitely required for a new blog, not so for an edit
     const cube = await Cube.getById(cubeId);
-    const redirectUrl = await getRedirectUrlForCube(cube);
+    const redirectUrl = await getRedirectUrlForCube(req, cube);
 
     if (req.body.title.length < 5 || req.body.title.length > 100) {
       req.flash('danger', 'Blog title length must be between 5 and 100 characters.');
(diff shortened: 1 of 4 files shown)


src/router/routes/cube/blog.ts
```diff
@@ -156,10 +199,11 @@ export const deleteBlogHandler = async (req: Request, res: Response) => {
     const { id } = req.params;
     const { user } = req;
 
+    //The ensureAuth middleware means this should never occur
     if (!user) {
       req.flash('danger', 'Please login to delete a blog post.');
 
-      return redirect(req, res, `/cube/blog/${encodeURIComponent(id)}`);
+      return redirect(req, res, `/cube/blog/blogpost/${encodeURIComponent(id)}`);
```
KaelenProctor, line 206: The redirect wasn't logical before. The Id is the blog post id, but was trying to redirect as if it is a cube id.
That said, since the ensureAuth middleware will redirect the user to login before the handler is called this part of the code is not used.

src/util/util.js
```diff
@@ -309,6 +309,33 @@ function getBaseUrl() {
   return (process.env?.HTTP_ONLY === 'true' ? 'http://' : 'https://') + process.env.DOMAIN;
 }
 
+/*
+ * Returns the root relative path, based on the referrer header. Since referrer header can be manipulated,
+ * we only return a non-null value if it is not to another domain
+ */
+const getSafeReferrer = (req /* Request */) => {
```
KaelenProctor, line 316: Referer's cannot be trusted in general. So we only allow it as a referer if it belongs to the application domain. That allows these cases:
1. The referer is relative (eg /foobar or foobar) because the use of getBaseUrl() in URL.parse makes it relative to the app
2. The referer is absolute and the host after parsing the url matches the application domain or its www subdomain

PR #2666, opened 2025-03-28
Add User auto blog setting, which determines whether "Create blog post" is checked or not by default (and defaults to false)

KaelenProctor:
From Discord feature request https://discord.com/channels/592787488523943937/1349131518055944242/1349131518055944242 though this only covers:

> req 1: setting handled at the account level
> req 2: setting off by default

of the 3 requested parts.

Happy to hear feedback on the wording around the setting/tooltips or even the property name.

This has a dependency on https://github.com/dekkerglen/CubeCobra/pull/2665 so depending on which of the two is merged first, the other will need small updates.

# Testing

## After
Auto blog setting configurable in display settings.
new-create-blog-post-user-setting

New tooltip beside the "Create Blog Post" in the editing cube section.
new-tooltip-for-create-blog-post

Just as before, the last checked status of "Create Blog Post" is stored in local storage.
create-blog-checked-in-localstorage

When auto blog is off for the user, can see that it is used when the local storage value is gone.
new-default-blog-posts-off-and-setting-used

When auto blog is on for the user, the same thing occurs.
new-default-blog-posts-on-and-setting-used

### DIFF
--- a/src/client/contexts/CubeContext.tsx
+++ b/src/client/contexts/CubeContext.tsx
@@ -240,7 +240,10 @@ export function CubeContextProvider({
   const [sortTertiary, setSortTertiary] = useQueryParam('s3', defaultSorts[2]);
   const [sortQuaternary, setSortQuaternary] = useQueryParam('s4', defaultSorts[3]);
   const [filterResult, setFilterResult] = useState({});
-  const [useBlog, setUseBlog] = useLocalStorage<boolean>(`${cube.id}-useBlog`, true);
+  const [useBlog, setUseBlog] = useLocalStorage<boolean>(
+    `${cube.id}-useBlog`,
+    typeof user?.autoBlog !== 'undefined' ? user.autoBlog : false,
+  );
 
   const allTags = useMemo(() => {
     const tags = new Set<string>();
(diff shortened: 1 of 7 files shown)


src/client/contexts/CubeContext.tsx
```diff
@@ -240,7 +240,10 @@ export function CubeContextProvider({
   const [sortTertiary, setSortTertiary] = useQueryParam('s3', defaultSorts[2]);
   const [sortQuaternary, setSortQuaternary] = useQueryParam('s4', defaultSorts[3]);
   const [filterResult, setFilterResult] = useState({});
-  const [useBlog, setUseBlog] = useLocalStorage<boolean>(`${cube.id}-useBlog`, true);
+  const [useBlog, setUseBlog] = useLocalStorage<boolean>(
+    `${cube.id}-useBlog`,
+    typeof user?.autoBlog !== 'undefined' ? user.autoBlog : false,
```
KaelenProctor, line 245: I could slightly improve the typing by updating the UserContextValue type (in UserContext) to not simply extend User, but define the defaultPrinting, gridTightness, and autoBlog settings as always defined since we are setting them in hydration if they aren't set.
Admittedly all that would do is slightly reduce the verbosity of a statement like this when having to check set or not, which really is only annoying for a boolean compared to a string.

PR #2672, opened 2025-04-19
Return draft notifications to the solo, within CubeCobra bot drafts

KaelenProctor:
The notifications were lost when the Draftmancer integration was created, potentially due to the conversion to Typescript. Unfortunately cannot see how to add the notification for Draftmancer drafts as the API endpoint does not include any information about the draft owner.

# Testing
Create unit tests before modifying the code. Full coverage apart from some of the type guard ternary conditions.

## After

Shows user TestUser2 doing a draft of the Synergy2 Cube owned by kaelenproctor, which has draft notifications enabled:
new-notifications0
new-notifications1
image
new-notifications2

### DIFF
--- a/src/router/routes/draft/finish.ts
+++ b/src/router/routes/draft/finish.ts
@@ -52,7 +52,7 @@ const FinishDraftBodySchema = Joi.object({
     .required(),
 }).unknown(true); // allow additional fields
 
-const validateBody = (req: Request, res: Response, next: NextFunction) => {
+export const validateBody = (req: Request, res: Response, next: NextFunction) => {
   const { error } = FinishDraftBodySchema.validate(req.body);
   if (error) {
     res.status(400).json({ error: error.details[0].message });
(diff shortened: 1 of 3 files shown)


src/router/routes/draft/finish.ts
```diff
@@ -81,7 +84,8 @@ const handler = async (req: Request, res: Response) => {
       });
     }
 
-    if (typeof draft.owner !== 'string' && draft.owner?.id !== req.user.id) {
+    const draftOwnerId = typeof draft.owner !== 'string' ? draft.owner?.id : draft.owner;
```
KaelenProctor, line 87: This change ended up being unnecessary, I thought I was going to use draftOwnerId in sendDraftNotification but turns out for addNotification I needed the objects

dekkerglen:
For the draftmancer route, we can make notifications from anonymous users - we used to do that before we required users to be logged in to draft

KaelenProctor:
> For the draftmancer route, we can make notifications from anonymous users - we used to do that before we required users to be logged in to draft

Great, I'll look into adding that

PR #2673, opened 2025-04-27
Backfilling tests on the Content dynamo model, which identified a bug

KaelenProctor:
1. Adding tests to backfill
2. Identified bug with Content.batchPut where the temporary object is written instead of the desired document - this only executes when a podcast updates its image and then it updates all episodes to the same image. eg not often

### DIFF
--- a/src/dynamo/models/content.ts
+++ b/src/dynamo/models/content.ts
@@ -262,7 +262,7 @@ const content = {
         await putBody(document.id, body);
       }),
     );
-    client.batchPut(docs);
+    client.batchPut(docs.map((doc) => doc.document));
   },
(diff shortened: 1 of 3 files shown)


src/dynamo/models/content.ts
```diff
@@ -262,7 +262,7 @@ const content = {
         await putBody(document.id, body);
       }),
     );
-    client.batchPut(docs);
+    client.batchPut(docs.map((doc) => doc.document));
```
KaelenProctor, line 265: docs is temporary object that was added when refactoring to TypeScript, that contains the document to write to dynamo plus the body to write to S3. The batchPut to dynamo was using the temp documents instead of the desired part

PR #2675, opened 2025-04-29
Update cardutil.ts

Jhedlundfay:
Modify the word counter function to split oracle text on all whitespace characters instead of just the space character

### DIFF
--- a/src/client/utils/cardutil.ts
+++ b/src/client/utils/cardutil.ts
@@ -476,7 +476,7 @@ export const cardWordCount = (card: Card): number => {
     return 0;
   }
   if (card.details.wordCount === undefined) {
-    card.details.wordCount = card.details.oracle_text.split(' ').length;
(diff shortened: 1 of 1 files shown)


src/client/utils/cardutil.ts
```diff
@@ -476,7 +476,7 @@ export const cardWordCount = (card: Card): number => {
     return 0;
   }
   if (card.details.wordCount === undefined) {
-    card.details.wordCount = card.details.oracle_text.split(' ').length;
+    card.details.wordCount = card.details.oracle_text.split('/\s/').length;
```
KaelenProctor, line 479: For a regular expression split the single-quotes must be removed, eg `.split(/\s/).length;`. Otherwise it splits on the string "/\s/"

dekkerglen:
I don't think this can be merged as is,  @Jhedlundfay are you planning on changing it?

PR #2676, opened 2025-04-30
Draftmancer API tests and addition of anonymous notification for the cube owner (assuming enabled)

KaelenProctor:
# Changes
1. Refactoring of logic from publish.ts to new draftmancerUtils and type file, in order to simplify unit testing
2. Complete test coverage of both
3. Make from optional in Notification to support anonymous notifications, not from an existing CubeCobra user
4. Related to that, adjust the Notifications history UI so that if from isn't set then it does not link

# Challenges
Even working with Copilot to help create the tests, it was a struggle to get the test data into a reasonable form such that it could be used for multiple tests.

# Testing
1. Ensuring all tests pass
2. Using curl to send a test draftmancer API body (shared by Dekkaru in discord on an old bug related to drafting) and validate the draft is saved sucessfully, and viewable in the UI

## After

Draft notification to the cube owner
new-notification-without-user-link-from-draftmancer

Draft created by importing from curl:
image

### DIFF
--- a/src/dynamo/models/notification.ts
+++ b/src/dynamo/models/notification.ts
@@ -26,7 +26,7 @@ const client = createClient({
   ],
 });
 
-module.exports = {
+const notification = {
   getById: async (id: string): Promise<Notification> => (await client.get(id)).Item as Notification,
   getByToAndStatus: async (
     to: string,
@@ -103,3 +103,7 @@ module.exports = {
     ),
   createTable: async (): Promise<DocumentClient.CreateTableOutput> => client.createTable(),
 };
+
+module.exports = notification;
+
+export default notification;
--- a/src/router/routes/api/draftmancer/publish.ts
+++ b/src/router/routes/api/draftmancer/publish.ts
@@ -1,49 +1,22 @@
 import Joi from 'joi';
 
-import { cardCmc, cardType, detailsToCard } from '../../../../client/utils/cardutil';
+import { detailsToCard } from '../../../../client/utils/cardutil';
 import { CardDetails } from '../../../../datatypes/Card';
 import type CubeType from '../../../../datatypes/Cube';
 import DraftType, { DraftmancerLog, DraftmancerPick } from '../../../../datatypes/Draft';
+import { PublishDraftBody } from '../../../../datatypes/Draftmancer';
 import type DraftSeatType from '../../../../datatypes/DraftSeat';
 import Cube from '../../../../dynamo/models/cube';
 import Draft from '../../../../dynamo/models/draft';
 import { NextFunction, Request, Response } from '../../../../types/express';
-import { cardFromId, getReasonableCardByOracle } from '../../../../util/carddb';
-import { deckbuild } from '../../../../util/draftbots';
+import { cardFromId } from '../../../../util/carddb';
+import {
+  buildBotDeck,
+  formatMainboard,
+  formatSideboard,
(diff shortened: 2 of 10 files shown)


src/dynamo/models/notification.ts
```diff
@@ -26,7 +26,7 @@ const client = createClient({
   ],
 });
 
-module.exports = {
+const notification = {
```
KaelenProctor, line 29: Somehow I missed updating the exports to modern style when this was converted to typescript

src/router/routes/api/draftmancer/publish.ts
```diff
@@ -1,50 +1,19 @@
...
 import Draft from '../../../../dynamo/models/draft';
+import Notification from '../../../../dynamo/models/notification';
 import { NextFunction, Request, Response } from '../../../../types/express';
-import { cardFromId, getReasonableCardByOracle } from '../../../../util/carddb';
-import { deckbuild } from '../../../../util/draftbots';
+import { cardFromId } from '../../../../util/carddb';
+import { buildBotDeck, formatMainboard, formatSideboard, getPicksFromPlayer } from '../../../../util/draftmancerUtil';
 import { setupPicks } from '../../../../util/draftutil';
 
-interface Pick {
```
KaelenProctor, line 15: Exported to datatypes to be shared

src/router/routes/api/draftmancer/publish.ts
```diff
@@ -122,103 +79,25 @@ export const handler = async (req: Request, res: Response) => {
...
-
-          pickorder.push(pickIndex);
-          draftmancerPicks.push({
-            booster: pack,
-            pick: pickIndex,
-          });
-        }
-      }
-
-      for (const oracleId of player.decklist.main) {
```
KaelenProctor, line 147: Before this these two loops over the main and sideboard would be done for bots, but those would be overwritten after doing the bot deckbuild().

As far as I can see anything in player.decklist.main and player.decklist.side should also be in the picks arrays, which are processed by getPicksFromPlayer below and adds to the cards array

src/util/draftmancerUtil.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  if (index === -1) {
+    cards.push(card);
+    return cards.length - 1;
+  }
+
+  return index;
+};
+
+export const getPicksFromPlayer = (
```
KaelenProctor, line 21: Functions all extracted as is, apart from a couple renames for clarity which I think was mainly cards -> cardDetails

src/router/routes/api/draftmancer/publish.ts
```diff
@@ -122,103 +79,25 @@ export const handler = async (req: Request, res: Response) => {
...
 
+    let drafterName: string = '';
     for (const player of publishDraftBody.players) {
-      const draftmancerPicks: DraftmancerPick[] = [];
       let mainboard: number[][][] = setupPicks(2, 8);
       let sideboard: number[][][] = setupPicks(1, 8);
-      const pickorder: number[] = [];
-      const trashorder: number[] = [];
-
-      for (const pick of player.picks) {
```
KaelenProctor, line 132: Do to all the loops and calls to functions that would have to be mocked like getReasonableCardByOracle, I extracted bits whole into the draftmancerUtils, which both simplified testing the extract bits but also this endpoint

PR #2679, opened 2025-05-20
Use distinct oracle ids when fetching combos, and only show the first copy of a card by oracle id in each combo UI.

KaelenProctor:
As reported in Discord.

# Testing

## Before

Duplicate combos:
old-duplicate-combos

## After

Distinct combos:
new-distinct-combos


KaelenProctor:
Failing tests fixed in https://github.com/dekkerglen/CubeCobra/pull/2678

PR #2690, opened 2025-06-08
Fixing a few errors in the Record feature

KaelenProctor:
# Fixes

1. UI error caused by linking to seat -1. Caused by the Trophy winner name not aligning with the player names
2. UI error caused by adding a new seat after creating the draft decks

# Testing

### Trophy error

Setup
1. Create record with players
2. Assign a trophy to a player
3. Edit the player list so the trophy winner name no longer exists
4. Go to the Trophy archive page
5. Click the "View deck" for the trophy winner

UI error about mainboard
image

Can see from the HTML the seat is -1 because the trophy winners name wasn't found in the array of player names, so array findIndex gives -1
image

### Added seats error

1. Create a record with N players
2. Add a deck to at least one player
3. Add another player to the draft
4. Page errors

3 seats
image

Clicking into the record
image

### Trophy error

Does not link to an invalid seat. Instead gives error and tooltip about fixing the misalignment
image

### Added seats error

Seats
image

Decks dropdown only shows the seats with drafts
image

### DIFF
--- a/src/router/routes/cube/records/edit/trophy.ts
+++ b/src/router/routes/cube/records/edit/trophy.ts
@@ -26,7 +26,7 @@ export const editTrophyHandler = async (req: Request, res: Response) => {
       return redirect(req, res, '/404');
     }
 
-    const trophy = JSON.parse(req.body.trophy);
+    const trophy: string[] = JSON.parse(req.body.trophy);
     if (!Array.isArray(trophy) || trophy.length === 0) {
       req.flash('danger', 'Trophy must be a non-empty array');
       return redirect(req, res, `/cube/record/${req.params.id}?tab=2`);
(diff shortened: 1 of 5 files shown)


src/router/routes/cube/records/edit/trophy.ts
```diff
@@ -26,23 +26,28 @@ export const editTrophyHandler = async (req: Request, res: Response) => {
       return redirect(req, res, '/404');
     }
 
-    const trophy = JSON.parse(req.body.trophy);
+    const trophy: string[] = JSON.parse(req.body.trophy);
     if (!Array.isArray(trophy) || trophy.length === 0) {
```
KaelenProctor, line 30: Should we allow the trophies to be removed?

Issue #2689, opened 2025-06-08
Directly show backside of card in markdown


F43nd1r:
**What would you like to be able to do?**
I'd like to embed the backside of cards directly, e.g. `[[!Sword of the Realms]]`

**Describe how you would use the feature**
We use the site on mobile, where hovering over DFCs isn't a thing

**Additional context**
Currently, we can only show the front side `[[!/Halvar, God of Battle]]` with the back side on hover.
Workaround is to add the backside as image, but that doesn't show up the same size as autocards. `![Sword of the Realms](https://cards.scryfall.io/small/back/f/e/feddb683-a8f8-4def-bd3e-6d630c42cd34.jpg?1631054082)`

Would also be happy with a new syntax based on the front side, something like `[[!//Halvar, God of Battle]]`

PR #2691, opened 2025-06-08
#2689 - Support showing card back image in the page with markdown

KaelenProctor:
Implements #2689

# Testing

new-markdown-card-backside

Of course it doesn't work so well when the card doesn't have a backside

image


src/client/pages/MarkdownPage.tsx
```diff
@@ -807,8 +829,4 @@ MarkdownPage.propTypes = {
   loginCallback: PropTypes.string,
 };
 
-MarkdownPage.defaultProps = {
```
KaelenProctor, line 810: Noticed a deprecated warning about this and saw other pages used `loginCallback = '/'` in the component definition, so updated

PR #2700, opened 2025-07-04
#2696 - Legal cards in Vintage are cards that are legal or restricted, when it comes to filtering

### LINKED ISSUE #2696: Using legal:vintage in card search sorts away the restricted cards
Unsure if this is a bug, intentional or just devs didnt think about, so i put it here in feature requests

I believe that legal:vintage should include all restricted cards, both for intuition and to be in line with orther sides searches (like scryfalls)

So you skip having to type legal:vintage or restricted:vintage (i know this is very minor but i think it is confusing as it is now)

**Additional context**
Scam links for comparisons:
https://cubecobra.com/tool/topcards?p=0&s=Elo&d=descending&f=legal%3Avintage+mox
https://scryfall.com/search?q=legal%3Avintage+mox&unique=cards&as=grid&order=name

KaelenProctor:
Fixes #2696 

Technically this should extend to all formats but functionally only Vintage has restricted cards

# Testing

## Scryfall as control

Legal moxes in Vintage
scryfall-vintage-mox-legality

Restricted moxes
scryfall-vintage-mox-restricted

Negation of legal moxes
scryfall-not-vintage-legal-mox

## Before

Legal moxes in Vintage, missing the OG moxes
old-vintage-mox-legality
Restricted filter shows the moxes
old-vintage-mox-restricted

## After

Now moxes show when searching legal vintage cards
new-vintage-mox-legality

Still show in restricted filter
new-vintage-mox-restricted

Negation of legal moxes. I dunno why, but gleemox is not in the Cube Cobra card database which is why it still doesn't show
new-not-vintage-legal-mox

### DIFF
--- a/src/client/generated/filtering/cardFilters.js
+++ b/src/client/generated/filtering/cardFilters.js
@@ -24,6 +24,9 @@ import {
   setCountOperation,
   devotionOperation,
   propertyComparisonOperation,
+  genericCondition,
+  comparisonCondition,
+  legalitySuperCondition,
 } from '../../filtering/FuncOperations';
 import {
   CARD_CATEGORY_DETECTORS,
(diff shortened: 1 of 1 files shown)


nearley/cardFilters.ne
```diff
@@ -107,19 +110,6 @@ condition -> (
   | notesCondition
 ) {% ([[condition]]) => condition %}
 
-@{%
-const genericCondition = (propertyName, propertyAccessor, valuePred) => {
```
KaelenProctor, line 111: Moving out of nearley into JS so easy to unit test

PR #2701, opened 2025-07-06
Lazy loading Card versions and Autocomplete information

KaelenProctor:
# Changes
Based on discussion from Discord on ways to improve performance / reduce network traffic.

I highly leveraged Copilot for this, enough that I ran out of free credits lol. I had googled about JavaScript proxies and React Lazy loading beforehand, but didn't have experience with the syntax.

1. Card versions in CubeContext lazy loaded when used, which is in the CardModal or ListView
2. Autocomplete input lazy loads the tree data when the input is not-empty. This mainly affects the edit part with Adds and Removes
    1. For removes we could built the tree from the cards already loaded, instead of making a fetch call
    2. In some usages of the Autocomplete such as cube/user image, it isn't actually lazy since once rendered the input has a value

## Potential downsides
By lazy loading people with slow connections might have reduced user experience since these changes don't use any loading state. That would be most visible looking at the cube list view.

## Looking at next
In CubeAnalysis page the recommendations and combos load right away. Part of the reason why is that the TabbedView renders all contents but hides the non-active tab, so the useEffect in the tabs triggers right away. I played around with only rendering the active tab and while that worked, it got a bit messy since each time the tab rendered it would fetch the data unless it was "cached" outside the component.

# Testing

Tested in:
* Desktop Chrome, Firefox, Edge
* Mobile Chrome and Firefox

## Before

Cube pages fetch two autocomplete lists and card versions right away
old-cube-list-eager-loads
Packages modal loads autocomplete list right away
old-package-eager-load
Advanced filter modal eager loading tags (also the could be computed from the current card list so tag updates exist)
old-eager-load-filter-tag

## After

On Cube page load the 3 calls are lazy loaded
new-lazy-load-3
Can see the card versions loaded in when looking at the list view, which uses them.
new-lazy-load-4
Or when clicking into the card modal.
new-lazy-load-5

The edit autocompletes lazy load too.
new-autocomplete-lazy-load
Same with card name autocomplete in record
new-lazy-load-autocomplete-records
Or in the cube filter advanced modal for tags
new-lazy-load-autocomplete-tags

### DIFF
--- a/src/client/components/base/AutocompleteInput.tsx
+++ b/src/client/components/base/AutocompleteInput.tsx
@@ -223,20 +223,27 @@ const AutocompleteInput: React.FC<AutocompleteInputProps> = ({
   const [visible, setVisible] = useState(false);
   const { hideCard } = useContext(AutocardContext);
 
+  // Lazy load the tree only when the user starts typing (value is non-empty)
   useEffect(() => {
+    if (!value) return; // Don't fetch until user types something
+    let cancelled = false;
     const wrapper = async () => {
       try {
         if (!treeCache[treeUrl]) {
           treeCache[treeUrl] = fetchTree(treeUrl, treePath);
         }
-        setTree((await treeCache[treeUrl]) ?? {});
+        const result = await treeCache[treeUrl];
+        if (!cancelled) setTree(result ?? {});
       } catch (e) {
         // eslint-disable-next-line no-console
         console.error('Error getting autocomplete tree.', e);
       }
     };
     wrapper();
-  }, [treePath, treeUrl]);
+    return () => {
+      cancelled = true;
+    };
+  }, [treePath, treeUrl, value]);
 
   const handleChange = useCallback(
     (event: React.ChangeEvent<HTMLInputElement>) => {
--- a/src/client/components/cube/ListView.tsx
+++ b/src/client/components/cube/ListView.tsx
@@ -100,7 +100,7 @@ const ListView: React.FC<ListViewProps> = ({ cards }) => {
         sortQuaternary || 'Alphabetical',
         cube.showUnsorted || false,
       ),
-    [cards, cube.showUnsorted, sortQuaternary, sortPrimary, sortSecondary],
+    [cards, sortPrimary, sortSecondary, sortTertiary, sortQuaternary, cube.showUnsorted],
   );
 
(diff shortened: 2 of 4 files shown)


src/client/components/base/AutocompleteInput.tsx
```diff
@@ -223,20 +223,27 @@ const AutocompleteInput: React.FC<AutocompleteInputProps> = ({
   const [visible, setVisible] = useState(false);
   const { hideCard } = useContext(AutocardContext);
 
+  // Lazy load the tree only when the user starts typing (value is non-empty)
   useEffect(() => {
+    if (!value) return; // Don't fetch until user types something
+    let cancelled = false;
```
KaelenProctor, line 229: Cancelled stuff probably can be dropped, comes from the AI but I didn't ask for it

src/client/components/cube/ListView.tsx
```diff
@@ -100,7 +100,7 @@ const ListView: React.FC<ListViewProps> = ({ cards }) => {
         sortQuaternary || 'Alphabetical',
         cube.showUnsorted || false,
       ),
-    [cards, cube.showUnsorted, sortQuaternary, sortPrimary, sortSecondary],
+    [cards, sortPrimary, sortSecondary, sortTertiary, sortQuaternary, cube.showUnsorted],
```
KaelenProctor, line 103: This and WithCardModal are unrelated lint fixes, since I dislike the IDE showing the files in red

src/client/contexts/CubeContext.tsx
```diff
@@ -227,6 +227,89 @@ export function CubeContextProvider({
...
+    setVersionDictLoaded(true);
+    setVersionDictLoading(false);
+  }, [csrfFetch, cube.cards, loadVersionDict, versionDictLoaded, versionDictLoading]);
+
+  // The versionDictProxy is a Proxy around the versionDict state.
+  // Its purpose is to lazily load the version dictionary from the server
+  // only when it is first accessed by a consumer (such as ListView or CardModal).
+  // This avoids fetching all card versions unless they are actually needed.
+
+  const versionDictProxy = useMemo(() => {
```
KaelenProctor, line 286: The Proxy methodology allowed the versionDict to be lazy loaded without changes to the components using it

PR #2705, opened 2025-07-20
Improve weirdness around editing card CMC

KaelenProctor:
# Problem

As reported in Discord on mobile it is very hard to edit the CMC on mobile. Deleting the number does not appear to have an effect, even though the card is in a changed state. This occurs because the UI uses cardCmc for the input value, so when the text field is cleared the function falls back to the details CMC which often looks the same. However the card is in a state where the override cmc is an empty string yet the text input shows the card's true cmc. And if you are typing quickly, you'd end up adding a the desired number to the true CMC instead of replacing the CMC.

On desktop this is alleviated because you can easily select all the text and enter a new number, which bypasses. But the same problem exists.

In the end a user can easily cause the card's override cmc to be an invalid value, though thankfully since the rest of the code uses cardCmc it will fallback to the true CMC when it comes to things like filtering and sorting.

# Solution
Don't use cardCmc for the value of the text input, so that whatever the user typed is shown in the input and the fallback doesn't mess with what they've typed. Then use browser input validation standards to warn the user about missing or invalid CMCs, with the hopes they will correct it before saving the edits.

Updated on both the card modal and the list view.

# Testing

## Before

To illustrate the problem better, I started with an Esper Sentinal set with a CMC of 4 (true CMC is 1). I delete the 4 but the text input shows 1 since its showing the true CMC
old-removing-cmc-falls-back-to-real-cmc

On mobile the same happens, though in this case I am deleting the 4 and entering a 3 but the end result is a 13.
old-removing-cmc-falls-back-to-real-cmc-mobile-hard-to-edit

Same action as the previous screenshot, except starting from the true cmc of 1.
old-removing-cmc-falls-back-to-real-cmc-mobile-hard-to-edit2

For extra technical details, can see that when you delete the text but it still shows 3, in the backend the card's cmc override is set to blank in comparison to the true cmc.
old-can-remove-cmc-but-wouldnt-know-in-ui

## After
**Note: Took these gifs before updating the messaging from CMC to Mana Value**

Showing the input warnings when the cmc is in a bad state. The input now uses the card's true CMC as the placeholder value, in case you need a reminder.
new-card-modal-cmc-warnings

Similar stuff on the list view
new-list-cmc-warnings

The validation will not prevent saving bad cmcs, so that has not changed.
new-can-still-save-invalid-cmc

Zero is valid CMC, but negative isn't
new-zero-valid-cmc
new-negative-cmcs-not-allowed

Firefox warnings (Edge looks the same as Chrome):
new-firefox-cmc-warnings
new-firefox-cmc-warnings2

Chrome mobile validation messages
new-mobile-cmc-warnings

### DIFF
--- a/src/client/components/card/CardModal.tsx
+++ b/src/client/components/card/CardModal.tsx
@@ -89,6 +89,8 @@ const CardModal: React.FC<CardModalProps> = ({
   allTags,
 }) => {
   const [versions, setVersions] = useState<Record<string, CardDetails> | null>(null);
+  //Default value is undefined so the field only shows red border when wrong, neutral otherwise
+  const [isCmcValid, setCmcValid] = useState<boolean | undefined>(undefined);
 
   useEffect(() => {
     if (!versionDict[normalizeName(cardName(card))]) {
@@ -118,6 +120,7 @@ const CardModal: React.FC<CardModalProps> = ({
   //When the card id changes, then update the image used. If we just checked card, it would be different
   //if any field in the modal is edited, which would lead the image to reset to the front face.
   if (prevCardID !== card.cardID) {
+    setCmcValid(undefined);
     setIsFrontImage(true);
     setPrevCardID(card.cardID);
     setImageUsed(getCardFrontImage(card));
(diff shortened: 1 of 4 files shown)


src/client/components/card/CardModal.tsx
```diff
@@ -456,16 +495,7 @@ const CardModal: React.FC<CardModalProps> = ({
...
-                color="primary"
-                block
-                className="items-center text-sm"
-                onClick={() => {
-                  if (card.editIndex !== undefined && card.board !== undefined) {
-                    revertEdit(card.editIndex, card.board);
-                  }
-                }}
-              >
+              <Button color="primary" block className="items-center text-sm" onClick={revertAction}>
```
KaelenProctor, line 498: I ended up not needing to change the onclick here, but didn't bother undoing the refactoring

src/client/components/cube/ListView.tsx
```diff
@@ -132,6 +132,26 @@ const ListView: React.FC<ListViewProps> = ({ cards }) => {
...
+    } else {
+      input.setCustomValidity('');
+    }
+    input.reportValidity();
+  }, []);
+
+  const onCmcChange = useCallback(
+    (card: CardType, e: React.ChangeEvent<HTMLInputElement>) => {
+      const input = e.target;
+      updateField(card, 'cmc', input.value);
```
KaelenProctor, line 149: I couldn't determine a good way to consolidate the logic between the two. In this version `card` is also needed because it isn't implicit like it is in the card modal

tests/utils/utils.test.ts
```diff
@@ -450,6 +450,16 @@ describe('cardCmc', () => {
     card.cmc = 'not a number' as any;
     expect(cardCmc(card)).toBe(4);
   });
+
+  it('defaults to details cmc if card cmc is empty', () => {
```
KaelenProctor, line 454: Added additional test case for empty string which should fallback to the true cmc

PR #2710, opened 2025-07-23
Extract CSRF from HTML response in order to improve caching performance (thus lower network out)

KaelenProctor:
Came from Discord discussion on performance.

CubeCobra uses a token synchronisation CSRF pattern, where the randomly generated csrf secret is stored in the user's session and a hashed version of that is written into the HTML page for React props. However since the version in the HTML is a hash which has its own randomness, this means the HTML contents always change even if you simply refresh the page so the ETag cache value is never useful. On a cube list for example, that can mean the server returns 150KB that is 99.99% the same as the last time.

# Testing

## General
1. Validated that the increased lengths of the secrets won't cause a release problem. The CSRF code will continue to use the old secret in the user's session, but the next time they logout and login they'd get a random secret of longer length
2. CSRF mismatches still are caught (but with improved error page now)

## Before

Gif showing how the csrf token in the HTML changes on each reload
old-csrf-token-changes-each-page-load

So on a cube list, that means the download size is large each time the page reloads
old-reloading-cube-list-full-download-each-time

Before if the CSRF didn't match (which I triggered by explicitly add extra characters) then we get a stack trace error
old-csrf-error-stack-trace

## After

Now the page contents are the same on subsequent refreshes, so we get 304 not modified and the network is much lower. The response time doesn't change since the server still has to do all the work, generate the HTML, and then see that its Etag matches what was sent in the request.
new-page-reload-304-not-modified-so-network-is-reduced

The fetch of the CSRF token on page load shows it change each time, just as before
new-csrf-fetch-different-each-time

If we edit something, such as cards in the cube, then on reload its a 200 OK since the ETag changed
new-editing-contents-results-in-etag-and-full-contents

Now on CSRF error its a simple unauthorized response
new-csrf-error-handling

### DIFF
--- a/src/routes/middleware.js
+++ b/src/routes/middleware.js
@@ -31,6 +31,14 @@ const csrfProtection = [
     saltLength: 16,
     secretLength: 24,
   }),
+  function (err, req, res, next) {
+    if (err.code !== 'EBADCSRFTOKEN') {
+      return next(err);
+    }
+
+    // handle CSRF token errors here
+    res.status(403).send('Unauthorized');
+  },
   (req, res, next) => {
     const { nickname } = req.body;
 
@@ -76,14 +84,14 @@ const answers = [
   'forest', // 'What is the name of the basic land that produces green mana?'
 ];
 
-async function recaptcha (req, res, next) {
+async function recaptcha(req, res, next) {
   const { captcha, question, answer } = req.body;
-  
+
   if (!question || !answer) {
     req.flash('danger', 'Please answer the security question');
     return redirect(req, res, '/');
   }
-  
+
   const index = questions.indexOf(question);
 
   if (index === -1 || answers[index].toLowerCase() !== answer.toLowerCase()) {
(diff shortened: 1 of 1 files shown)


src/routes/middleware.js
```diff
@@ -27,9 +27,20 @@ const ensureRole = (role) => async (req, res, next) => {
 };
 
 const csrfProtection = [
-  csurf(),
+  csurf({
+    saltLength: 16,
```
KaelenProctor, line 31: Changes to the salt and secret length aren't required, but I feel its nice to go above the defaults

src/types/express.ts
```diff
@@ -7,6 +7,7 @@ export interface Request extends express.Request {
   logger: {
     error: (...args: any[]) => void;
   };
+  csrfToken: () => string;
```
KaelenProctor, line 10: So TS files don't complain it doesn't exist. csrfToken is added to the request via the csurf middlware

PR #2723, opened 2025-08-27
Fix downloading models from S3.

KaelenProctor:
Used code from Google AI overview

# Testing
In my local S3 bucket the model folder only contained indexToOracleMap.json hence when I tested the code it worked as expected. I uploaded the rest of my local model folder, which contained all of the below, to the bucket using awslocal s3 cp --recursive. I renamed the model folder on disk to model.bak in order to compare.
```
total 1416
drwxr-xr-x  7 kp kp    4096 Aug 27 07:37 .
drwxr-xr-x 32 kp kp    4096 Aug 27 07:47 ..
drwxr-xr-x  2 kp kp    4096 Jan  9  2025 correlation_decoder
drwxr-xr-x  2 kp kp    4096 Jan  9  2025 cube_decoder
drwxr-xr-x  2 kp kp    4096 Jan  9  2025 deck_build_decoder
drwxr-xr-x  2 kp kp    4096 Jan  9  2025 draft_decoder
drwxr-xr-x  2 kp kp    4096 Jan  9  2025 encoder
-rw-r--r--  1 kp kp 1420323 Aug  4 09:43 indexToOracleMap.json
```

## Before

Ran the download. `diff` showed indexToOracleMap.json was the same, but encoder/group1-shard10of14.bin (as example) to be different. All the .bin were different.

When running the app and trying an predictions, all the results are null.

## After
Removed the model folder and ran the download.

`diff` shows no differences. Predictions in the app have results.

### DIFF
--- a/src/util/downloadModel.js
+++ b/src/util/downloadModel.js
@@ -31,9 +31,24 @@ const downloadFromS3 = async () => {
       }
     }
 
-    fs.writeFileSync(file.Key, await res.Body.transformToString());
-    // eslint-disable-next-line no-console -- Debugging
-    console.log(`Downloaded ${file.Key}`);
+    //Models are a mix of JSON and binary, can't simply call res.Body.transformToString
+    if (res.Body) {
+      // Create a writable stream to the local file
+      const fileStream = fs.createWriteStream(file.Key);
+
+      // Pipe the S3 object's Body (readable stream) to the file stream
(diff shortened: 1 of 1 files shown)


KaelenProctor:
Looking at all the getObject calls in the code, this was the only one I found that isn't on a JSON file where transformToString works

PR #2728, opened 2025-09-07
Eb/feature/p1p1 UI tweaks

Eetai:
UI Tweaks per feedback:

- Alert added when share button is clicked
- Borrowed 'Get Image' pattern from samplepack
- Also borrowed og:image pattern from samplepack so the links look nicer
- Put the UI in a CubeLayout with 'playtest' selected, this feature is now more clearly 'happening within the context of a cube'
- Because of that I also moved the URLs from tool/p1p1 to cube/p1p1, and put a redirect so links generated yesterday don't break. That can probably be removed at some point in the future.

### DIFF
--- a/src/routes/cube/index.js
+++ b/src/routes/cube/index.js
@@ -41,16 +41,64 @@ const Changelog = require('../../dynamo/models/changelog');
 const router = express.Router();
 router.use(csrfProtection);
 
-// Shared function for generating pack images
+// Shared function for generating pack images for samplepack and p1p1pack images
+// Overarching logic here - we want the most visually appealing layout for the pack
+// For non-prime numbers that's a balanced rectangle (ideally a square)
+// For prime numbers, we use whatever the next non-prime's results would be
+// Only use factors in the range 2-7 to avoid things like 2x11 for 22
+const generatePackLayout = (count) => {
+  const isPrime = (n) => {
+    if (n < 2) return false;
+    for (let i = 2; i <= Math.sqrt(n); i++) {
+      if (n % i === 0) return false;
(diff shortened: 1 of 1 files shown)


src/routes/cube/index.js
```diff
@@ -41,6 +41,27 @@ const Changelog = require('../../dynamo/models/changelog');
...
 router.use(csrfProtection);
 
+// Shared function for generating pack images
+const generatePackImage = async (cards) => {
+  if (cards.some((card) => !card.imgUrl && !card.details.image_normal)) {
+    throw new Error('One or more cards in this pack are missing images.');
+  }
+  
+  // Try to make it roughly 5 times as wide as it is tall in cards.
+  const width = Math.floor(Math.sqrt((5 / 3) * cards.length));
```
dekkerglen: I know we do the same thing for sample packs but would be nice to prefer square images - like 16 should be 4x4, 12 should be 4x3, 18 should be 6x3

src/routes/cube/index.js
```diff
@@ -896,33 +917,98 @@ router.get('/samplepackimage/:id/:seed', async (req, res) => {
...
+  } catch (err) {
+    return handleRouteError(req, res, err, '/404');
+  }
+});
+
+router.get('/p1p1packimage/:packId', async (req, res) => {
+  try {
+    req.params.packId = req.params.packId.replace('.png', '');
+    const { packId } = req.params;
+    const p1p1PackModel = require('../../dynamo/models/p1p1Pack');
```
dekkerglen, line 1036: should move these imports to top of the file

PR #2731, opened 2025-09-20
Create Daily Pack 1 Pick 1 feature

Eetai:
New daily P1P1 feature on dashboard page. 

Screenshot 2025-09-19 at 1 55 32 AM

Screenshot 2025-09-20 at 1 20 36 AM

- Scheduler uses rotateDailyP1P1 to generates a pack a day from a random featured queue cube every day at 12:01 am
- generateBalancedPack makes 10 potential packs and picks the one with the lowest highest bot weight for a card. This ensures we don't create a boring pack with a Sol Ring in it.
- Owner get an notification when their cube's pack is featured

- Daily packs managed in new dynamo table DAILY_P1P1: 
 {
  id: string;
  packId: string; // P1P1_PACKS id
  cubeId: string;
  date: number; // Date when this became daily P1P1
  isActive: boolean; // Whether this is currently the active daily P1P1
}

The big new things in the UI:
- Dashboard card (DailyP1P1Card.tsx)
- Archive page (DailyP1P1HistoryPage.tsx)

Notes/Next steps
- Not yet added to /landing so logged out users don't see this, would need to figure out where it would go
- Future PR: expose generateBalancedPack to users directly


src/client/pages/DailyP1P1HistoryPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+              const cards = item.pack.cards.map((card) => {
+                return card.details ? detailsToCard(card.details) : card;
+              });
+
+              return (
+                <Card key={item.id} className="border">
+                  <CardHeader>
+                    <Flexbox direction="row" justify="between" alignItems="center">
+                      <Text semibold md>
+                        {new Date(item.date).toLocaleDateString('en-US', {
```
KaelenProctor, line 82: Might be worth starting a helper for date formatting so we have consistency

src/dynamo/models/dailyP1P1.ts
```diff
@@ -0,0 +1,114 @@
...
+    {
+      name: 'ByDate',
+      partitionKey: 'isActive',
+      sortKey: 'date',
+    },
+  ],
+  attributes: {
+    id: 'S',
+    date: 'N',
+    isActive: 'S', // 'true' for active, 'false' for inactive
```
KaelenProctor: Could use a BOOL type?

src/jobs/rotateDailyP1P1.js
```diff
@@ -0,0 +1,164 @@
...
+
+    // Get the cube
+    const cube = await Cube.getById(selectedQueueItem.cube);
+    if (!cube) {
+      // eslint-disable-next-line no-console
+      console.error('Selected cube not found:', selectedQueueItem.cube);
+      return;
+    }
+
+    // eslint-disable-next-line no-console
```
KaelenProctor: nit: might as well disable for the file, have fewer comments per lines

src/router/routes/tool/api/deletep1p1.ts
```diff
@@ -30,20 +29,24 @@ export const deleteP1P1Handler = async (req: Request, res: Response) => {
       return res.status(404).json({ error: 'P1P1 pack not found' });
     }
 
+    // Prevent deletion of packs created by CubeCobra (daily P1P1 packs)
+    if (pack.createdBy === 'CubeCobra') {
```
KaelenProctor, line 33: Perhaps rather than a check here the pack could have an attribute defining whether deletable? Allows for other reasons defined later with less conditional logic

src/dynamo/models/dailyP1P1.ts
```diff
@@ -0,0 +1,114 @@
...
+  },
+
+  createTable: async (): Promise<CreateTableCommandOutput> => client.createTable(),
+
+  delete: async (id: string): Promise<void> => {
+    await client.delete({ id });
+  },
+};
+
+export default dailyP1P1;
```
KaelenProctor: By also doing `module.exports = dailyP1P1;` it aligns with the other typescript dynamo models, and doesn't need the `require(..).default` in root.js

KaelenProctor (commented):
Pretty much all my comments are minor and up to you if desired to make changes. I would suggest adding unit or integration tests, to make sure we keep adding to test coverage instead of decreasing.

src/client/pages/DailyP1P1HistoryPage.tsx
Eetai: so FormatttedDate already exists, but uses a different pattern...I'm not sure creating another variant makes sense untli there's something else that needs to use it

src/router/routes/tool/api/deletep1p1.ts
Eetai: deletability is user dependent outside daily-p1p1s, and it's possible the logic will change in the future, so I'd rather keep it as is

src/dynamo/models/dailyP1P1.ts
Eetai: bool isn't an option for the declared types used as keys in dynamo, but we don't need isActive to be one, so I tweaked that

dekkerglen (approved):

PR #2734, opened 2025-09-28
Fix card versions, such as in markdown, not matching exact scryfall id

KaelenProctor:
# Testing

## Before

Old border card shown even though the scryfall id is a newer version
image

Versions list shows backside cards 
image

# After

Correct versions of front and back sides show
new-versions-and-backs

Versions list only shows frontside. Unlike scryfall, CubeCobra doesn't include the current version of the card being looked at in the side list
image

Can see on Scryfall the versions doesn't contain backsides only
image

### DIFF
--- a/src/routes/tools_routes.js
+++ b/src/routes/tools_routes.js
@@ -2,6 +2,7 @@
 require('dotenv').config();
 
 const express = require('express');
+import { validate as uuidValidate } from 'uuid';
 
 import { Period } from '../datatypes/History';
 import carddb, {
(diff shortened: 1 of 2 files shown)


src/routes/tools_routes.js
```diff
@@ -32,6 +33,34 @@ router.use(csrfProtection);
...
+const chooseIdFromInput = (req) => {
+  //ID is scryfall id or a card name (eg. via Autocomplete hover)
+  const printingPreference = req?.query?.defaultPrinting || req?.user?.defaultPrinting;
+  let { id } = req.params;
+
+  if (!uuidValidate(id)) {
+    // if id is a cardname, redirect to the default version for that card
+    const possibleName = cardutil.decodeName(id);
+    const ids = getIdsFromName(possibleName);
+    if (ids !== undefined && ids.length > 0) {
```
KaelenProctor, line 45: This if statement was the root problem. getIdsFromName now returns an empty array as a default instead of undefined, which mean it didn't go into getMostReasonable. At that point it ended up finding the wrong version.

src/routes/tools_routes.js
```diff
@@ -32,6 +33,34 @@ router.use(csrfProtection);
 /* Minimum number of picks for data to show up in Top cards list. */
 const MIN_PICKS = 100;
 
+const chooseIdFromInput = (req) => {
+  //ID is scryfall id or a card name (eg. via Autocomplete hover)
+  const printingPreference = req?.query?.defaultPrinting || req?.user?.defaultPrinting;
+  let { id } = req.params;
+
+  if (!uuidValidate(id)) {
```
KaelenProctor, line 41: If we have a uuid then rather than trying to figure a card by name, we can skip that. We end up returning nothing from this method, and then cardFromId gets it

PR #2735, opened 2025-09-28
Improve blog create/edit so that validation failures can be fixed right away, without loss

KaelenProctor:
As reported on Discord, you can hit the limit for blog post text (10K characters) without knowing ahead and when you hit the limit the rest of the text is cut off. 

This improves the UX by changing from a form post to a POST fetch, so on failures we can show an error message and the user can fix. As well, properly handle max length on text areas including showing the current character count against the limit.

# Testing

# After

Errors are shown without a page reload, so nothing is lost. The user can modify and try saving again.
new-failure-message-with-page-reload

The server side checks are also done, even in this case where I used dev tools to change the max length of the field (though not the component).
new-max-length-server-side-error

Can see the character count as typed and when the limit is hit, shows in red.
new-text-limit-and-indicator

Text max length and limit also shows on places like Cube Overview
new-overview-character-limit

Successful editing of a blog post, even from the user's blog page.
new-edit-blog-post

Cube update blog also has max length and count indicator
image

### DIFF
--- a/src/client/components/EditBlogModal.tsx
+++ b/src/client/components/EditBlogModal.tsx
@@ -1,11 +1,12 @@
-import React, { useMemo, useState } from 'react';
+import React, { useCallback, useContext, useState } from 'react';
 
 import BlogPost from '../../datatypes/BlogPost';
+import { CSRFContext } from '../contexts/CSRFContext';
+import Alert, { UncontrolledAlertProps } from './base/Alert';
 import Button from './base/Button';
 import Input from './base/Input';
 import { Flexbox } from './base/Layout';
 import { Modal, ModalBody, ModalFooter, ModalHeader } from './base/Modal';
-import CSRFForm from './CSRFForm';
 import TextEntry from './TextEntry';
 
 interface EditBlogModalProps {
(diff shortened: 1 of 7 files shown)


src/client/components/EditBlogModal.tsx
```diff
@@ -16,48 +17,74 @@ interface EditBlogModalProps {
...
+        id: postId,
+      }),
+    });
+
+    const body = await response.json();
+    if (response.ok) {
+      setAlerts([{ color: 'success', message: body.ok }]);
+      //Reload page to ensure state is updated
+      window.location.replace(body.redirect);
+    } else if (response.status === 403) {
```
KaelenProctor, line 47: We don't have a standard way to handle authentication errors with csrfFetch, at least to provide a meaningful error message.

src/client/components/EditBlogModal.tsx
```diff
@@ -16,48 +17,74 @@ interface EditBlogModalProps {
...
+            minLength={5}
+            maxLength={100}
+            name="title"
+            type="text"
+            value={title}
+            onChange={(e) => setTitle(e.target.value)}
+          />
+          {post && <input type="hidden" name="id" value={post.id} />}
+          <TextEntry name="markdown" value={markdown} setValue={setMarkdown} maxLength={10000} />
+          {alerts.map(({ color, message }) => (
```
KaelenProctor, line 71: Add alerts to show errors

src/client/components/base/TextArea.tsx
```diff
@@ -65,11 +83,22 @@ const TextArea: React.FC<TextAreaProps> = ({
...
+        maxLength={maxLength}
+        {...props}
       />
+      {showCharacterLimit && maxLength ? (
+        <div
+          className={classNames('p-2 rounded-b-md', {
+            'bg-red-100 text-red-800': textLength >= maxLength,
+          })}
+        >
+          {textLength} / {maxLength}
```
KaelenProctor, line 99: Text count against limit, with background red when the limit is hit

src/router/routes/cube/blog.ts
```diff
@@ -58,64 +58,62 @@ export const createBlogHandler = async (req: Request, res: Response) => {
...
     const { user } = req;
 
     if (!user) {
-      req.flash('danger', 'Please Login to publish a blog post.');
+      res.status(401).json({ error: 'Please Login to publish a blog post.' });
+      return;
+    }
 
-      return redirect(req, res, redirectUrl);
+    if (req.body.markdown && req.body.markdown.length > 10000) {
```
KaelenProctor, line 74: Moved up so the length validation also applies to edits. The dynamo model cuts anything over 10K off anyways so it wasn't a loophole

src/routes/middleware.js
```diff
@@ -12,6 +12,15 @@ const ensureAuth = (req, res, next) => {
   return redirect(req, res, '/user/login');
 };
 
+const ensureAuthJson = (req, res, next) => {
+  if (req.isAuthenticated()) {
+    return next();
+  }
+
+  res.status(403).json({ error: 'You must be logged in.' });
```
KaelenProctor, line 20: Version of ensureAuth that returns JSON error instead of redirect, for fetch API calls

tests/cube/blog/api.test.ts
```diff
@@ -52,94 +52,90 @@ jest.mock('../../../src/dynamo/models/feed', () => ({
...
     const cube = createCube({ id: 'cube-id' });
     (Cube.getById as jest.Mock).mockResolvedValue(cube);
-    (util.getSafeReferrer as jest.Mock).mockReturnValue(`/cube/blog/${cube.shortId}`);
 
-    await call(createBlogHandler).withFlash(flashMock).withParams({ id: 'cube-id' }).withBody({ title: 'Hi' }).send();
+    const res = await call(createBlogHandler).withParams({ id: 'cube-id' }).withBody({ title: 'Hi' }).send();
 
-    expect(render.redirect).toHaveBeenCalledWith(expect.anything(), expect.anything(), `/cube/blog/${cube.shortId}`);
-    expect(flashMock).toHaveBeenCalledWith('danger', 'Blog title length must be between 5 and 100 characters.');
+    expect(res.status).toBe(400);
```
KaelenProctor, line 65: Changes all from redirect/flash to http response/JSON

PR #2736, opened 2025-09-28
Fix redirect back to the page you were on after login

KaelenProctor:
The loginCallback wasn't doing anything. Fixed by using safe redirect.

99% of these changes are removing the loginCallback prop

# Testing

## Before

Always sent to dashboard on login
old-login-always-to-dashboard

## After

Sent to the page you were on
new-after-login-return-to-page-you-were-on

Exception is when you started on the public landing page, take you to your dashboard on login
new-landing-login-to-dashboard

### DIFF
--- a/src/routes/users_routes.js
+++ b/src/routes/users_routes.js
@@ -8,6 +8,7 @@ const util = require('../util/util');
 const fq = require('../util/featuredQueue');
 import sendEmail from '../util/email';
 const { handleRouteError, render, redirect } = require('../util/render');
+const { getSafeReferrer } = require('../util/util');
 
 // Bring in models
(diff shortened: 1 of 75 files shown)


src/routes/users_routes.js
```diff
@@ -433,13 +434,10 @@ router.post('/login', async (req, res) => {
   }
 
   req.body.username = user.username;
-  // TODO: fix confirmation and check it here.
-  let redirectRoute = '/';
-  if (req.body.loginCallback) {
```
KaelenProctor, line 438: loginCallback was always /, so never did anything. getSafeReferrer tells us what page you were on within Cubecobra

PR #2740, opened 2025-09-29
Fix double body in post for pasting text into cube

KaelenProctor:
# Testing

## Before

There are two body values in the post, so on the backend req.body,body is a list and not a string, so when we match() to get lines that function doesn't exist on a list
old-duplicate-body

## After

Import works.
new-paste-import-working2 gif

### DIFF
--- a/src/client/components/base/TextArea.tsx
+++ b/src/client/components/base/TextArea.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect,useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import classNames from 'classnames';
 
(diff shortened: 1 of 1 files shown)


src/client/components/base/TextArea.tsx
```diff
@@ -38,7 +38,6 @@ const TextArea: React.FC<TextAreaProps> = ({
   disabled = false,
   showCharacterLimit = false,
   maxLength,
-  ...props
```
KaelenProctor, line 41: The rest of props includes stuff like name. When the Textarea is in a CSRFForm, the fact the textarea has a name caused duplication with the same state value passed to CSRFForm

PR #2758, opened 2025-10-11
Custom card name sorting and other fixes

KaelenProctor:
# Changes

1. Use custom card names in alphabetic sorting
2. Use custom card names in lower and full name functions
3. Use custom card names and other overrideable values in exports
4. Support custom cards in import/export better
5. Add tests

# Testing

## Before
Custom card names not sorting
old-custom-cards-not-sorting

Exporting CSV doesn't contain custom names
old-export-without-names

## After

Custom card names sorting, alongside regular cards
image

Importing custom cards (NOTE: the names don't match the images b/c I exported this cube from prod which doesn't have the custom names, so I had to stitch things together)
new-importing-custom-cards

Export contains custom names
new-updated-export

### DIFF
--- a/src/routes/cube/helper.js
+++ b/src/routes/cube/helper.js
@@ -212,7 +212,7 @@ function writeCard(res, card, maybe) {
   res.write(`${cardutil.cardRarity(card)},`);
   res.write(`${colorCategory},`);
   res.write(`${cardutil.cardStatus(card) || ''},`);
-  res.write(`${cardutil.cardFinish(card.finish)},`);
+  res.write(`${cardutil.cardFinish(card)},`);
   res.write(`${maybe},`);
   res.write(`${imgUrl},`);
   res.write(`${imgBackUrl},"`);


src/client/utils/Sort.ts
```diff
@@ -25,7 +26,7 @@ import {
   COLOR_COMBINATIONS,
   convertFromLegacyCardColorCategory,
 } from './cardutil';
-import { alphaCompare, arrayIsSubset, fromEntries } from './Util';
```
KaelenProctor, line 28: Moved into this code mainly to not have circular dependencies

src/client/utils/Sort.ts
```diff
@@ -257,7 +258,11 @@ const normalizeCollectorNumberForSorting = (card: Card): string => {
 };
 
 export const SortFunctions: Record<string, (a: any, b: any) => number> = {
-  Alphabetical: alphaCompare,
+  Alphabetical: (a: Card, b: Card): number => {
+    const textA = cardName(a).toUpperCase();
```
KaelenProctor, line 262: use cardName instead of direct details.name

src/client/utils/cardutil.ts
```diff
@@ -393,15 +395,28 @@ export const cardIsToken = (card: Card): boolean => card.details?.isToken ?? fal
...
+export const cardNameLower = (card: Card): string => {
+  if (isCustomCard(card)) {
+    return (card.custom_name?.toLowerCase() || card.details?.name_lower) ?? '';
+  }
+  return card.details?.name_lower ?? '';
+};
 
-export const cardFullName = (card: Card): string => card.details?.full_name ?? '';
+export const cardFullName = (card: Card): string => {
+  if (isCustomCard(card)) {
```
KaelenProctor, line 412: Use the same style of logic for custom cards

src/routes/cube/download.js
```diff
@@ -56,7 +57,7 @@ router.get('/cubecobra/:id', async (req, res) => {
     res.setHeader('Content-type', 'text/plain');
     res.charset = 'UTF-8';
     for (const card of mainboard) {
-      res.write(`${card.details.full_name}\r\n`);
+      res.write(`${cardName(card)}\r\n`);
```
KaelenProctor, line 60: Use the utils functions instead of direct access. Provides consistent when it comes to deal with card/custom overrides

src/routes/cube/helper.js
```diff
@@ -15,7 +15,7 @@ const { FeedTypes } = require('../../datatypes/Feed');
 const CARD_HEIGHT = 680;
 const CARD_WIDTH = 488;
 const CSV_HEADER =
-  'name,CMC,Type,Color,Set,Collector Number,Rarity,Color Category,status,Finish,maybeboard,image URL,image Back URL,tags,Notes,MTGO ID';
+  'name,CMC,Type,Color,Set,Collector Number,Rarity,Color Category,status,Finish,maybeboard,image URL,image Back URL,tags,Notes,MTGO ID,Custom';
```
KaelenProctor, line 18: Custom column is a flag for the import, since we don't have a column for id

src/util/cubefn.js
```diff
@@ -248,7 +249,15 @@ function CSVtoCards(csvString) {
         colorCategory: validatedColorCategory || null,
       };
 
-      const potentialIds = getAllVersionIds(card);
+      let potentialIds = [];
+      if (custom?.toLowerCase() === 'true') {
+        potentialIds = ['custom-card'];
```
KaelenProctor, line 254: The card catalog has the id 'custom-card' so we can look it up like any other card (eg `getAllVersionIds(card);`)

I think these two fields are the only overrides to do?

tests/cards/cardutil.test.ts
```diff
@@ -0,0 +1,89 @@
+import { cardFullName, cardName, cardNameLower } from '../../src/client/utils/cardutil';
+import { createCard, createCardDetails } from '../test-utils/data';
+
+describe('cardName', () => {
```
KaelenProctor, line 4: Thanks Copilot

PR #2762, opened 2025-10-13
Add Git commit to the versions page V2

KaelenProctor:
Better solution that in https://github.com/dekkerglen/CubeCobra/pull/2761

Assumptions
1. All changes are committed before running publish.js

# Testing

## After

See how the commit matches this branch
image

Also if the write-git-info script fails then both `npm run build` and `npm run devstart` will abort.

### DIFF
--- a/src/routes/root.js
+++ b/src/routes/root.js
@@ -13,7 +13,7 @@ const { handleRouteError, render, redirect } = require('../util/render');
 const { csrfProtection, ensureAuth } = require('./middleware');
 const { isCubeListed } = require('../util/cubefn');
 
-import { GIT_COMMIT } from '../util/git';
+const { GIT_COMMIT } = require('../util/git');
 
 const router = express.Router();
 


package.json
```diff
@@ -19,7 +19,7 @@
     "webpack-dev-server": "NODE_OPTIONS=--max_old_space-size=18192 npx webpack serve --config webpack.dev.mjs",
     "nearley": "sh nearley/helper.sh",
     "ci-build": "npm run nearley && tsc -b && NODE_OPTIONS=--max_old_space_size=16384 node_modules/.bin/webpack --mode production --config webpack.prod.mjs",
-    "devstart": "npm run nearley --watch & npm run nodemon & npm run webpack-dev-server & npx tailwindcss -i ./src/client/css/stylesheet.css -o ./public/css/stylesheet.css --watch=always",
+    "devstart": "npm run write-git-info && (npm run nearley --watch & npm run nodemon & npm run webpack-dev-server & npx tailwindcss -i ./src/client/css/stylesheet.css -o ./public/css/stylesheet.css --watch=always & wait)",
```
KaelenProctor, line 22: Make sure write-git-info passes before starting the rest of the commands in parallel

scripts/write-git-info.ts
```diff
@@ -0,0 +1,15 @@
+const { execSync } = require('child_process');
+const fs = require('fs');
+
+//Include the GIT_COMMIT into our src so it gets compiled into the build
+const OUT_FILE = 'src/util/git.ts';
+try {
+  const commitHash = execSync('git rev-parse HEAD').toString().trim();
+
+  const content = `export const GIT_COMMIT = '${commitHash ?? ''}';\n`;
```
KaelenProctor, line 9: If git command fails we get empty. Mostly as a hedge around people's dev environments.

PR #2768, opened 2025-10-20
Fix backside UUID validation in getversions

KaelenProctor:
# Changes

1. Fix endpoint failing on validation when a backside card is in the cube. Those don't have true uuids for their scryfall/card id because the update cards process appends "2"
2. Fix versions in the response for transforming cards. Both the front and back side versions were listed for both, both duplicating in the dropdown but also letting you switch between the front and back 

# Testing

## Before

400 bad request because insectile abberation's scryfall id isn't a uuid
old-back-face-ids-arent-uuids

Transforming cards have both back and front versions listed in the dropdown
old-transforming-cards-list-both-faces-in-versions

## After

Front face versions only include the front side
new-front-face-limited-versions

Back face versions only include the back side. Also shows the 400 bad request no longer fails
new-back-face-limited-versions

The alternate name stuff still works as before
new-alternate-names-still-good

### DIFF
--- a/src/routes/cube/api.js
+++ b/src/routes/cube/api.js
@@ -393,6 +393,7 @@ router.post(
         // Fixes duplicate versions listed in both the front and backsides
         result[normalized] = versionDetails
           .filter((c) => c.isExtra === card.isExtra)
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
           .map(({ isExtra, ...rest }) => rest); // remove isExtra from the response objects
       });
     });


src/routes/cube/api.js
```diff
@@ -376,19 +376,24 @@ router.post(
...
+          isExtra,
         }),
       );
       versions.forEach((card) => {
         const normalized = cardutil.normalizeName(card.name);
-        result[normalized] = versionDetails;
+        // Filter based on isExtra which is true if this is the backside of a card.
+        // Fixes duplicate versions listed in both the front and backsides
+        result[normalized] = versionDetails
+          .filter((c) => c.isExtra === card.isExtra)
```
KaelenProctor, line 395: tbh isExtra probably should be renamed at this point

src/util/validation.ts
```diff
@@ -1,7 +1,8 @@
 // Shared validation utilities
 
-// UUID v4 regex pattern
-const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+// UUID v4 regex pattern, PLUS allowing extra "2" at end for backsides of cards.
+// See convertId in update_cards.ts. Relevant to card IDs, not Oracle ids.
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}2?$/i;
```
KaelenProctor, line 5: Allow optional 2 at the end.

I checked in drafting where uuid validation is also used, but in that case it passes the oracle id to the backend so doesn't have the same problem

tests/cube/api/getversions.api.test.ts
```diff
@@ -301,4 +301,130 @@ describe('POST /cube/api/getversions', () => {
       },
     });
   });
+
+  it('should accept the backside id and include both back and front sides with the same multiple versions', async () => {
```
KaelenProctor, line 305: Failing test created to start, then passing after the API changes made

PR #2773, opened 2025-11-01
Roto Draft

davlin93:
WIP


src/client/hooks/usePollGoogleSheet.ts
```diff
@@ -0,0 +1,38 @@
...
+      return;
+    }
+
+    // Fetch immediately
+    fetchUrl();
+
+    // Set up polling every 60 seconds
+    const intervalId = setInterval(() => {
+      fetchUrl();
+    }, 6000);
```
dekkerglen: this is every 6 seconds, comment says 60

dekkerglen (dismissed):

src/client/hooks/usePollGoogleSheet.ts
davlin93: yeah I had it set to 6 seconds for testing. Will finish writing tests and cleaning up today. Copilot changes all look sane.

PR #2775, opened 2025-11-02
Support custom card types in sorting

KaelenProctor:
# Changes

1. Types sort isn't limited to MTG types, works with custom cards. Supertype sort still is, because we can't distinguish a supertype from a type in a type line
2. Subtypes sort works with custom cards
3. Types-Multicolor sort works with custom cards
4. New tests. Didn't go hog wild on how many cases.
5. A bit more info on how to write a valid type line
6. Handle "Time Lord" as a type made up of two words

## Questions
1. For Types-Mutlicolor, I put the custom types go before "Other". Should they maybe go before color combos?
2. I noticed for Contraptions (like https://scryfall.com/card/ust/205/targeting-rocket) we explicitly make the Type "Contraption" rather than Artifact, and then it has no subtype. That seems kinda backwards?
3. Some other sorts like artist, rarity, or legality don't work with custom cards since those fields can't be overridden. Should we support editing all fields for a custom card?

## Testing

I used https://cubecobra.com/cube/list/NR2025Corp from the Discord post as the example, exporting it to my local. I edited some types to make the format match MTG type lines with dashes to separate type from sub-types.

### Before

Sorting with custom types does nothing
old-custom-card-sorting-does-nothing

"Time Lord" treated as two sub types
image

### After

Can sort custom cards by custom types
new-custom-cards-sorting-by-type

Custom cards can sort based on Supertypes, but only MTG ones
new-custom-cards-can-support-mtg-supertypes

Subtypes work with custom cards
new-custom-cards-sorting-by-subtypes

And like normal, a card with multiple sub-types appears multiple times
new-custom-cards-multiple-subtypes-works

With Types-Multicolor a custom card can be grouped by MTG color combo names
new-custom-card-grouped-into-izzet

Custom card types "Valiant" appear in the grouping (at the end).
new-custom-types-in-sort

"Time Lord" is a single type
image

### DIFF
--- a/src/client/utils/Sort.ts
+++ b/src/client/utils/Sort.ts
@@ -423,9 +423,6 @@ const getTypesMulticolorLabels = (cube: Card[] | null): string[] => {
     const indexA = CARD_TYPES.indexOf(a);
     const indexB = CARD_TYPES.indexOf(b);
 
-    // If an element's id is not found in sortingOrderArray,
-    // it will be placed at the end (or handled as desired).
-    // For simplicity, this example assumes all IDs are present.
     return indexA - indexB;
   });
   //Sort alphabetically
@@ -731,7 +728,6 @@ export function cardGetLabels(card: Card, sort: string, showOther = false): stri
     //Overriding the types for contraptions and planes
     if (typesAndSuperTypes.includes('Contraption')) {
       ret = ['Contraption'];
-      //TODO: Shouldn't be this imo
     } else if (typesAndSuperTypes.includes('Plane')) {
       ret = ['Plane'];
     } else {


src/client/components/card/CardModal.tsx
```diff
@@ -344,7 +344,7 @@ const CardModal: React.FC<CardModalProps> = ({
                   }}
                 />
                 <Input
-                  label="Type"
+                  label={isCustomCard(card) ? 'Type (use - or — to separate types and subtypes)' : 'Type'}
```
KaelenProctor, line 347: Maybe should always show?

src/client/utils/Sort.ts
```diff
@@ -353,6 +355,94 @@ function getEloBucket(elo: number): string {
   return `${bucketFloor}-${bucketFloor + 49}`;
 }
 
+//Types - To handle custom card types, we will assume anything before a dash it a type and after is a sub-type.
+//Basically we will not handle custom supertypes since there is no way for us to distinguish a custom supertype from a type
+//Reference: https://mtg.fandom.com/wiki/Type_line
+const splitCardTypes = (card: Card): { typesAndSuperTypes: string[]; subtypes: string[] } => {
+  const split = cardType(card).split(/[-–—]/);
```
KaelenProctor, line 362: Consolidating a bunch of code that was doing this splitting and mapping

src/client/utils/Sort.ts
```diff
@@ -353,6 +355,94 @@ function getEloBucket(elo: number): string {
...
+      },
+      [[], []] as string[][], // Initial value: an array containing two empty arrays
+    );
+  };
+
+  const types = getAllTypesOfCards(cube).filter((type) => type !== 'Land');
+  const [mtgTypes, otherTypes] = splitMtgVersusOtherTypes(types);
+
+  //Keep the old order of MTG types. Custom types will be placed
+  mtgTypes.sort((a, b) => {
```
KaelenProctor, line 422: Keeping the same sort of MTG types for "Types-Multicolor"

PR #2784, opened 2025-11-16
Tag Color Synchronization

KaelenProctor:
# Problem

If a tag was saved with a color, but no cards have that tag anymore, it would appear in the modal forever.

# Changes

1. Make sure Tag in the Tag colors modal only show tags that exist on at least one card
2. When cube changes are committed, recompute the Tag colors using the current information and set of tags on all cards after commit
3. Use the Tag colors from CubeContext instead of maintaining separate state in the modal

Until the tag colors modal is saved the cube's tag colors may still have old tags that no longer matter, but at least they won't be shown forever. On the next tag color modal save, the old tags will be flushed away.

I also tossed in some docker repo changes while I am at it.

### Before

A tag with a color set, would stay in the modal (even after page reload) even when no cards have it
old-tags-with-color-set-stay-in-modal-even-when-no-cards-have-it

Also when new tags were added to cards, the tag colors modal doesn't show them until a page reload
old-saving-tags-doesnt-update-modal

### After

Once the tag no longer exists on any card, it is not seen in the modal
new-tags-with-color-set-gone-from-modal-when-no-cards-have-it

When a new tag is added and committed, it appears in the tag colors modal.
new-tag-added-appears-in-modal-right-away

### DIFF
--- a/packages/docs/setup/docker-setup.md
+++ b/packages/docs/setup/docker-setup.md
@@ -19,20 +19,20 @@ docker compose -f packages/scripts/src/docker-compose.yml -f packages/scripts/sr
 **Note**: This can take tens of minutes to complete. When it finishes successfully you should see in the last docker logs output similar to the following (order of lines, and container name, may be different):
 
 ```
-cubecobralocalstack  | 2025-10-27T21:47:56.503  INFO --- [et.reactor-0] localstack.request.aws     : AWS s3.UploadPart => 200
-cubecobralocalstack  | 2025-10-27T21:47:58.803  INFO --- [et.reactor-1] localstack.request.aws     : AWS s3.CompleteMultipartUpload => 200
-cubecobracube        | Finished comboDict.json
-cubecobracube        | Uploading manifest...
-cubecobralocalstack  | 2025-10-27T21:47:58.811  INFO --- [et.reactor-2] localstack.request.aws     : AWS s3.PutObject => 200
-cubecobracube        | Finished manifest
-cubecobracube        | done
-cubecobracube        | Complete
-cubecobralocalstack  | 2025-10-27T21:47:59.912  INFO --- [ead-7 (_run)] localstack_persist.state   : Persisting state of service s3...
-cubecobracube exited with code 0
+localstack  | 2025-10-27T21:47:56.503  INFO --- [et.reactor-0] localstack.request.aws     : AWS s3.UploadPart => 200
+localstack  | 2025-10-27T21:47:58.803  INFO --- [et.reactor-1] localstack.request.aws     : AWS s3.CompleteMultipartUpload => 200
+cube        | Finished comboDict.json
+cube        | Uploading manifest...
+localstack  | 2025-10-27T21:47:58.811  INFO --- [et.reactor-2] localstack.request.aws     : AWS s3.PutObject => 200
+cube        | Finished manifest
+cube        | done
+cube        | Complete
+localstack  | 2025-10-27T21:47:59.912  INFO --- [ead-7 (_run)] localstack_persist.state   : Persisting state of service s3...
+cube exited with code 0
 Aborting on container exit...
 [+] Stopping 2/2
- ✔ Container cubecobracube        Stopped                                                                          0.0s
- ✔ Container cubecobralocalstack  Stopped
+ ✔ Container cube        Stopped                                                                          0.0s
+ ✔ Container localstack  Stopped
 ```
 
 ## First-Time Data Download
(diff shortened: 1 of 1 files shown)


packages/client/src/components/modals/TagColorsModal.tsx
```diff
@@ -60,16 +60,16 @@ interface TagColorsModalProps {
 
 const TagColorsModal: React.FC<TagColorsModalProps> = ({ isOpen, setOpen }) => {
   const { csrfFetch } = useContext(CSRFContext);
+  //Continually update the tagColors in CubeContext as the modal contents change. They won't persist until the actual save
   const { tagColors, setTagColors, showTagColors, updateShowTagColors, canEdit, cube } = useContext(CubeContext);
   const [loading, setLoading] = useState(false);
-  const [modalColors, setModalColors] = useState([...tagColors]);
```
KaelenProctor, line 65: Basically the functional change, not keeping a separate state that doesn't change when cube changes are committed

packages/client/src/contexts/CubeContext.tsx
```diff
@@ -348,14 +348,24 @@ export function CubeContextProvider({
...
+    return getAllTags(cube.cards);
   }, [cube.cards]);
 
-  const [tagColors, setTagColors] = useState<TagColor[]>([
-    ...(cube.tagColors || []),
-    ...allTags
-      .filter((tag) => !(cube.tagColors || []).map((tc) => tc.tag).includes(tag))
-      .map((tag) => ({ tag, color: 'no-color' })),
-  ]);
+  const computeTagColors = (tagColors: TagColor[], allTags: string[]) => {
```
KaelenProctor, line 357: Moved into a function and also filtered out tags in cube tagColors when that tag is no longer on any cards

packages/client/src/contexts/CubeContext.tsx
```diff
@@ -348,14 +348,24 @@ export function CubeContextProvider({
...
+    //Filter out tags in cube tag colors that are no longer tags on cards
+    const tagColorsWithActiveTags = cubeTagColors.filter((tc) => allTags.includes(tc.tag));
+    //If all tags on cards, is not within the cubeTagColors, then add it to the set with no color
+    const newTagsWithoutColorsYet = allTags
+      .filter((tag) => !cubeTagColors.map((tc) => tc.tag).includes(tag))
+      .map((tag) => ({ tag, color: 'no-color' }));
+    return [...tagColorsWithActiveTags, ...newTagsWithoutColorsYet];
+  };
+
+  const [tagColors, setTagColors] = useState<TagColor[]>(() => computeTagColors(cube.tagColors, allTags));
```
KaelenProctor, line 368: Compute the value one as the initial tagColors state

packages/docs/setup/docker-setup.md
```diff
@@ -19,20 +19,20 @@ docker compose -f packages/scripts/src/docker-compose.yml -f packages/scripts/sr
...
-cubecobralocalstack  | 2025-10-27T21:47:58.803  INFO --- [et.reactor-1] localstack.request.aws     : AWS s3.CompleteMultipartUpload => 200
-cubecobracube        | Finished comboDict.json
-cubecobracube        | Uploading manifest...
-cubecobralocalstack  | 2025-10-27T21:47:58.811  INFO --- [et.reactor-2] localstack.request.aws     : AWS s3.PutObject => 200
-cubecobracube        | Finished manifest
-cubecobracube        | done
-cubecobracube        | Complete
-cubecobralocalstack  | 2025-10-27T21:47:59.912  INFO --- [ead-7 (_run)] localstack_persist.state   : Persisting state of service s3...
-cubecobracube exited with code 0
+localstack  | 2025-10-27T21:47:56.503  INFO --- [et.reactor-0] localstack.request.aws     : AWS s3.UploadPart => 200
```
KaelenProctor, line 22: Ended up removing the whole COMPOSE_PROJECT_NAME so that the container and volume names are fixed

packages/docs/setup/docker-setup.md
```diff
@@ -71,16 +71,16 @@ Once setup is complete, the containers will exit gracefully. The card databases
 After initial setup, start the program with:
 
 ```sh
-docker compose up -f packages/scripts/src/docker-compose.yml
+docker compose -f packages/scripts/src/docker-compose.yml up
```
KaelenProctor, line 74: up must be after files

packages/docs/setup/docker-setup.md
```diff
@@ -19,20 +19,20 @@ docker compose -f packages/scripts/src/docker-compose.yml -f packages/scripts/sr
...
+cube exited with code 0
 Aborting on container exit...
 [+] Stopping 2/2
- ✔ Container cubecobracube        Stopped                                                                          0.0s
- ✔ Container cubecobralocalstack  Stopped
+ ✔ Container cube        Stopped                                                                          0.0s
+ ✔ Container localstack  Stopped
 ```
 
 ## First-Time Data Download
```
KaelenProctor, line 38: In order to do this the user needs access key and secret in their .env file. How would they get those?

PR #2790, opened 2025-12-13
Support custom cards in decklist upload

KaelenProctor:
From Discord bug report

# Changes

1. Check if the card name matches a custom card name in the cube
2. Add tests (Copilot written)
3. Adjust jest coverage so it doesn't run coverage on client copied dir
4. Adjust ts config so it covers test directory too (otherwise aliases like serverutils confuse VSCode)

# Testing
Using the follow text for the import per the Discord report, where both "The Wheel of Time" and "Ray of Suppression" are custom cards (yes the latter is odd)
```
The Wheel of Time
Island
Island
3x Swamp
Plains
Ray of Suppression
Magar of the Magic Strings
```

## Before

Neither custom card is in the deck
image

## After

Both custom cards are imported
new-upload-deck-with-custom-card

### DIFF
--- a/packages/server/jest.config.ts
+++ b/packages/server/jest.config.ts
@@ -45,7 +45,7 @@ const jestConfig: Config = {
   collectCoverage: true,
   coverageDirectory: 'coverage',
   coverageReporters: ['json', 'text'],
-  collectCoverageFrom: ['src/**/*.{js,ts,tsx}', 'src/dynamo/models/*.js', '!src/**/*.d.ts', '!src/**/index.{js,ts}'],
+  collectCoverageFrom: ['src/**/*.{js,ts,tsx}', '!src/**/*.d.ts', '!src/**/index.{js,ts}', '!src/client/**'],
 };
 
 export default jestConfig;
--- a/packages/server/tsconfig.json
+++ b/packages/server/tsconfig.json
@@ -31,5 +31,5 @@
     "noFallthroughCasesInSwitch": true
   },
   "exclude": ["node_modules", "src/client"],
-  "include": ["src/**/*.js", "src/**/*.ts", "src/**/*.tsx", "src", "../utils/src", "jest.setup.ts", "jest.config.ts"]
+  "include": ["src/**/*.js", "src/**/*.ts", "src/**/*.tsx", "src", "../utils/src", "jest.setup.ts", "jest.config.ts", "test"]
 }


packages/server/jest.config.ts
```diff
@@ -45,7 +45,7 @@ const jestConfig: Config = {
   collectCoverage: true,
   coverageDirectory: 'coverage',
   coverageReporters: ['json', 'text'],
-  collectCoverageFrom: ['src/**/*.{js,ts,tsx}', 'src/dynamo/models/*.js', '!src/**/*.d.ts', '!src/**/index.{js,ts}'],
+  collectCoverageFrom: ['src/**/*.{js,ts,tsx}', '!src/**/*.d.ts', '!src/**/index.{js,ts}', '!src/client/**'],
```
KaelenProctor, line 48: Dynamo models are no longer JS. Ignore client files from coverage

packages/server/src/dynamo/models/cube.ts
```diff
@@ -139,7 +139,7 @@ const stripDetails = (cards: any[]): void => {
   });
 };
 
-const getCards = async (id: string): Promise<any> => {
+const getCards = async (id: string): Promise<CubeCards> => {
```
KaelenProctor, line 142: Minor improvement in typing

packages/server/src/router/routes/cube/deck.ts
```diff
@@ -667,6 +665,19 @@ export const uploadDecklistHandler = async (req: Request, res: Response) => {
       added.push([]);
     }
 
+    const cubeCards = await cubeDao.getCards(cube.id);
```
KaelenProctor, line 668: Moved down for tests, but also slightly more efficient to check ownership first

packages/server/src/router/routes/cube/deck.ts
```diff
@@ -667,6 +665,19 @@ export const uploadDecklistHandler = async (req: Request, res: Response) => {
...
+    const { mainboard } = cubeCards;
+
+    const customCardNameMap = mainboard
```
KaelenProctor, line 671: Build map of custom card names in the Cube, to use in lookups if the name isn't found in the card database

packages/server/src/router/routes/cube/deck.ts
```diff
@@ -681,7 +692,7 @@ export const uploadDecklistHandler = async (req: Request, res: Response) => {
       } else {
         let selected = null;
         // does not have set info
-        const normalizedName = cardutil.normalizeName(item);
+        const normalizedName = normalizeName(item);
```
KaelenProctor, line 695: This was moving to best practice imports instead of *

PR #2796, opened 2025-12-22
Create the consolidated dynamo table using AWS CDK (cdklocal)

KaelenProctor:
# Changes

1. Define a local stack for the CDK to use in the local docker setup
2. Update the localstack image to latest to work with the CDK and perform the CDK deploy during init
3. Misc fixes to migration scripts as I ran into problems updating my local

Given the local setup only requires the dynamo table I created its own stack, but I could also see refactoring such that each component could define whether it is needed for the environment. Though that might end up more complex since there are some inter-dependencies such as beanstalk using the dynamo table name (not from the config)

# Challenges

I tried to use the CDK stack as they exist but ran into many problems. Some were fixed by updating the localstack version, but others were not viable to figure out such as:
1. Could not figure out how to setup a local hosted zone so the lookup could work
2. Deploying the lambda requires a valid zip in S3 which to create it outside the CDK scope
3. ElasticBeanstalk tried to figure out the load balancer but couldn't

# Testing

1. Run the cdk from my linux environment to bootstrap and deploy: `cdklocal bootstrap --context environment=local --context version=1.0.0` then `cdklocal deploy --context environment=local --context version=1.0.0`
2. Clean that from the localstack persisted data
3. Run the docker init steps and validate the CDK ran
4. Validate the LOCAL_CUBECOBRA dynamo table exists
5. Run all the migration scripts

### DIFF
--- a/packages/server/src/dynamo/dao/DraftDynamoDao.ts
+++ b/packages/server/src/dynamo/dao/DraftDynamoDao.ts
@@ -923,12 +923,18 @@ export class DraftDynamoDao extends BaseDynamoDao<Draft, UnhydratedDraft> {
       }
     }
 
-    return cards.map((card: any) => ({
-      ...card,
-      details: {
-        ...cardFromId(card.cardID),
-      },
-    }));
+    if (!cards) {
+      return [];
+    }
+
+    return cards
+      .filter((c: any) => c !== null)
+      .map((card: any) => ({
+        ...card,
+        details: {
+          ...cardFromId(card.cardID),
+        },
+      }));
   }
 
   /**


packages/cdk/cdk.context.json
```diff
@@ -3,5 +3,21 @@
     "Id": "/hostedzone/Z5BLGHAZY2IZ8",
     "Name": "cubecobra.com."
   },
-  "availability-zones:account=816705121310:region=us-east-2": ["us-east-2a", "us-east-2b", "us-east-2c"]
+  "availability-zones:account=816705121310:region=us-east-2": [
+    "us-east-2a",
+    "us-east-2b",
+    "us-east-2c"
+  ],
+  "acknowledged-issue-numbers": [
```
KaelenProctor, line 11: Mostly the acknowledged issue and turning off cli-telemetry are nice, so that people don't get weird warning messages

packages/cdk/config.ts
```diff
@@ -15,6 +15,22 @@ export interface EnvironmentConfiguration {
...
+    stackName: 'CubeCobraLocalStack',
+    account: '000000000000',
+    region: 'us-east-1',
+    dataBucket: 'local',
+    appBucket: 'app-code',
+    domain: 'localhost',
+    awsLogGroup: 'CUBECOBRA',
+    awsLogStream: 'DEVELOPMENT',
+    downTimeActive: false,
+    dynamoPrefix: 'LOCAL',
```
KaelenProctor, line 28: Kinda weird that dynamoPrefix is used in the CDK but we don't use it in the .env anymore

packages/scripts/src/docker-compose.yml
```diff
@@ -29,21 +29,28 @@ services:
...
     depends_on:
       - localstack
+    # Force DNS via the localstack container
+    dns:
+      - 10.5.0.5
   localstack:
     # Localstack community does not persist anything out of the box, that is available in localstack-pro.
     # This variant (https://hub.docker.com/r/gresau/localstack-persist) causes the state to be persisted on actions and reloaded when the container starts
-    image: gresau/localstack-persist:3.8
+    image: gresau/localstack-persist:4.12
```
KaelenProctor, line 38: Version 3.8 has a bunch of problems with the CDK such as failing to create IAM roles. As well ended up also having to set PROVIDER_OVERRIDE_CLOUDFORMATION to the legacy engine because otherwise the cloudformation state failed to load, so after a restart the CDK would think it has to deploy again (but the resources existed)

packages/scripts/src/docker/init.sh
```diff
@@ -15,8 +15,9 @@ echo "Building the server"
 npm run build --workspace=packages/server
 # Given the time it takes to do npm install/build, we assume localstack has started and completed its init to setup the S3 bucket
 npm run setup:local:files --workspace=packages/scripts
-npm run setup:local:db --workspace=packages/scripts
```
KaelenProctor, line 18: setup:local:db replaced by the CDK

packages/scripts/src/docker/localstack-init.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+AWS_REGION=us-east-1
```
KaelenProctor, line 3: I don't think this ended up being necessary, but I added it when I was having a bunch of issues validating whether the cdk actions were completed

packages/server/src/dynamo/dao/DraftDynamoDao.ts
```diff
@@ -923,12 +923,18 @@ export class DraftDynamoDao extends BaseDynamoDao<Draft, UnhydratedDraft> {
...
-      details: {
-        ...cardFromId(card.cardID),
-      },
-    }));
+    if (!cards) {
+      return [];
+    }
+
+    return cards
+      .filter((c: any) => c !== null)
```
KaelenProctor, line 931: Likely this is irrelevant given the production migration is done, but for some reason I had at least one draft where most of the cards were null

PR #2812, opened 2026-01-17
move card updates to ECS

dekkerglen:
- Adds ECS task definition
- Adds pipeline steps to publish images
- Adds lambda to poll for scryfall bulk data changes and orchestrate ECS task
- Updates server data file polling to use manifest to reduce data staleness
- Adds Card Update page to view status of card import pipeline


JOBS_CONTAINERIZATION.md
```diff
@@ -0,0 +1,227 @@
...
+
+This implementation containerizes the CubeCobra jobs execution, moving from local npm scripts to ECS Fargate tasks with CDK-managed deployment.
+
+## What Was Changed
+
+### 1. Docker Configuration
+
+**File: `packages/jobs/Dockerfile`**
+
+- Multi-stage build using Node.js 20.18.0
```
KaelenProctor: The app uses node 22 and so does the task, so this is out of date

packages/cardUpdateMonitorLambda/publish.js
```diff
@@ -0,0 +1,59 @@
+const { S3Client, PutObjectCommand } = require('@aws-sdk/client-s3');
```
KaelenProctor, line 1: Any reason to be .js instead of .ts?

packages/cdk/lib/card-update-monitor-lambda.ts
```diff
@@ -0,0 +1,107 @@
...
+        actions: ['ecs:RunTask', 'ecs:DescribeTasks'],
+        resources: ['*'],
+      }),
+    );
+
+    // Grant permission to pass the ECS task role and execution role
+    executionRole.addToPolicy(
+      new iam.PolicyStatement({
+        actions: ['iam:PassRole'],
+        resources: ['*'],
```
KaelenProctor: nit: Probably should be specific about which roles can be passed and which ECS tasks can be run. Least privilege actions

packages/jobs/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM node:20.18.0
```
KaelenProctor: Should use Node 22 to align with everything (the local docker uses 22.11.0)

JOBS_CONTAINERIZATION.md
dekkerglen: Oh I see what you mean - I think before I merge this I'm going to expose some public files so we don't have to run this process for local development

packages/cardUpdateMonitorLambda/publish.js
dekkerglen: should probably update it to be TS but it isn't a super big deal for this case

PR #2815, opened 2026-01-22
Update draft name on save so that it matches the first seat's colors.

KaelenProctor:
# Problem

In the [old dynamo code for drafts](https://github.com/dekkerglen/CubeCobra/blob/a768a76159f01bbe9aa3246e00e51d4b50c3d49f/src/dynamo/models/draft.js#L426) the put() function did the color assessment, setting both the seat and draft names. When the [draft finishes the put() method is called](https://github.com/dekkerglen/CubeCobra/blob/a768a76159f01bbe9aa3246e00e51d4b50c3d49f/src/router/routes/draft/finish.ts#L160). In the new code it calls the update() method when the draft finishes, which didn't update the draft name with the newly assessed colors.

# Testing

## Before

BR draft deck named C for colorless
old-draft-color-C

But the seat names have correct colors
old-seat-names-are-accurate

## After

WU draft deck accurately named
new-seat-names-match-draft

Matching the seats names
new-draft-color-is-set


packages/server/src/dynamo/dao/DraftDynamoDao.ts
```diff
@@ -1014,6 +1019,19 @@ export class DraftDynamoDao extends BaseDynamoDao<Draft, UnhydratedDraft> {
     return colorKeysFiltered;
   }
 
+  private getDeckColors(
+    draft: Pick<DraftType, 'seats' | 'type' | 'cards'>,
```
KaelenProctor, line 1025: I dislike this use of Pick. createDraft uses a custom type similar to Draft except with lots of optional, so this was the easiest solution.

dekkerglen:
Can you rebase to fix the merge conflicts?

Also, the integration tests are failing but it's not the fault of this PR, I am happy to merge regardless. The fix is just to have the write sample mocked files script write {} to cube embeddings

KaelenProctor:
Will do. I honestly have no idea how there are merge conflicts as I rebased
multiple times =(

PR #2818, opened 2026-01-25
Add user option for sorting Your cubes by alphabetical.

KaelenProctor:
# Testing

Sorting by last updated date
old-sorting-by-date

## After

Options on profile. Defaults to last updated as part of the hydration of User Daos
new-your-cubes-order

Saving sort order
new-sort-order-saved

Sorting by last updated as before
new-cubes-sorted-by-last-updated

Sorting by alpha
new-cubes-sorted-by-alpha

Also applies sort on your profile
new-alpha-sorting-on-your-profile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ yarn-error.log*
 
 # migration checkpoints
 .migration-checkpoints/
+
+# clinic profiling output
+.clinic
--- a/package-lock.json
+++ b/package-lock.json
@@ -90,6 +90,13 @@
       "integrity": "sha512-JNAzZcXrCt42VGLuYz0zfAzDfAvJWW6AfYlDBQyDV5DClI2m5sAmK+OIO7s59XfsRsWHp02jAJrRadPRGTt6SQ==",
       "license": "ISC"
     },
+    "node_modules/@assemblyscript/loader": {
+      "version": "0.19.23",
+      "resolved": "https://registry.npmjs.org/@assemblyscript/loader/-/loader-0.19.23.tgz",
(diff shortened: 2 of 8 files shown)


packages/server/src/dynamo/dao/UserDynamoDao.ts
```diff
@@ -164,6 +166,10 @@ export class UserDynamoDao extends BaseDynamoDao<UserWithBaseFields, StoredUserW
       hydrated.gridTightness = DefaultGridTightnessPreference;
     }
 
+    if (!hydrated.yourCubesSortOrder) {
+      hydrated.yourCubesSortOrder = DefaultYourCubesSortOrder;
```
KaelenProctor, line 170: Load the default value if the property isn't set, keeps thinks simple

packages/server/src/router/routes/user/view.ts
```diff
@@ -18,7 +18,11 @@ export const handler = async (req: Request, res: Response) => {
       return redirect(req, res, '/404');
     }
 
-    const cubes = (await cubeDao.queryByOwner(user.id)).items.filter((cube: any) => isCubeListed(cube, req.user));
+    const { sort, ascending } = getCubesSortValues(user);
+
+    const cubes = (await cubeDao.queryByOwner(user.id, sort, ascending)).items.filter((cube: any) =>
```
KaelenProctor, line 23: I think it seems logical to use the user's sort preference on their profile page, even when others look. But if that isn't desired I won't object to removing it.

dekkerglen: I agree

PR #2819, opened 2026-01-25
CSS consistency cleanup

KaelenProctor:
# Problem

Running locally always changed the public stylesheet in server package, so before committing I always had to restore to clean it. I thought it was small lint issues at first, but then I realised the main changes were in the browser specific prefixes like -web-kit, pointing me to the prod vs dev browserlist configs, and validated by running `NODE_ENV=production npm --workspace=packages/client run tailwind:build` which showed no changes.

Secondly prettier the file after build so we should not have to deal with lint issues in pipelines.

### DIFF
--- a/packages/server/public/css/stylesheet.css
+++ b/packages/server/public/css/stylesheet.css
@@ -1,4 +1,6 @@
-*, ::before, ::after {
+*,
+::before,
+::after {
   --tw-border-spacing-x: 0;
   --tw-border-spacing-y: 0;
   --tw-translate-x: 0;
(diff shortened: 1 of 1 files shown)


packages/client/package.json
```diff
@@ -52,7 +52,7 @@
   },
   "scripts": {
     "tailwind": "npx postcss ./src/css/stylesheet.css -o ../server/public/css/stylesheet.css --watch",
-    "tailwind:build": "npx postcss ./src/css/stylesheet.css -o ../server/public/css/stylesheet.css",
+    "tailwind:build": "npx postcss ./src/css/stylesheet.css -o ../server/public/css/stylesheet.css && npx prettier --write ../server/public/css/stylesheet.css",
```
KaelenProctor, line 55: technically there is a postcss plugin for prettier to do this, but simple works

PR #2837, opened 2026-02-09
Filter cards with flavor names from the "reasonable" check

KaelenProctor:
# Problem

Reported in Discord https://discord.com/channels/592787488523943937/1318662199333355571 that recommender shows “La danza del pueblo" instead of Expressive Iteration. Looking into it that is because that Secret Lair isn't a promo or a Universes Beyond type, so it is considered "reasonable". Also even though the name is not English, the card has lang=en =(

# Solution

Cards with a flavor name should not be considered reasonable given they imply a non-standard card.

Per scryfall there are [66 cards with a flavor name and aren't universes beyond](https://scryfall.com/search?as=grid&order=name&q=%28game%3Apaper%29+-is%3Auniversesbeyond+is%3Aflavorname+prefer%3Abest), of which only [3 are promos](https://scryfall.com/search?q=%28game%3Apaper%29+-is%3Auniversesbeyond+is%3Aflavorname+prefer%3Abest+is%3Apromo&unique=cards&as=grid&order=name).
[327 cards have a flavor name and are UB](https://scryfall.com/search?q=%28game%3Apaper%29+is%3Auniversesbeyond+is%3Aflavorname+prefer%3Abest+&unique=cards&as=grid&order=name)

# Other

This also gets recommender working with the docker setup.

# Testing

## Before

Using first printing preference it is nice
old-ei-recommendation-with-first-printing

But with most recent getting the non-english card
old-ei-recommendation-with-most-recent-printing

## After

In the card data, the SDL from Avatar does not have a flavour name. It isn't reasonable because it is Universes Beyond though
new-ei-2305-no-flavour-name

Now “La danza del pueblo" has a flavor name in the data
new-ei-2060-has-flavour-name

And then recommender skips it and uses the Tarkir Commander printing
new-ei-most-recent-non-promo-print

Also tested in card search. Was able to get EI with first and most recent printing, and also old school flavor cards like Brokkos, Apex of Forever vs Bio-Quartz Spacegodzilla


packages/recommenderService/src/mlutils/downloadModel.ts
```diff
@@ -1,16 +1,24 @@
...
+import { fromNodeProviderChain } from '@aws-sdk/credential-providers';
+import dotenv from 'dotenv';
 import fs from 'fs';
 import path from 'path';
 import { Readable } from 'stream';
-
-import 'dotenv/config';
+dotenv.config();
 
 const publicS3 = new S3Client({
```
KaelenProctor, line 9: TODO: Consolidate s3 logic into utils package

packages/recommenderService/.env.example
```diff
@@ -4,7 +4,7 @@ NODE_ENV=development
 
 # AWS Configuration
 AWS_REGION=us-east-2
-DATA_BUCKET=cubecobra-data
+DATA_BUCKET=cubecobra-public
```
KaelenProctor, line 7: This fixes the download of the models

packages/recommenderService/package.json
```diff
@@ -43,6 +43,7 @@
     "@types/archiver": "^6.0.2",
     "@types/compression": "^1.8.1",
     "@types/response-time": "^2.3.8",
+    "@types/uuid": "^10.0.0",
```
KaelenProctor, line 46: Missing types prevented recommender running in dev using ts-node

packages/server/.env_EXAMPLE
```diff
@@ -28,4 +28,4 @@ HTTP_ONLY="true"
 STRIPE_SECRET_KEY="apikey"
 LOCALSTACK_SES="true"
 ENABLE_BOT_SECURITY="true"
-ML_SERVICE_URL="http://localhost:5002"
+ML_SERVICE_URL="http://recommender:5002"
```
KaelenProctor, line 31: Direct the calls to the recommender docker container

yarn.lock
```diff
@@ -31,11 +31,6 @@
     "@csstools/css-tokenizer" "^3.0.3"
     lru-cache "^10.4.3"
 
-"@assemblyscript/loader@^0.19.21":
```
KaelenProctor, line 34: Can I just delete yarn lock?

dekkerglen: I think so

PR #2856, opened 2026-03-01
Set DndContext around the DeckStacks which is required.

KaelenProctor:
# Problem

Page error occurs from the JS code

Fixes https://github.com/dekkerglen/CubeCobra/issues/2860

# Solution

Set DndContext around the DeckStacks which is required.

Use logic from CubeDeckBuilderPage to implement the drag and drop within each player's picks.

# Testing

## After

new-grid-draft-dnd


KaelenProctor:
Flaky tests

PR #2862, opened 2026-03-16
Feature/collaborate

Eetai:


packages/client/src/components/settings/CollaboratorsSettings.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const [loading, setLoading] = useState(false);
+  const [fetchLoading, setFetchLoading] = useState(true);
+  const [collaborators, setCollaborators] = useState<CollaboratorEntry[]>([]);
+  const [confirmRemoveId, setConfirmRemoveId] = useState<string | null>(null);
+
+  // Fetch collaborator details from server on mount
+  useEffect(() => {
+    if (!cube.id) return;
+    setFetchLoading(true);
+    fetch(`/cube/api/collaborators/${cube.id}`)
```
KaelenProctor: To be consistent it would be great to use await instead of promise chaining. Reduces boilerplate and makes the code more readable imo

packages/client/src/components/settings/CollaboratorsSettings.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const handleAdd = useCallback(async () => {
+    const username = usernameInput.trim();
+    if (!username) return;
+
+    setLoading(true);
+    setError(null);
+    setSuccess(null);
+
+    try {
+      const res = await fetch(`/cube/api/collaborators/${cube.id}/add`, {
```
KaelenProctor: Please use csrfFetch so the CSRF logic is centralized

packages/client/src/components/settings/CollaboratorsSettings.tsx
```diff
@@ -0,0 +1,196 @@
...
+              Loading…
+            </Text>
+          ) : collaborators.length === 0 ? (
+            <Text sm className="text-text-secondary italic">
+              No collaborators yet.
+            </Text>
+          ) : (
+            <Flexbox direction="col" gap="2">
+              {collaborators.map((c) => (
+                <Flexbox key={c.id} direction="row" alignItems="center" gap="2" className="py-1">
```
KaelenProctor, line 138: Would using `<UserPreview>` work here? It looks similar but I could be wrong whether this is a grid of users or a table like view

packages/client/src/pages/CubeSettingsPage.tsx
```diff
@@ -33,9 +35,15 @@ interface CubeSettingsPageProps {
   pricePurchase: number | null;
 }
 
+const OWNER_ONLY_VIEWS = new Set(['overview', 'options']);
+
 const CubeSettingsPageContent: React.FC<CubeSettingsPageProps> = ({ cube, cards, versions }) => {
   const settingsViewContext = useContext(SettingsViewContext);
-  const view = settingsViewContext?.view || 'overview';
+  const user = useContext(UserContext);
+  const isOwner = !!user && cube.owner?.id === user.id;
```
KaelenProctor: Can this be grabbed from CubeContext to keep the logic in one place?

packages/server/src/router/routes/cube/api/collaborators.ts
```diff
@@ -0,0 +1,158 @@
...
+      return res.status(400).json({ success: 'false', message: 'Cube ID is required' });
+    }
+    const cube = await cubeDao.getById(req.params.id);
+
+    if (!cube || !isCubeViewable(cube, req.user)) {
+      return res.status(404).json({ success: 'false', message: 'Cube not found' });
+    }
+
+    const collaboratorIds = cube.collaborators ?? [];
+    const users = await Promise.all(collaboratorIds.map((id) => userDao.getById(id)));
```
KaelenProctor: userDao.batchGet() would work here. It would be slightly more optimal since it de-dupes the IDs before calling dynamo

packages/server/src/router/routes/cube/api/collaborators.ts
```diff
@@ -0,0 +1,158 @@
...
+      return res.status(400).json({ success: 'false', message: 'Cube ID is required' });
+    }
+    const cube = await cubeDao.getById(req.params.id);
+
+    if (!cube || !isCubeViewable(cube, req.user)) {
+      return res.status(404).json({ success: 'false', message: 'Cube not found' });
+    }
+
+    // Only the owner can add collaborators
+    if (!req.user || cube.owner.id !== req.user.id) {
```
KaelenProctor: Nice to have: Use isCubeOwner, which means it has to move into utils from client

packages/server/src/router/routes/cube/api/collaborators.ts
```diff
@@ -0,0 +1,158 @@
...
+      return res.status(404).json({ success: 'false', message: 'Cube not found' });
+    }
+
+    const userId = req.params.userId ?? '';
+    const isOwner = req.user && cube.owner.id === req.user.id;
+    const isSelf = req.user && req.user.id === userId;
+
+    // Owner can remove anyone; a collaborator can only remove themselves
+    if (!isOwner && !isSelf) {
+      return res.status(403).json({ success: 'false', message: 'Only the cube owner can remove collaborators' });
```
KaelenProctor:
```suggestion
      return res.status(403).json({ success: 'false', message: 'Only the cube owner can remove other collaborators' });
```

packages/server/src/router/routes/cube/api/collaborators.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    if (!cube.collaborators.includes(userId)) {
```
KaelenProctor, line 125: tbh I could see returning a success since in effect the user is not a collaborator

packages/server/src/router/routes/cube/api/customsorts.ts
```diff
@@ -23,7 +23,7 @@ export const customsortsHandler = async (req: Request, res: Response) => {
       });
     }
 
-    if (!req.user || cube.owner.id !== req.user.id) {
+    if (!isCubeEditable(cube, req.user)) {
```
KaelenProctor, line 26: nice consolidation!

packages/server/src/serverutils/cubefn.ts
```diff
@@ -513,15 +513,26 @@ function isCubeViewable(cube: any, user: any): boolean {
     return true;
   }
 
-  return user && (cube.owner.id === user.id || util.isAdmin(user));
+  if (!user) return false;
```
KaelenProctor: Personally i'd like the brackets like in isCubeEditable, I find that much easier to read

packages/client/src/components/settings/CollaboratorsSettings.tsx
Eetai: I think what we have is a better fit for what we're doing here

packages/server/src/router/routes/cube/api/collaborators.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (!req.params.id) {
+      return res.status(400).json({ success: 'false', message: 'Cube ID is required' });
+    }
+    const cube = await cubeDao.getById(req.params.id);
+
+    if (!cube || !isCubeViewable(cube, req.user)) {
+      return res.status(404).json({ success: 'false', message: 'Cube not found' });
+    }
+
+    const collaboratorIds = cube.collaborators ?? [];
```
KaelenProctor, line 25: An additional thought I had is how this should handle banned or deleted users? I presume based on the data model it would not be easy to remove people from collaborators when they are banned/deleted, so it might have to filter those on read or in the update calls.
Could be a very edge case concern

PR #2864, opened 2026-03-24
Apply the size filter even for single hash queries.

KaelenProctor:
Copilot code

# Testing

## Context

The amazingly unique cubes I have on my local, at least with different sizes
image

## Before

Filter not applying on prod https://cubecobra.com/search?q=size%3C400
image

## After

Size filter works
new-size-filter-works-by-itself

I don't have good enough cube hashes on my local to show it working with other filters too


KaelenProctor:
Build fails seem unrelated, I can take a look at fixing separately

PR #2895, opened 2026-04-21
fixed small cube search bug

sudokufan:
searching by size had buggy implementation: rather than returning all cubes regardless of size, size=# returns matching cubes, but the results only display by page.                        
                                                                                
For example Size=180, sorted by popularity, only returns the 180 card cubes on the first page of all popular cubes (3 in this case). Clicking Next Page adds the matching cubes from the next page, and so forth. The larger problem is that, if a cube size isn't on the front page (such as with Size=96), it will return no results with no option to go to the next page.     

**Root Cause**:
                                                                                
In queryByHashWithSort (CubeDynamoDao.ts:1689), the size filter (filterByCardCount) was applied after fetching a fixed page of results from DynamoDB. The flow was:                                                       
                                                                  
1. DynamoDB returns 36 items (the Limit)
2. filterByCardCount removes non-matching cubes, potentially leaving 0-3 results                                                                       
3. The lastKey cursor still advances past those 36 items        
4. The client receives few/no results per page, explaining both symptoms:
- Size=180: Only 3 matching cubes on the first DynamoDB page are returned, more appear when clicking "Next" (which fetches the next DynamoDB page)       
- Size=96: Zero matching cubes on the first DynamoDB page → "No Results" with no pagination                                                            
                                                                  
**Fix (two changes in CubeDynamoDao.ts)**:
                                                                  
1. queryByHashWithSort now loops through multiple DynamoDB pages when a cardCountFilter is active, accumulating results until it has enough (up to limit) or exhausts the data. Also fetches 3x the limit per DynamoDB page when filtering to reduce round trips

2. queryByMultipleHashes now has a fix for an early-return bug where the method would return immediately after finding hash-matching cubes even if filterByCardCount eliminated all of them. Now continues paging when the card count filter is active and no results have accumulated yet

### DIFF
--- a/packages/server/src/dynamo/dao/CubeDynamoDao.ts
+++ b/packages/server/src/dynamo/dao/CubeDynamoDao.ts
@@ -1614,21 +1614,21 @@ export class CubeDynamoDao extends BaseDynamoDao<Cube, UnhydratedCube> {
         const filteredCubes = this.filterByCardCount(cubes, cardCountFilter);
 
         matchingCubes.push(...filteredCubes);
-
-        // Return with pagination key for next query
-        // Note: Results are already sorted by the GSI query
-        return {
-          items: matchingCubes,
-          lastKey: candidatesResult.lastKey,
-        };
       }
 
-      // No matches in this page, continue to next page
       currentLastKey = candidatesResult.lastKey;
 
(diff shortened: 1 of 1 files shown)


sudokufan:
@dekkerglen 👀

PR #2911, opened 2026-05-11
misc: revert auto opening cube edit pane

haganbmj:
See: https://x.com/wtwlf123/status/2053646653320822804


dekkerglen:
Won't merge as-is because I think it is very important that the new user workflow for getting started with a cube is as easy as possible. so let's do this instead: if the cube is empty, open up the edit sidebar, otherwise only open it by default if there are local pending changes. Also there are merge conflicts

dekkerglen:
I've fixed this myself

haganbmj:
Ah sorry, my github notifications are a complete mess due to work so I'm terrible at catching replies. Sounds like a good solve to me.

PR #2913, opened 2026-05-14
Fix styling issue for Mix Boards button

iammoonman:


dekkerglen:
Thanks for the follow up!

PR #2944, opened 2026-06-11
fix: reduce complexity of isReasonable check

haganbmj:
I think it's probably preferable to have promo match scryfall where possible, that'll be more intuitive. From there that lets isReasonable have more control over defining its own conditions. I still think the primary issue here, though, is that you can't do this in isolation. It's possible to have an unreasonable card be the _only_ option. We might instead just need to rely on filtering out known bad things, like tokens and art_series cards, since those should never be normal cube inclusions.

### DIFF
--- a/packages/jobs/src/update_cards.ts
+++ b/packages/jobs/src/update_cards.ts
@@ -522,11 +522,12 @@ function convertCard(
   newcard.popularity = 0;
   newcard.cubeCount = 0;
   newcard.pickCount = 0;
-  
+
   // Mark component=meld_result as extra.
-  const isMeldResult = card.layout === 'meld' && card.all_parts?.some((part) => part.id === card.id && part.component === 'meld_result');
+  const isMeldResult =
+    card.layout === 'meld' && card.all_parts?.some((part) => part.id === card.id && part.component === 'meld_result');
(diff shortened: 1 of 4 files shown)


packages/jobs/src/update_cards.ts
```diff
@@ -542,16 +546,7 @@ function convertCard(
...
-    card.promo ||
-    (card.frame_effects && card.frame_effects.includes('extendedart')) ||
-    (card.frame_effects && card.frame_effects.includes('showcase')) ||
-    card.textless ||
-    card.frame === 'art_series' ||
-    card.set.toLowerCase() === 'mps' || // kaladesh masterpieces
-    card.set.toLowerCase() === 'mp2' || // invocations
-    card.set.toLowerCase() === 'exp' || // expeditions
-    card.set.toLowerCase() === 'amh1'; // mh1 art cards
+  newcard.promo = card.promo;
```
haganbmj, line 550: This, and the meld change, will result in a number of `promo`/`isExtra` changes.

I'm not familiar enough with how heavy card migrations are, so if this is something we need to avoid or split up let me know.

The goal here is just to remove the set checks that are not up to date, and defer these kind of checks to where they're needed. I think it's probably preferable for these base values to match Scryfall where sensible.

PR #2965, opened 2026-06-28
Fix Record paging errors, and consolidate into IndefinitePaginatedTable

KaelenProctor:
# Problems

1. Records page in Cube and in Import record did not set Content-Type JSON so the backend did not see lastKey. Some other endpoints had the same issue
2. Also their frontend logic for paging had problems (ones already fixed) with being able to past past the end

# Solution
I built on the changes commit by commit so might be easier to review that way. By the end I consolidated the paging logic within IndefinitePaginatedList into a utility class and made IndefinitePaginatedTable as the tabular version.

A bunch of other usages of Pagination likely can be converted to IndefinitePaginatedTable, solving slight logic differences in how the paging/fetching occurs.

# Testing

While testing I set the page size to 2 in frontend and backend since I don't have that many records

## Before

Paging never refreshes the list b/c the backend never got the lastKey. It also takes 2 clicks to get to the next page
old-records-list

## After

Paging works as smooth as possible.
new-import-record-paging-v2

Due to the way dynamodb works, if the number of records is evenly divisible by the page size, there has to be one more click on next page before we know there are no more results.
new-draft-records-v2

# Future

PRs to update other Pagination usages, such as CreateRecordFromDraftModal, to use IndefinitePaginatedTable or List

### DIFF
--- a/packages/client/src/components/IndefinitePaginatedTable.tsx
+++ b/packages/client/src/components/IndefinitePaginatedTable.tsx
@@ -8,31 +8,27 @@ import {
   hasIndefiniteMore,
 } from '../utils/indefinitePagination';
 import { Card, CardBody, CardHeader } from './base/Card';
-import { Col, Flexbox, Row } from './base/Layout';
+import { Flexbox } from './base/Layout';
 import Pagination from './base/Pagination';
+import Table from './base/Table';
 import Text from './base/Text';
 
-interface IndefinitePaginatedListProps<T> {
+interface IndefinitePaginatedTableProps<T> {
   items?: T[];
   setItems: (items: T[]) => void;
   fetchMoreRoute: string;
-  renderItem: (item: T) => React.ReactNode;
+  renderItem: (item: T) => { [key: string]: React.ReactNode };
   lastKey: any;
   setLastKey: (lastKey: any) => void;
   pageSize: number;
   noneMessage?: string;
   itemsKey?: string;
   header: string;
-  xs?: number;
-  sm?: number;
-  md?: number;
-  lg?: number;
-  xl?: number;
-  xxl?: number;
+  headers: string[];
   inCard?: boolean;
 }
 
-const IndefinitePaginatedList = <T,>({
+const IndefinitePaginatedTable = <T,>({
   items,
   setItems,
   fetchMoreRoute,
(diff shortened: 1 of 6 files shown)


packages/client/src/components/IndefinitePaginatedList.tsx
```diff
@@ -35,6 +42,7 @@ const IndefinitePaginatedList = <T,>({
   header,
   pageSize,
   noneMessage = 'No items found.',
+  itemsKey = 'items',
```
KaelenProctor, line 45: Add this option so the component is easier to use with different endpoints

packages/client/src/components/IndefinitePaginatedList.tsx
```diff
@@ -43,56 +51,42 @@ const IndefinitePaginatedList = <T,>({
   xxl,
   inCard = false,
 }: IndefinitePaginatedListProps<T>) => {
-  const { csrfFetch } = useContext(CSRFContext);
+  const { callApi } = useContext(CSRFContext);
```
KaelenProctor, line 54: Small change but callApi always uses JSON content-type and a POST so cannot forget to set the right content type

packages/client/src/components/IndefinitePaginatedTable.tsx
```diff
@@ -0,0 +1,131 @@
...
+          pageCount,
+          setPage,
+          fetchMoreData,
+        });
+      }}
+      loading={loading}
+    />
+  );
+
+  const body =
```
KaelenProctor, line 86: HTML taken from ImportRecordPage/DraftReports

packages/client/src/pages/ImportRecordPage.tsx
```diff
@@ -129,123 +128,75 @@ const SelectRecordStep: React.FC<SelectRecordStepProps> = ({ cubeId, selectedRec
...
-    fetchInitial();
-  }, [csrfFetch, cubeId]);
 
-  const fetchMore = useCallback(async () => {
-    if (loading || !lastKey) return;
-    setLoading(true);
-    try {
-      const response = await csrfFetch(`/cube/records/list/${cubeId}`, {
-        method: 'POST',
-        body: JSON.stringify({ lastKey }),
```
KaelenProctor, line 160: Lack of content-type JSON is one of the core problems

packages/client/src/pages/ImportRecordPage.tsx
```diff
@@ -129,123 +128,75 @@ const SelectRecordStep: React.FC<SelectRecordStepProps> = ({ cubeId, selectedRec
...
-      });
-      const data = await response.json();
-      if (data.records) {
```
KaelenProctor, line 163: As we fixed in IndefinitePaginatedList before, if the response returned no more items we need to update the UI but we wouldn't here. This is more appropriate to be a check of "does the response contain the 'records' prop'

packages/client/src/records/TrophyArchive.tsx
```diff
@@ -34,9 +34,8 @@ const TrophyArchive: React.FC<TrophyArchiveProps> = ({ records, lastKey }) => {
     if (loading || !lastKeyState) return;
     setLoading(true);
     try {
-      const response = await csrfFetch('/api/draftreports', {
-        method: 'POST',
-        body: JSON.stringify({ lastKey: lastKeyState }),
+      const response = await callApi('/api/draftreports', {
```
KaelenProctor, line 37: Another example I found of a POST missing content type, so using callApi instead

packages/client/src/utils/indefinitePagination.ts
```diff
@@ -0,0 +1,79 @@
...
+  try {
+    const response = await callApi(fetchMoreRoute, {
+      lastKey,
+    });
+
+    if (response.ok) {
+      const json = await response.json();
+
+      // eslint-disable-next-line no-restricted-syntax
+      if (json.success === 'true' || itemsKey in json) {
```
KaelenProctor, line 43: I am pretty sure `itemsKey in json` is fine even if banned by ESLint, given the frontend will be compiled from typescript into browser compatible syntax

KaelenProctor, line 43: I updated the logic a bit here so that response can include `success === 'true'` or `itemsKey in json`. Make it easier to use without needing backend change with wouldn't be obvious

PR #2974, opened 2026-07-05
Upgrade to Typescript 6

KaelenProctor:
# Changes

1. Switch typescript to dev dependency in the top-level package
2. Upgrade to typescript 6
3. Run 5 to 6 migration on the tsconfigs
4. Update client tsconfig moduleResolution from node to bundler, to appease the TS gods
    1. Which for server package meant I had to convert await require to imports
    2. Also ts-node hated express-messages types so needed to add a ts-node config option

What I don't understand is why some packages are still using typescript 5.X, as evidence by running `tsc --version` in the package folders and based on the package-lock.json additions for those packages. Still trying to figure out what those are holding it back.
Edit: I've got all the packages but client upgrade to 6 now

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51722,7 +51722,7 @@
         "@types/react": "^18.3.0",
         "@types/react-dom": "^18.3.0",
         "@vitejs/plugin-react": "^4.3.0",
-        "typescript": "^5.7.0",
+        "typescript": "^6.0.3",
         "vite": "^6.0.0"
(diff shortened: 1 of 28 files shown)


KaelenProctor:
Updates:
1. Fixed merge conflicts
2. When through each app and ran `npm run build` or equivalent to validate tsc and/or ts-node worked
3. Ran local app successfully as before

PR #2984, opened 2026-07-25
Allow CDN-hosted card images through the image proxy (fixes #2981)

### LINKED ISSUE #2981: pdf export results in single empty page
**Describe the bug**
Exporting works in text format, but I get a single empty page when exporting pdf.

**To Reproduce**

1. Go to ['my cube'](https://cubecobra.com/cube/list/c8f0bb02-4caf-4cbe-905b-2ffc622d721f)
2. Click on Export
3. Select "Print and Play (.pdf)"
4. Download will produce a 3kb apparently empty pdf.

I tried the two cubes I have, text export seems to work as expected, but the pdf export is consistently broken.

**Expected behavior**
A pdf with one image per card

**Screenshots**
Think i saw some flashing error message but could not catch it on a screenshot

**Desktop:**

- OS: Win 11 Home
- Browser MS Edge 
- Version 150.0.4078.83 64 bit

**Smartphone:**

Haven't tried on my phone.

**Additional context**
Just started using the site, might be something I did wrong, but a better error message would be appreciated.

Chr0maticDev:
Fixing #2981.

The Print and Play feature was always producing an empty PDF with `Failed to convert image: Error: Failed to fetch image` errors for each card image.

The error was occurring in getImageDataUrl in /components/modals/PrintAndPlayExportModal.tsx.

The route `/tool/imageproxy` is used in that function, but that route filters out any URLs that don't start with `https://cards.scryfall.io/`.

However, it seems like the cardImageNormal of each card (card.details?.image_normal) was changed in [this commit](https://github.com/dekkerglen/CubeCobra/commit/bcbac8199ff7265af0c3bcf25b255dc339817495) to be self hosted instead of straight from Scryfall.

This PR adjusts this by allowing URLs from Scryfall and from the CARD_IMAGE_BASE_URL.

### DIFF
--- a/packages/server/src/router/routes/tool/imageproxy.ts
+++ b/packages/server/src/router/routes/tool/imageproxy.ts
@@ -8,9 +8,12 @@ export const imageProxyHandler = async (req: Request, res: Response) => {
       return res.status(400).send('Missing or invalid URL parameter');
     }
 
-    // Only allow Scryfall image URLs to prevent abuse
-    if (!url.startsWith('https://cards.scryfall.io/')) {
-      return res.status(403).send('Only Scryfall image URLs are allowed');
+    // Only allow Scryfall or our own card-image CDN URLs to prevent abuse
+    const allowedPrefixes = ['https://cards.scryfall.io/', process.env.CARD_IMAGE_BASE_URL].filter(
+      (prefix): prefix is string => Boolean(prefix),
+    );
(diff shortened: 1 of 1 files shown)


dekkerglen:
@Chr0maticDev can you fix the failing CI?

dekkerglen:
FIxed in another pr

PR #3025, opened 2026-09-10
Add quantity/copies/singleton filter for cube-list boards

haganbmj:
Per-board name-grouped copy counts are stamped on cards in `CubeContext` before the filter runs. Adds `quantity:`, `count:`, `copies:` numeric filters plus `is:singleton` / `not:singleton` category aliases. Scoped per board to match Collapse Duplicates.

Outside a cube list every card reads as one copy, so `quantity>1` returns nothing and `is:singleton` matches everything (documented in the wiki entry).

Also fixes the `isOpValue` grammar wrapper that was silently dropping the `card` argument, so category detectors that need the card object (`is:voucher`, `is:foil` / `nonfoil` / `etched` / `altfoil`, and the new `singleton`) now actually receive it.

EDIT: Removed the generated nearley file because it looks to be out of sync with master.
EDIT2: Bleh.. that causes the tests to fail.
EDIT3: Fixed.


dekkerglen:
@haganbmj can you rebase?

haganbmj:
Yep, done.

